PR #1, opened 2015-02-25
changed initial pacman command

nalla:
Due to the requirement to run post install scripts, the pacman -Syu
command cannot be run in the windows cmd window. Only the initial
msys2-runtime update should be run here.

### DIFF
--- a/setup-git-sdk.bat
+++ b/setup-git-sdk.bat
@@ -29,13 +29,29 @@ SigLevel = Optional\n^
 
 :haverepo
 
-@usr\bin\pacman -Syu --noconfirm
+@usr\bin\pacman -Sy --noconfirm msys2-runtime
 
 @FOR /f "delims=" %%a IN ('usr\bin\uname -m') DO SET arch=%%a
-@usr\bin\pacman -S --noconfirm ^
-	git mingw-w64-%arch%-toolchain python less openssh patch make tar ^
-	diffutils ca-certificates ^
-	perl-Error perl perl-Authen-SASL perl-libwww perl-MIME-tools ^
-	perl-Net-SMTP-SSL perl-TermReadKey winpty-git mingw-w64-%arch%-curl ^
-	mingw-w64-%arch%-expat mingw-w64-%arch%-openssl mingw-w64-%arch%-tcl ^
-	mingw-w64-%arch%-pcre
+@IF "i686"=="%arch%" GOTO setmingw32
+
+@set MSYSTEM=MINGW64
+
+@GOTO runinstall
+
+:setmingw32
+
+@set MSYSTEM=MINGW32
+
+:runinstall
+
+@usr\bin\sh -c "export PATH=/usr/bin:$PATH && pacman -Syu --noconfirm
+
+@set dependencies=git mingw-w64-%arch%-toolchain ^
+ python less openssh patch make tar diffutils ca-certificates ^
+ perl-Error perl perl-Authen-SASL perl-libwww perl-MIME-tools ^
+ perl-Net-SMTP-SSL perl-TermReadKey winpty-git ^
+ mingw-w64-%arch%-curl mingw-w64-%arch%-expat ^
+ mingw-w64-%arch%-openssl mingw-w64-%arch%-tcl ^
+ mingw-w64-%arch%-pcre
+
+@usr\bin\sh -c "export PATH=/usr/bin:$PATH && pacman -S --noconfirm %dependencies%"


dscho:
Oh, I misunderstood... I thought the problem was that the post-install script problem was caused by Pacman (with uses `msys-2.0.dll`) running a bash post-install script *just after updating `msys-2.0.dll`*... was I wrong? Do the post-install scripts still fail when `pacman` is run from `cmd`, even if `msys2-runtime` is left untouched?

nalla:
Yea the `pacman` post install scripts fail in general from `cmd` because they can not find the *linux like path*:

 `error: command (/usr/bin/bash /usr/bin/bash -c . /tmp/alpm_1REmKW/.INSTALL; post_upgrade 20141019-2 20141019-2 ) failed to execute correctly`.

dscho:
Okay, can we call `pacman` in the following way, then (I really do not like the idea of having to keep two files in sync when one should be sufficient):

```
sh -c "pacman -S ..."
```
?

nalla:
```
C:\msys32>usr\bin\sh -c "pacman -S ..."
usr/bin/sh: pacman: command not found
```

```
C:\msys32>usr\bin\sh -c "/usr/bin/pacman -S ..."
[...]
/tmp/alpm_JH7Uly/.INSTALL: line 14: mkdir: command not found
/tmp/alpm_JH7Uly/.INSTALL: line 15: cp: command not found
/tmp/alpm_JH7Uly/.INSTALL: line 16: cp: command not found
/tmp/alpm_JH7Uly/.INSTALL: line 17: cp: command not found
error: command (/usr/bin/bash /usr/bin/bash -c . /tmp/alpm_JH7Uly/.INSTALL; post_upgrade 20141019-2 20141019-2 ) failed to execute correctly
```

```
C:\msys32>cd usr\bin
C:\msys32\usr\bin>sh -c "/usr/bin/pacman -S ..."
[...]
/tmp/alpm_JH7Uly/.INSTALL: line 14: mkdir: command not found
/tmp/alpm_JH7Uly/.INSTALL: line 15: cp: command not found
/tmp/alpm_JH7Uly/.INSTALL: line 16: cp: command not found
/tmp/alpm_JH7Uly/.INSTALL: line 17: cp: command not found
error: command (/usr/bin/bash /usr/bin/bash -c . /tmp/alpm_JH7Uly/.INSTALL; post_upgrade 20141019-2 20141019-2 ) failed to execute correctly
```

`C:\msys32>usr\bin\sh -c "export PATH=/usr/bin:$PATH && pacman -S ..."` seems to work!

dscho:
> `C:\msys32>usr\bin\sh -c "export PATH=/usr/bin:$PATH && pacman -S ..."` seems to work!

Awesome!

> I was unable to accomplish the line split because of the `sh -c "..."`

Yeah, line continuation is tricky with `cmd`. I *think* that something like

```
@usr\bin\sh -c "export PATH=/usr/bin:$PATH && ^
    pacman -S --noconfirm git mingw-w64-%arch%-toolchain ^
    python less openssh patch make tar diffutils ca-certificates ^
    perl-Error perl perl-Authen-SASL perl-libwww perl-MIME-tools ^
    perl-Net-SMTP-SSL perl-TermReadKey winpty-git ^
    mingw-w64-%arch%-curl mingw-w64-%arch%-expat ^
    mingw-w64-%arch%-openssl mingw-w64-%arch%-tcl ^
    mingw-w64-%arch%-pcre"
```

should work.

nalla:
@dscho your proposal did not work. sh -c could somehow not handle it. I declared a variable. That worked.

dscho:
Nice!

Now, I'd like to ask you for a few more things before merging:

1. Could you Sign-off on those commits?
2. I would like to avoid splitting the script and merging it right away; In general, we try to avoid introducing changes with a topic branch that are reverted in the same topic branch. The commit message, however, is an excellent location to describe what we tried, and why we went with a different solution
3. In the relevant commit message, we should describe *how* calling `pacman` directly from `cmd` did not work: its post-installation shell script could not be executed because the shell was not found.

To make those changes, I highly recommend a so-callled [*interactive rebase*](http://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) (`git rebase -i origin/master`).

Sound good?

nalla:
@dscho I hope i did this right.. first time working with `rebase -i`

dscho:
Oh, it looks fine to me! I'm okay with merging as-is.

sschuberth:
FYI, we can probably deprecate `setup-git-sdk.bat` altogether in favor of https://github.com/git-for-windows/MSYS2-packages/pull/7. Please give the latter a try.

dscho:
I agree, of course. @nalla thanks!

PR #2, opened 2015-03-06
Add tooling to work with the Pacman repository on Bintray

dscho:
So far, we (ab-)used a repository on GitHub to serve a Pacman repository. However, it is much better to use Bintray for that, as its mission is to deliver binaries. To do so, we need new tooling.

With the `pacman-mirror.sh` script added in this Pull Request, one can perform a `fetch` to initialize and update a local mirror of the Pacman repository, a `add <file>...` to add packages to the local mirror, and a `push` to index and (incrementally) upload the local files to Bintray.

### DIFF
--- a/pacman-mirror.sh
+++ b/pacman-mirror.sh
@@ -230,7 +230,7 @@ push () {
 			;;
 		*)
 			package_exists $basename ||
-			die "The package $basename does not yet exist..."
+			die "The package $basename does not yet exist... Add it at https://bintray.com/git-for-windows/pacman/new/package?pkgPath="
 			;;
 		esac
 	done
--- a/pacman-mirror.sh
+++ b/pacman-mirror.sh
@@ -52,32 +52,32 @@ fetch () {
 			dir="$(arch_dir $msystem $arch)"
 			mkdir -p "$dir"
 			(cd "$dir" &&
-			 curl -sO $arch_url/git-for-windows.db.tar.xz &&
+			 curl -sfO $arch_url/git-for-windows.db.tar.xz ||
+			 continue
 			 for name in $(package_list git-for-windows.db.tar.xz)
 			 do
 				filename=$name-$arch.pkg.tar.xz
 				test -f $filename ||
 				curl --cacert /usr/ssl/certs/ca-bundle.crt \
-					-sLO $base_url/$arch/$filename ||
+					-sfLO $base_url/$arch/$filename ||
 				exit
-			 done ||
-			 exit
+			 done
 			)
 		done
 	done
 }
 
 upload () { # <package> <version> <msystem> <arch> <filename>
-	curl --netrc -T "$5" "$content_url/$1/$2/$3/$4/$5"
+	curl --netrc -fT "$5" "$content_url/$1/$2/$3/$4/$5"
 }
 
 publish () { # <package> <version>
-	curl --netrc -X POST "$content_url/$1/$2/publish"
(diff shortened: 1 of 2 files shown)


dscho:
We will need to adjust the, and build a new, SDK installer, too... :grinning:

dscho:
@sschuberth @t-b could you have a look?

Oh, and I guess that it is `pacman` more than the SDK installer that requires the change... Sorry!

sschuberth:
@dscho I have somewhat related changes in the pipe for the pacman and pacman-mirror packages, and will test your script in the context of rebuilding these packages.

t-b:
@sschuberth @dscho Sorry I won't find the time for a review. Therefore go ahead.
Just a quick question, we will use the bintray repositories only for the sdk installer or will also the end user installer get packages from it?

sschuberth:
@t-b For both, I believe. If the end user installer is retrieving package at all, as my understanding was it should be self-contained.

t-b:
@sschuberth "self-contained" meaning needing no internet for installation, correct?

dscho:
@t-b correct. Although the current installer still tries to update via internet first thing after installing files into place.

dscho:
> @dscho I have somewhat related changes in the pipe for the pacman and pacman-mirror packages, and will test your script in the context of rebuilding these packages.

Cool! I hope I did not interfere with your changes, though!

sschuberth:
> Cool! I hope I did not interfere with your changes, though!

No, all fine.

dscho:
BTW I wonder whether it would make sense to rename the `add` subcommand to `addlocal` and then introduce a new `add` function that performs a `fetch`, then an `addlocal` and an immediate `push` right away, i.e. the common workflow when trying to upload new package versions?

sschuberth:
> BTW I wonder whether it would make sense to rename the add subcommand to addlocal

Let's leave this as a future improvement when we've used the script for a while and know how much the lack of such a command bothers us.

I think this is good to merge to expose the script to testing in daily use.

dscho:
> Let's leave this as a future improvement

No, that is not possible anymore. We cannot change the meaning of the `add` subcommand now, that would be very poor development style, and I won't do that.

sschuberth:
Come on, isn't that a bit over-cautious? Just because we made the script public does not mean it has users. Not even @t-b or me have used it yet. It reminds me a bit of the story why Makefiles require tabs for indentation ;-)

Anyway, we could always call the new command `addremote` or `sync` or whatever.

dscho:
> Come on, isn't that a bit over-cautious?

It is a bad idea to get into the habit to break good practices. So: no, it is not over-cautious, it is just keeping a good development style. I am adamant about that, if you haven't noticed yet.

`addremote` is an option. The problem is that it is a poor design choice: in the future, I envisage the main usage to be that `addremote` function, not the `add` function. Good design dictates that the goal is to make the common case the easiest one (it's just like `git cherry-pick` being much more common than `git cherry`, but the latter taking the better name away from the former).

`sync` is a poor choice, though, because we are working in the `Pacman` context where `sync` already has the opposite meaning of what we are trying to do.

sschuberth:
If that's how you think you shouldn't have made it sound like an optional suggestion to rename the add subcommand to addlocal. Now it sounds to me as if there's actually no alternative to that in your opinion.

I'm also adamant about good development style, but still IMO changing this 10 minutes after is has been merged and is not used by anyone yet is a valid option.

PR #3, opened 2015-03-12
new-aliases-file

nalla:
Added an `aliases.sh` file in the profile.d folder that has the aliases that were
previously defined in `msysgit/etc/bash_profile`.


dscho:
Nice, thanks!

PR #4, opened 2015-03-12
vim-clipboard-mode

nalla:
By default `VIM` has no access to the systems clipboard under windows. The
setting `set clipboard=unnamed` will grant this access by default. This
behavior is something that IMHO many endusers would expect. See also this
posting in the `VIM` wiki: http://vim.wikia.com/wiki/Easy_pasting_to_Windows_applications

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -9,6 +9,7 @@ set backspace=indent,eol,start  " make that backspace key work the way it should
 set nocompatible                " vi compatible is LAME
 set background=dark             " Use colours that work well on a dark background (Console is usually black)
 set showmode                    " show the current mode
+set clipboard=unnamed           " set clipboard to unnamed to access the system clipboard under windows
 syntax on                       " turn syntax highlighting on by default
 
 " Show EOL type and last modified timestamp, right after the filename


dscho:
Nice find!

justinmk:
@dscho Thanks for the reply. I thought my comment was constructive but I will expand it. 

Defaults should not be changed unless the benefit strongly outweighs the cost. This change satisfies some unknown slice of Vim users, so it isn't a clear win. Not only is it not a clear win, but it most likely does more harm than good, because:

- Most Git-for-Windows users probably aren't very familiar with Vim, so they will only be confused if an operation overwrites their clipboard.
- Experienced Vim users will be annoyed because the Vim defaults were changed.
- The remaining subset of experience Vim users who _want_ this change can simply set the option in their vimrc.

Since you consider `"+yy` and/or setting the option in your vimrc to be "jumping through hoops" you can understand why other Vim users would be annoyed. If there isn't overwhelming support for a change, then **upstream defaults should always take precedence.** 

> merged into master for ages

Most GFW users were on 1.9.5 until recently, so this change broke the previous GFW Vim defaults (or lack thereof).

> Do you have any constructive idea how to fix this without breaking things (things such as my expectation that the clipboard is now shared between Windows and vi)?

It doesn't make sense to insist that other users change their `.vimrc`s unless there is a clear and obvious gain to be had by changing a default. There isn't, so you should change your `.vimrc` instead of shipping a system default. In general, changing the defaults of upstream software should be avoided totally.

dscho:
> In general, changing the defaults of upstream software should be avoided totally.

This would be too limiting.

At the time this PR was opened, there was unilateral support for the change. The first complaint came a full 5 months afterwards. Still a full five weeks after the first official version of Git for Windows.

So now I am no longer willing to change our default. I would be willing to accept an opt-out in the installer, though.

justinmk:
Yeah, I shouldn't have bothered. I just wanted to let you know this choice was awful.

dscho:
> Yeah, I shouldn't have bothered. I just wanted to let you know this choice was awful.

Now, this is truly unhelpful. Please follow my example when I at least tried to find a way to give everybody what they want. I want the clipboard, you don't. That would be possible, both, if you contributed a Pull Request to offer that option in the installer. Of course, it would take a little bit more than just talkin' ;-)

justinmk:
An installer option is disproportionate to the problem being solved, since as we established, this setting can be corrected in vimrc. I gave a cost-benefit argument for choosing defaults, the reply was aw-shucks.

dscho:
Sorry, this kind of response cannot possibly be designed to be constructive. What other goal could this statement ("this setting can be corrected... aw-shucks") have that to make other people feel bad? This is not how this project operates. If you want to make other people feel good about your comments, please try again.

dscho:
> Yeah, I shouldn't have bothered. I just wanted to let you know this choice was awful.

@justinmk I understand that you are frustrated with my preference that is completely diametrically opposed to your preference. However, there is no substitute for a constructive discussion.

I still think that it would make sense to add that option to the installer.

But I am not married to that idea. If you can think of another solution that lets both of us have what we want -- you can have your vim with a private clipboard, I can have mine that shares the clipboard with the system just like all the other editors on Windows -- I am definitely interested to hear it.

dra27:
Can I chip in with a hopefully constructive point of view? Having used Vim (mainly on Windows) for the last 15 years, I too think this is a terrible default but with some different reasons:
1. @nalla stated "By default  VIM  has no access to the systems clipboard under windows". This is not true - it's available in the `*` register, same as on any other system. I would have expected, 5 months ago, this pull request to have been rejected with information on using `"*` to access the system clipboard
2. Windows applications usually have an explicit command for accessing the clipboard (i.e. using CTRL+C or CTRL+X) and I would go so far as to say that there is no other Windows application where simply deleting (e.g. using backspace or the Delete key - i.e. rough equivalents to using x or dd in Vim). I would have much less (or even no problem) if this setting affected the yank command only (so for example if `yj` yanked to the system clipboard by default, but `dd` used the unnamed register)
Out of personal curiosity only, are either @dscho or @nalla users of Vim on Windows for anything other than Git?
Can I suggest one of two alternatives to the present default:
1. It's possible to change the yank command only via remap (so Windows Vim power users don't get unexpected results from `x`, `dd`, `c`, etc.) - see http://stackoverflow.com/questions/13380643/vim-use-as-default-register-only-for-yank-command. So in this way the actual yank-and-put model defaults to the Windows clipboard. That may surprise a Vim power user, but I would hope be less likely to annoy!
2. That said, it's mildly surprising to me given that we're Git-for-Windows that the vimrc doesn't include `behave mswin` (in fact, I can't see that Vim's `mswin.vim` is even installed). If the aim is to make Vim friendlier for the Windows users who only use it for Git, why not map the default Windows shortcuts for accessing the clipboard - which is the default if you install Vim on Windows. These lines taken from mswin.vim in the 7.4.752 runtime compiled on my machine:

```vim
" CTRL-X and SHIFT-Del are Cut
vnoremap <C-X> "+x
vnoremap <S-Del> "+x

" CTRL-C and CTRL-Insert are Copy
vnoremap <C-C> "+y
vnoremap <C-Insert> "+y

" CTRL-V and SHIFT-Insert are Paste
map <C-V>		"+gP
map <S-Insert>		"+gP

cmap <C-V>		<C-R>+
cmap <S-Insert>		<C-R>+

imap <S-Insert>		<C-V>
vmap <S-Insert>		<C-V>
```

The existing default could then be reduced to a Git-for-Windows Wiki item - which would probably reference the Vim tip as well.

dscho:
> vimrc doesn't include behave mswin.

Pull Request?

dra27:
Sure, if you want. Forgive me, I know there is more than one way of running a project, but in other projects I work on, we discuss changes and then we do the code, especially when there's already a discussion chain which suggests the change may not be wanted! Equally happy to put this in an issue on its own which can then be upgraded to a full pull request, but I regard the change as related to this PR (see 2 below)
I'd be keen to include:
```vim
behave mswin
```
or even better:
```vim
source $VIMRUNTIME/mswin.vim
```
but the change has two consequences:
1. It would make the Vim in Git-for-Windows behave more like Vim-for-Windows, but therefore not necessarily the way a Unix user suddenly forced to use Git on Windows might expect (I personally think that's tolerable, but it is a change)
2. I would suggest that e32a158de5075c0d22cf3441fa420e1f0257a978 should be reverted, as the Windows support for standard clipboard shortcuts would be available by a better mechanism

dscho:
> Sure, if you want. Forgive me, I know there is more than one way of running a project, but in other projects I work on, we discuss changes and then we do the code, especially when there's already a discussion chain which suggests the change may not be wanted

Well, I thought the `behave mswin` patch might be uncontroversial enough, and I would really like to see a kick-ass commit message that describes what it implies. 'cause I don't know that behavior.

PR #5, opened 2015-03-12
Adjust Pacman mirror script to current layout

dscho:
Our current layout differentiates only between architectures, no longer between msystems. Let's adjust the `pacman-mirror.sh` script to reflect that.


dscho:
@sschuberth please feel free to merge if you have no objection to the changes.

PR #6, opened 2015-03-12
default-gitconfig

nalla:
Added a `default-gitconfig`. This file will be parsed by the `git-extra`
`PKGBUILD` file. The content will then be written in the respective
location based on the `MinGW` system (either `mingw32` or `mingw64`) installed.


nalla:
Only `gitconfig` sounds indeed better.

dscho:
Thanks!

PR #7, opened 2015-03-13
git-extra: Relocation of the git-extra package

nalla:
The splitup of the package file and the source was not such a good idea.
It was decided that the package file should be placed with the source
files. The `PKGBUILD` was adapted to handle the source files relative to
its location. The old location of this packe was in the
https://github.com/git-for-windows/msys2-packages repository.

Notes:
* This is for https://github.com/git-for-windows/git/issues/31
* Before the first build you need to run `touch git-extra.install`. If you dont do that, `makepkg` will complain about the `git-extra.install` file being missing.

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -1,0 +1,63 @@
+# Maintainer: nalla <nalla@hamal.uberspace.de>
+
+pkgname=('git-extra')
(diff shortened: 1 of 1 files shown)


dscho:
Excellent! Thank you so much! Even the `nsswitch.conf` file is handled.

PR #11, opened 2015-03-18
Proposal: Modify output color of `__git_ps1` call

nalla:
To distinguish between the working directory and the current git branch
the output of the `__git_ps1` call should be in a different color than the
rest of the prompt.

This is more like a proposal. If you feel this is unnecessary and is up to the end user in general that's fine. The default color is of course debatable.


dscho:
Good point!

PR #12, opened 2015-03-20
git-prompt: Reset to default color after changing it for the branch name

sschuberth:
This makes color display correctly for programs that do not explicitly set
it but rely on the default color to be the current color.


sschuberth:
This is a follow-up to https://github.com/git-for-windows/build-extra/pull/11.

dscho:
Thanks!

PR #14, opened 2015-03-23
net-installer: Dependency fixes

nalla:
The install of `pacman` has a dependency for `gettext`. The `gettext`
package seems to fail with its post install script if the `info` package
is not installed. So to overcome this unresolved dependency, just install
the required package so that the post install script will succeed.

### DIFF
--- a/net-installer/setup-git-sdk.bat
+++ b/net-installer/setup-git-sdk.bat
@@ -38,8 +38,8 @@
 	@EXIT 1
 )
 
-@REM next, force update pacman
-@%cwd%\usr\bin\pacman -S --force --noconfirm pacman
+@REM next, force update pacman, but first we need bash and info for that.
+@%cwd%\usr\bin\pacman -S --force --noconfirm bash info pacman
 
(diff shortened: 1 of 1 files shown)


sschuberth:
Do we agree that the correct solution for this would be to make `info` a dependency of `gettext`, but this is not done to save us from overriding the `gettext` packaged compared to upstream? If so, and given how quick upstream has merged our recently contributions, shouldn't we simply fix this upstream right away?

nalla:
Well. It's not really a dependency. Its an optional dependency for installing the info packages. There seems to be a bug in the post-install script. Not just in the `gettext` package, but in others as well. We could look through all packages that runs post-install scripts and are executing the `install-info` command. But that should not be our priority?

sschuberth:
> But that should not be our priority?

It should not be our priority to fix MSYS2 issues, no. But whenever we fix something here instead of upstream I'd like to see a justification why that is, because we should try to not diverge from upstream too much. Basically your "Not just in the `gettext` package, but in others as well" for me is justification enough right now, as it indicates the problem is a bigger one than fixing a single dependency in a single package file.

PR #15, opened 2015-03-23
portable-release: handle spaces in install path

nalla:
The `git-bash.bat` can not handle setups that were installed in a path
with a space in it. According to `help start` you are providing a title if
you use quotes. Because of this a dummy title must be provided to quote
the path to the console application.


dscho:
Thank you!

PR #17, opened 2015-03-31
build-extra: add $HOME/bin to path

nalla:
This behavior was present in the *msysgit* version of the `/etc/profile`
file. The `git-extra` package now adds a `homepath.sh` into the
`/etc/profile.d` directory to restore the behavior in the *MSYS2*
environment.

This *should* resolve https://github.com/git-for-windows/git/issues/57.

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -65,5 +65,6 @@ package() {
   install -m755 create-shortcut.exe $pkgdir/$mingwdir/bin
   install -m755 $startdir/git-prompt.sh $pkgdir/etc/profile.d
   install -m755 $startdir/aliases.sh $pkgdir/etc/profile.d
+  install -m755 $startdir/env.sh $pkgdir/etc/profile.d
   install -m644 $startdir/msys2-32.ico $pkgdir/usr/share/git
 }
--- a/git-extra/env.sh
+++ b/git-extra/env.sh
@@ -1,0 +1,2 @@
+# Add bin path in the home directory ontop of the PATH variable
+export PATH="$HOME/bin:$PATH"


git-extra/PKGBUILD
```diff
@@ -65,5 +65,6 @@ package() {
   install -m755 create-shortcut.exe $pkgdir/$mingwdir/bin
   install -m755 $startdir/git-prompt.sh $pkgdir/etc/profile.d
   install -m755 $startdir/aliases.sh $pkgdir/etc/profile.d
+  install -m755 $startdir/homepath.sh $pkgdir/etc/profile.d
```
sschuberth: Maybe rename the file to something else? "homepath" sounds to me as if the path to home, i.e. the HOME variable, was being tampered with. Why not just call it "env.sh" or so to say that any environment overrides should be put into there?

dscho: I do not like `env` because it says even less about what purpose the file serves. The name `homepath.sh` is good enough for me.

sschuberth:
> I do not like env because it says even less about what purpose the file serves.

@dscho Saying less about the *current* use of the file was the intention of my proposal. In exchange we do not need to create another file if we ever wanted to modify another environment variable. Think of it like e.g. Ubuntu's `/etc/environment` file.

dscho:
If we choose a name as generic as `env.sh`, it takes *just one* other package to make the same choice to cause trouble.

No, I still prefer the `homepath.sh`. And I also prefer to avoid spending so much time to discuss the color of this bike shed when we still have an entire parking lot to build.

sschuberth:
So retain your picture: I don't think we *can* build an entire parking lot with poor material. Since you prefer not to spend more time discussing, the quickest thing to do is to merge as-is, which is what I'll do now.

dscho:
If we spend as much time discussing the parking lot as we spend on the color of this bike shed (which BTW almost nobody will ever *see*), then we won't get started with anything in the way of building the parking lot for a long, long time. I am not prepared to wait for that long, so please understand I have to spend substantially less time clarifying the goals, the technical issues, and possible solutions. It would be good to keep the discussions more to the important points, for that reason. Because at the end of the day, I care about the result, not about the discussion.

PR #18, opened 2015-03-31
net-installer: install requirements to build doc

nalla:
Since the *Git for Windows SDK* should provide functionality to build git
it should also provide all required packages to build the git doc.
Building doc is needed when shipping the git mingw package lets just
install those reuqirements too then.


sschuberth:
Thanks!

PR #21, opened 2015-04-02
git-extra: Add inputrc

elyscape:
This brings Git Bash's input and completion behavior in line with the
previous msysGit-based releases.

Fixes git-for-windows/git#62.

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -60,6 +60,7 @@ package() {
   install -d -m755 $pkgdir/usr/bin
   install -d -m755 $pkgdir/usr/share/git
   install -d -m755 $pkgdir/$mingwdir/bin
+  install -m644 $startdir/inputrc $pkgdir/etc
   install -m644 $startdir/vimrc $pkgdir/etc
   install -m755 $startdir/vi $pkgdir/usr/bin
   install -m755 $startdir/wordpad $pkgdir/usr/bin
--- a/git-extra/inputrc
+++ b/git-extra/inputrc
@@ -1,0 +1,26 @@
+# none, visible or audible
+set bell-style visible
+
+# Ask before displaying >40 items
+# Since $WINDIR $PATH var can be in $PATH, this could list
+# all window executables in C:\WINDOWS
+set completion-query-items 40
+
+# Ignore case for the command-line-completion functionality
+# on:  default on a Windows style console
+# off: default on a *nix style console
+set completion-ignore-case on
+
+# disable/enable 8bit input
+set input-meta on
+set output-meta on
+set convert-meta off
+
+# visible-stats
+# Append a mark according to the file type in a listing
+set visible-stats off
+set mark-directories on
+set mark-symlinked-directories on
+
+# Show all instead of beeping first
+set show-all-if-ambiguous on


dscho:
Thank you!

elyscape:
@dscho: In the process of working on some improvements to the inputrc file (e.g. supporting page-up/page-down and home/end), I discovered that the filesystem package already drops a customized .inputrc file into the /etc/skel directory.  It doesn't appear to be part of the developer's preview, though, despite that package being listed as a dependency for git-extra.  Was that intentional?  I want to make sure I don't end up with conflicts in my changes.

dscho:
@elyscape good point. The `filesystem` package is actually too broad to include wholesale. That is why the `git-extra` package [lists it as an optional dependency](https://github.com/git-for-windows/build-extra/blob/f655faf9319ecc4e71d6ba3009b30d3a9aaf4ace/git-extra/PKGBUILD#L12).

In any case, the `/etc/skel/` directory is only relevant when a new home directory is created. And we do not do that; We [ensure that the *real* home directory is used](https://github.com/git-for-windows/build-extra/blob/f655faf9319ecc4e71d6ba3009b30d3a9aaf4ace/git-extra/PKGBUILD#L25-L26), i.e. the directory *Windows* thinks is the user's home directory.

elyscape:
Makes sense.

> In any case, the `/etc/skel/` directory is only relevant when a new home directory is created. And we do not do that; We [ensure that the real home directory is used](https://github.com/git-for-windows/build-extra/blob/f655faf9319ecc4e71d6ba3009b30d3a9aaf4ace/git-extra/PKGBUILD#L25-L26), i.e. the directory Windows thinks is the user's home directory.

Good to know.  With this in mind, it's worth noting that the current developer's preview ships with the file `/etc/skel/.vimrc`.  I'm not quite sure where it came from, though this is due in part to the fact that I'm not familiar with the final packaging process for the Git for Windows releases.

dscho:
To find out, you could run a variation of [these commands](https://github.com/git-for-windows/build-extra/blob/f655faf9319ecc4e71d6ba3009b30d3a9aaf4ace/portable/release.sh#L50-L63) manually (e.g. skipping [this line](https://github.com/git-for-windows/build-extra/blob/f655faf9319ecc4e71d6ba3009b30d3a9aaf4ace/portable/release.sh#L58) which strips out the package name from the output of Pacman, leaving only the file name).

elyscape:
Looks like it came from the vim package.

dscho:
> Looks like it came from the vim package.

Nicely researched. Maybe we should just skip the `/etc/skel/` files by adding `-e '^/etc/skel/'` [here](https://github.com/git-for-windows/build-extra/blob/675a89245d220736a5e3a1e07d7d5e003edb6845/installer/release.sh#L94)?

elyscape:
Probably a good idea.

While we're looking at the release script, why are the man files excluded from the distribution? It seems to me that, if the mandb package were added, they would be useful. I know that I've pretty often tried something like `man sed` to look up some syntax info and subsequently had to find the manpages online since they weren't included.

dscho:
> why are the man files excluded from the distribution?

Well, given that the installer almost doubled in size since 1.9.5 (and that even *without* man pages), I am reluctant to add things, unless we really need them. And Windows users are not used to man pages, that's why we default to HTML pages when the user calls `git help <command>`.

elyscape:
Fair enough.

PR #24, opened 2015-04-02
installer: Prepare release notes and documentation

nalla:
Adapted the release notes for the upcoming first *Git for Windows*
release. Furthermore the instructions to build and deploy a new upstream
release were adapted to the current process. That process is described in
the `HowToRelease.txt` file.


dscho:
Thank you! I will most likely touch both files up a bit more (I really would like to avoid the `make install` step and instead release via Pacman, then update via `pacman -Syu`) and separate the changes into two different commits, but you gave me a head start!

PR #25, opened 2015-04-03
git-extra: Enable core.fscache

elyscape:
Enabling core.fscache provides a significant performance boost.

### DIFF
--- a/git-extra/gitconfig
+++ b/git-extra/gitconfig
@@ -1,6 +1,7 @@
 [core]
 	symlinks = false
 	autocrlf = true
+	fscache = true
 [color]
 	diff = auto
 	status = auto


elyscape:
Pinging @kblees, who wrote the fscache code, for input on if there's any reason not to enable it by default.  Thoughts?

dscho:
It's a good idea, but I was worried that this feature might still be a bit too experimental. How much of testing did you give it?

Maybe a more cautious way would be to add this as a new page to the installer (like the [line endings one](https://github.com/git-for-windows/build-extra/blob/b0d60ef013e98e3e667787f722c7e5c16a04e734/installer/install.iss.in#L669-L772)). We could mark the fscache as an experimental option and default to the "classic" option.

elyscape:
I've tested it a little but not particularly in-depth.  It's been around for a few years, though, and I've seen a decent number of recommendations that people enable it.  As for changing this such that the installer has a page that offers to turn this on, that'd work too, though I'm curious what @kblees' thoughts on the matter are.

dscho:
The new installer page will give this feature the coverage we require to determine whether it is robust enough for every-day use. I just merged the `inno-setup` branch into the `build-extra` repository, after pulling, `/usr/src/build-extra/release.sh 2.3.5-elyscape` will build an installer. Wanna give it a try?

elyscape:
I'll try it ou!

elyscape:
Installer page created and submitted as #26.

kblees:
I consider the fscache code stable, and haven't used git without it for a long time...

IMO the only reason _not_ to enable it by default is because there hasn't been a thorough code review of compat/win32/fscache.[ch].

See also http://thread.gmane.org/gmane.comp.version-control.git/250572

PR #26, opened 2015-04-04
installer: Add installer page for performance tweaks

elyscape:
The core.fscache config option provides a significant performance boost,
but it's still experimental. Since wider use can help flush out any
issues and push this from being experimental to production-ready, this
adds a page to the installer advising users of the setting's existence
and allowing them to enable it.

### DIFF
--- a/installer/install.iss.in
+++ b/installer/install.iss.in
@@ -195,6 +195,9 @@ const
     GC_CRLFAlways     = 2;
     GC_CRLFCommitAsIs = 3;
 
+    // Git performance tweaks options
+    GP_FSCache        = 1;
+
     // BindImageEx API constants.
     BIND_NO_BOUND_IMPORTS  = $00000001;
     BIND_NO_UPDATE         = $00000002;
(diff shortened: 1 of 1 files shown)


sschuberth:
Thanks, I like the idea of having a separate installer page for performance tweaks / experimental settings. However, instead of using radio buttons for settings where you only have the choice to enable or disable, I'd suggest to use checkboxes instead. Also, I'd suggest to change the wording to be a bit more specific. "Do not enable experimental performance tweaks" could mean anything. In the future, there might be more performance related tweaks to toggle, each with its own checkbox, so the existing description should be less general. Finally, usually positive logic ("Enable caching of file system data") is better understood by the user.

elyscape:
@sschuberth Good idea.  On the other hand, I added this page because @dscho worried that the `core.fscache` feature might be too experimental, and @kblees (who wrote the feature) has since expressed the view that it is stable.  Given that, should we scrap this and just enable it by default?

dscho:
No, we need to mark it as experimental, still. @elyscape I am far from an InnoSetup expert, so the only constructive information I can give you is a quick pointer: http://www.jrsoftware.org/ishelp/index.php?topic=scriptclasses, sorry!

dscho:
(Actually, I should have mentioned that `TNewCheckBox` is probably what you need, or at least related to the solution.

dscho:
Have you started converting the radio buttons to checkboxes? Otherwise I will merge thos PR as-is and make the change myself today.

elyscape:
Give me 20 minutes or so and I'll have it ready.

elyscape:
Okay, we should be good to go now.

dscho:
You are awesome! I will build a full installer and test it right now.

PR #28, opened 2015-04-06
Fixes of git-for-windows/git/#65

lygstate:


lygstate:
anyway, mark this closed first, this is not the best option.

PR #30, opened 2015-04-07
git-extra: Add key mappings to inputrc

elyscape:
These are imported from MSYS2, except for the Mintty-specific macros.


dscho:
Looks good. I trust you tested a couple of these keyboard combinations and verified that the `inputrc` changes are required to make them work?

elyscape:
Not all of them, but some, yeah.  The Lines 36-42 I added myself, and the rest came from the `/etc/skel/.inputrc` file provided by the filesystem package.

dscho:
That's good enough for me. Thanks!

PR #31, opened 2015-04-07
make-file-list: Exclude skel files

elyscape:
These aren't used by Git for Windows, as we don't create the user's home
directory.


elyscape:
This is a follow-up on the conversation in #21.

dscho:
Thank you!

PR #34, opened 2015-04-13
git-extra: silence missing package error

nalla:
When checking for the https fix the script checks for every valid
comination to be patched. Since those packages are not installed on every
system lets not worry the user when updating `git-extra`.


dscho:
Oh, right, I test for both architectures... Good catch!

PR #37, opened 2015-04-14
Simplify Windows MKDIR calls a bit

sschuberth:
Windows' "MKDIR creates any intermediate directories in the path, if
needed", so just suppress the "A subdirectory or file a already exists"
message eventually. Also ensure to create the "dev" directories in the
installation directory in case setup-git-sdk.bat was called from another
directory.


dscho:
Very nice!

PR #45, opened 2015-04-27
portable: Install Portable Git to a portable directory

elyscape:
Portable applications should not be installed to Program Files. `%%S`
expands to the folder containing the self-extracting archive.

### DIFF
--- a/portable/release.sh
+++ b/portable/release.sh
@@ -73,7 +73,7 @@ echo "Creating archive" &&
  echo 'ExtractTitle="Extracting..."' &&
  echo 'GUIFlags="8+32+64+256+4096"' &&
  echo 'GUIMode="1"' &&
- echo 'InstallPath="%'$PROGRAMFILESENV'%\\Git"' &&
+ echo 'InstallPath="%%S\\PortableGit"' &&
  echo 'OverwriteMode="0"' &&
  echo ';!@InstallEnd@!' &&
  cat "$TMPPACK") > "$TARGET" &&


portable/release.sh
```diff
@@ -73,7 +73,7 @@ echo "Creating archive" &&
  echo 'ExtractTitle="Extracting..."' &&
  echo 'GUIFlags="8+32+64+256+4096"' &&
  echo 'GUIMode="1"' &&
- echo 'InstallPath="%'$PROGRAMFILESENV'%\\Git"' &&
+ echo 'InstallPath="%%S\\PortableGit"' &&
```
sschuberth: If PROGRAMFILESENV is not used here, its initialization should also get removed from the beginning of the script.

sschuberth:
> Portable applications should not be installed to Program Files.

I agree. With reason, Windows users would expect that anything that goes into "Program Files" has been installed by a "real" installer, and can also be cleanly uninstalled by launching the associated uninstaller from the Control Panel.

I just checked, and downloads from [PortableApps.com](http://portableapps.com/) default to extract to a sub-directory of the directory containing the portable installer, just like this PR suggests, so I think we should do that, too.

dscho:
I register my concern that this default is often, maybe most of the time, unhelpful: if you download the installer with the option "run it right away", and the user does not pay attention, there is a good chance to unpack the portable Git into a directory *that is removed soon after*.

But for the time being, let's just go with this PR because the Program Files default is even more unhelpful.

sschuberth:
@dscho It would have been nice if you waited with the merge to give @elyscape a chance to addresse my remark about removing the `PROGRAMFILESENV` assignment.

portable/release.sh
elyscape: Good point.  Too late now, unfortunately.  I'll whip up another pull request.

dscho:
@sschuberth funny, I just wanted to do it myself, only to find that you already merged a PR to that extent. Seems we are stepping on each other toes again.

sschuberth:
> Seems we are stepping on each other toes again.

Fear not, this was the last time, I've already [said farewell](https://github.com/git-for-windows/git/issues/82#issuecomment-96565443). You're free now to continue hasty merging PRs and cleaning them up later yourself instead of me or others helping you doing that.

PR #46, opened 2015-04-27
Include post-install scripts and run after installation

elyscape:
This pull request adds the useful post-install scripts to the release packages and ensures that they are executed once after installation/extraction, at which point they are deleted.  Deletion is not done if `msys2.ico` is in the root directory, as that is (at present) a reliable indicator that we're in the SDK, and the post-install scripts need to be in place for the packages to be made.

Resolves git-for-windows/git#111.


elyscape:
Hold a sec.  Just realized that it would be a good idea to log potential failures in the installer of the post-install script.

elyscape:
Okay, failure logging added.

dscho:
Thank you very much!

PR #48, opened 2015-04-27
installer: remove temporary `git-wrapper.exe`

nalla:
After copying the `git-log.exe` to the `tmp` directory for copying the
different `builtins` the temporary file is never cleaned up. This cleanup
call is now issued properly.

This fixes git-for-windows/git#114

Signed-off-by: 마누엘 <nalla@hamal.uberspace.de>

### DIFF
--- a/installer/install.iss.in
+++ b/installer/install.iss.in
@@ -1070,6 +1070,7 @@ begin
             end;
         end;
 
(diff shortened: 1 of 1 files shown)


dscho:
Excellent, thank you!

PR #50, opened 2015-04-29
installer: honor `%PROGRAMDATA%\Git\config`

nalla:
Before trying to set the selected config values in the installer the code
now checks if the system wide git config exists. If that is not the case
the installer copies the shipped config to `%PROGRAMDATA%\Git\config` and
then tries to alter it as choosen by the user.

The shipped gitconfig is deleted after checking if the system wide git
config already exists.

### DIFF
--- a/installer/install.iss.in
+++ b/installer/install.iss.in
@@ -1019,7 +1019,7 @@ end;
 
 procedure CurStepChanged(CurStep:TSetupStep);
 var
-    AppDir,DllPath,FileName,Cmd,Msg:String;
+    AppDir,ProgramData,DllPath,FileName,Cmd,Msg:String;
     BuiltIns,ImageNames,EnvPath,EnvHome:TArrayOfString;
     Count,i:Longint;
     LinkCreated:Boolean;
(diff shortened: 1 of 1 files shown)


nalla:
@dscho http://www.jrsoftware.org/ishelp/index.php?topic=consts indicates `{commonappdata}`. Worked fine for `Win7`. I updated the PR. *Could* you verify that it behaves correctly on *XP*? I have no tools to do that.

dscho:
> Could you verify that it behaves correctly on XP?

Will do!

dscho:
Actually, it is much easier for me to test after merging ;-) Thanks!

dscho:
Okay, I finally was able to test. The sad news is that I had considerable trouble to test this. The good news is that my troubles were caused [by a different issue](https://github.com/git-for-windows/git/issues/125).

But I realized one thing: the location of `http.sslCAInfo` is actually *relative* to the prefix. That means that putting it into the Windows-wide configuration (which is accessed possibly by multiple Git installations) is probably incorrect. We *might* want to ensure that the file path is expanded to a full Windows path. Or put that particular setting back into `/mingw64/etc/gitconfig`...

PR #51, opened 2015-04-29
installer: Fix .sh file association

elyscape:
`%1` expands to the short name of the file, rather than the directory
the file is contained in, so `--cd=%1` fails when `git-bash.exe` tries
to change the working directory to a file. `%w` expands to the file's
containing directory and `%L` expands to the long name of the file.

Additionally, this sets the `ChangesAssociations` flag to `yes`, so that
Explorer will refresh the system's file association information after
the install. It also changes the `DropHandler` to the WSH `DropHandler`,
which sends long file names instead of short file names, and forces
update of the command string.

Fixes git-for-windows/git#120.

Helped-by: @Animeye

### DIFF
--- a/installer/install.iss.in
+++ b/installer/install.iss.in
@@ -36,6 +36,7 @@ AppName={#APP_NAME}
 AppPublisher=The Git Development Community
 AppPublisherURL={#APP_URL}
 AppVersion={#APP_VERSION}
+ChangesAssociations=yes
 ChangesEnvironment=yes
 CloseApplications=no
 DefaultDirName={pf}\{#APP_NAME}
--- a/installer/install.iss.in
+++ b/installer/install.iss.in
@@ -16,7 +16,7 @@
 
 #define PLINK_PATH_ERROR_MSG 'Please enter a valid path to a Plink executable.'
 
-#define DROP_HANDLER_GUID '{{86C86720-42A0-1069-A2E8-08002B30309D}'
+#define DROP_HANDLER_GUID '{{60254CA5-953B-11CF-8C96-00AA00B8708C}'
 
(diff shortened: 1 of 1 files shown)


elyscape:
Actually, hold a sec.  Lemme make one more change.

elyscape:
Okay, we're good now.

elyscape:
@dscho Split into three commits.  How's that lok?

dscho:
Very good, thank you! I merged this already, even before the Unicode question is addressed, because it is easier to answer that question when I can simply build an installer to test with.

dscho:
I have *terrible* news:

> `%w` expands to the file's containing directory and `%L` expands to the long name of the file.

This does not work at all in my test on XP. @elyscape Where is this documented?

dscho:
`%L` works, but `%w` does not.

The excellent news is that we do not need the directory at all: we can just pass the `--no-cd` option and be done. And that is what I did: 9eddd2b

The *even better* news is that I could test with non-ASCII file names, for the script as well as for the drag-n-dropped files, and it *works*! (My intention was to verify [the claim that the WSH Drop Handler cannot handle non-ASCII file names](http://answers.microsoft.com/en-us/windows/forum/windows8_1-files/wshvbs-drophandler-doesnt-support-unicode/6c74a306-0655-489a-a22c-99909bf74a02) and I can now reject it because even the oldest Windows version I want to support -- Windows XP -- can handle it alright.)

elyscape:
> `%L` works, but `%w` does not.

Hm.  I guess they must have added it in Vista.  The documentation for command variable expansion for file type handlers is incredibly lacking.

PR #52, opened 2015-04-30
installer: Invoke bash directly for post-install

elyscape:
Invoking `git-bash.exe` to run the post-install scripts results in a
mintty window appearing, even when invoked with `SW_HIDE`. Invoking bash
directly avoids this.

Fixes git-for-windows/git#126.


dscho:
Heh, just when I was about to push my identical commit, I stumbled over this PR... ;-)

PR #55, opened 2015-05-06
Create README.md

kasal:
This is copied from top of https://github.com/msysgit/msysgit/blob/master/README.md

I was expecting to find the build instruction in this place (instead of the home page/Contribute).  Perhaps this will help someone else as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,5 @@
+# Build environment for Git for Windows
+
+This is Git for Windows SDK, the build environment for [Git for Windows](http://git-for-windows.github.io/).
+
+The easiest way to install Git for Windows SDK is via the [net installer](https://github.com/git-for-windows/build-extra/releases/latest). This installer will clone our [repositories](http://github.com/git-for-windows/), including all the necessary components to build Git for Windows, and perform an initial build.


dscho:
Thanks! I will touch it up in a moment because `build-extra` is actually a little bit more than just the Git SDK installer :smiley_cat:.

dscho:
@kasal could you have a look at the current README, please?

kasal:
I have looked at the top, the install instructions for the impatient, and it looks fine, with one exception, see below.
I was not patient enough to read more.

One issue: the readme says "It will also install a shortcut to the Git SDK Bash on the desktop."
IIRC, I have not got the link.

Actually, I was a bit confused with this and lost some time with it:
1) the install process opens a bash window at the end: it seems to be a mintty with the violet `Msys2.ico`
I tried to pin this to the taskbar, or otherwise reproduce this terminal window, but I failed.
2) So I created a shortcut to git-bash.exe, similar to the link created by the installer of the regular Git for Windows (pre)release.  But that is a different terminal window.  Which is better?

So there is a space for improvement of the default install, that could help people that know about Windows as little as I do.

dscho:
Thanks for the feedback. For the Git SDK, however, I thought I'd not need to polish it so much because developers are more likely to figure out why there is a second window popping up. BTW the shortcut was created for me very reliably, even when I recently installed a fresh Git SDK into a 32-bit Windows XP VM...

PR #57, opened 2015-05-08
add tar to list of packaged GNU utilities

pdelre:
As per the discussion in https://groups.google.com/forum/#!topic/git-for-windows/jerxhdoS-20.

Adding tar in the list of packaged GNU utilities. This to increase compatibility with users coming from msysgit 1.x and causes a slight increase in filesize.

Comparing the Portable built package:

Build | Installer size (kb) | Unpacked size (kb)
------ | ---------------------- | ----------------------
Without tar | 48720 | 400196
With tar | 49094 | 402516

### DIFF
--- a/make-file-list.sh
+++ b/make-file-list.sh
@@ -28,7 +28,7 @@ pacman_list () {
 
 pacman_list mingw-w64-$ARCH-git mingw-w64-$ARCH-git-doc-html \
 	git-extra ncurses mintty vim openssh winpty \
-	sed awk less grep gnupg findutils coreutils \
+	sed awk less grep gnupg tar findutils coreutils \
 	dos2unix which subversion mingw-w64-$ARCH-tk "$@" |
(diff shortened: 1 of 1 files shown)


dscho:
This information is really helpful. May I ask you to amend the commit message with it? Maybe before the link a tiny summary of the discussion (just in case somebody reads the commit message while stuck in a place without internet connection), e.g. "The `tar` utility was included in Git for Windows 1.x, and this developer wrote scripts that rely on it (see discussion in ...).".

Also, could you please [sign off](https://github.com/git/git/blob/v2.3.5/Documentation/SubmittingPatches#L234-L286) your patch?

Thanks!

pdelre:
I've updated the commit message for review.

I believe I signed off on the commit, but please let me know if it isn't correct.

dscho:
Perfect. Thanks for taking the time to make this change. The next Git for Windows preview will ship with `tar`!

pdelre:
Awesome. Thank you for the support.

PR #58, opened 2015-05-12
installer: Remove the radio button group for Windows Explorer integr…

sschuberth:
…ation

Now that we do not ship git-cheetah anymore there is only one choice
left, thus we do not need a radio button group.


dscho:
Thanks.

PR #59, opened 2015-05-14
Complete on Shift-Tab

orgads:
Preserve MsysGit behavior


orgads:
See https://github.com/Alexpux/MSYS2-packages/issues/247

dscho:
Thanks!

For future record, it would be nice to include information such as "If you want to tab-complete after typing an upper-case letter, it would be nice not having to release the Shift key first." as mentioned in the ticket. It is not *necessary*, of course, but gives a nice rationale for the patch. As I said, just for future record.

PR #61, opened 2015-05-28
Include `diff.exe` into the installers

dscho:
It is required by `git mergetool`.

This fixes https://github.com/git-for-windows/git/issues/163


dscho:
Oh well, I thought about it and it is obvious enough that this is an improvement.

PR #62, opened 2015-06-01
installer: Do not be silent when uninstalling a previous version

sschuberth:
It is bad practice to remove files from the user's system without any
confirmation. Especially users who just want to give the new MSYS2-based
Git a try most likely want to keep their existing MSYS1-based Git
installation. Also, even just installing the 64-bit version of GfW 2.x
removes Gfw 1.x, which is 32-bit, preveting to have a 32-bit and 64-bit
version of Git installed in parallel, which again is useful for testing.

### DIFF
--- a/installer/install.iss.in
+++ b/installer/install.iss.in
@@ -1168,13 +1168,14 @@ begin
             end;
         end;
 
-        if RegQueryStringValue(Domain,Key,'UninstallString',UninstallString) then
+        if RegQueryStringValue(Domain,Key,'UninstallString',UninstallString) then begin
             // Using ShellExec() here, in case privilege elevation is required
-            if not ShellExec('',UninstallString,'/SILENT /NORESTART /SUPPRESSMSGBOXES','',SW_HIDE,ewWaitUntilTerminated,ErrorCode) then begin
+            if not ShellExec('',UninstallString,'/NORESTART','',SW_SHOW,ewWaitUntilTerminated,ErrorCode) then begin
                 Msg:='Could not uninstall previous version. Trying to continue anyway.';
                 SuppressibleMsgBox(Msg,mbError,MB_OK,IDOK);
                 Log(Msg);
             end;
+        end;
     end;
 end;
 


dscho:
First of all, the uninstaller is not *really* silent. It still pops up a window when uninstalling the previous version.

Second, we already had the discussion whether it should be allowed to install two different Git for Windows versions at the same time and concluded that this should not be the case.

Therefore I am closing this Pull Request.

sschuberth:
> It still pops up a window when uninstalling the previous version.

That's not true as the original code contains the SW_HIDE window flag. My old version of Git for Windows was just silently removed without any window popping up.

> Second, we already had the discussion whether it should be allowed to install two different Git for
> Windows versions at the same time and concluded that this should not be the case.

I'm not questioning that for the average user it makes sense to keep only one version, and that removing the old version by default makes somewhat sense. But not without a chance to opt-out of that default behavior. And until there's a way to opt-out, I strongly vote for not unsolicitedly deleting any files.

dscho:
> > we already had the discussion whether it should be allowed to install two different Git for
> > Windows versions at the same time and concluded that this should not be the case.
> 
> I'm not questioning that for the average user it makes sense to keep only one version, and that removing the old version by default makes somewhat sense. But not without a chance to opt-out of that default behavior. And until there's a way to opt-out, I strongly vote for not unsolicitedly deleting any files.

We are not deleting any random files here. We are removing/overwriting obsolete files. This is consistent with Git for Windows 1.x, or for that matter, any sane software project I know of.

sschuberth:
You're comparing a GfW 1.x to GfW 1.y (with x < y) upgrade to a GfW 1.i to GfW 2.k upgrade, which are totally different things. As GfW 2.k is a complete rewrite, it's effectively a new application. Just for convenience we're still using the same `AppId` in Inno Setup.

I know that you once had the misunderstanding that a GfW 1.x to GfW 1.y upgrade would also uninstall GfW 1.x before GfW 1.y is installed. This is not the case. Instead, the files contained in the CfW 1.y installer are just copied to the previously existing directory, overwriting any files. In particular, files that were part of GfW 1.x but are not anymore part of GfW 1.y are *not* removed as part of the upgrade. Same goes for files that the user / admin might have manually copied to the installation directory.

I'm explaining this because a GfW 1.x to GfW 1.y upgrade would not make the user lose any files because no uninstaller is run.

Now with how the GfW 1.i to GfW 2.k upgrade is currently implemented, the user might very well lose files he's still interested in as the uninstaller *is* run.

So no, this is *not* consistent with Git for Windows 1.x upgrades.

dscho:
Git for Windows 2.x is an upgrade of Git for Windows 1.x, for all practical matters. So no, it is not a totally different application. (I will not confuse implementation details with semantic meaning.)

> Now with how the GfW 1.i to GfW 2.k upgrade is currently implemented, the user might very well lose files he's still interested in as the uninstaller is run.

First of all, I very much welcome female users. You will therefore never hear me refer to a Git for Windows user as a "he".

Second, we always overwrote files without asking. By the quoted reasoning that was inappropriate, but I disagree. When a user wants to have the newest Git for Windows, they simply need to download the newest installer and run it. As easy as that. That's it. And that is how I will keep it.

sschuberth:
> Second, we always overwrote files without asking.

I'm not talking about overwriting files for which an update is available. I'm talking about unsolicitedly deleting files for which no update is available.

> And that is how I will keep it.

Ok, that saves me work to further try to convince you with valid arguments.

dscho:
Let's keep the tone professional.

> > Second, we always overwrote files without asking.
> 
> I'm not talking about overwriting files for which an update is available. I'm talking about unsolicitedly deleting files for which no update is available.

When adding a file that was not installed by the InnoSetup installer, this file will *not* be deleted when uninstalling Git for Windows. That behavior has not changed.

When modifying a file that *was* installed by the InnoSetup installer, this file *will* be deleted when uninstalling Git for Windows and it *will* be overwritten when upgrading Git for Windows, whether we uninstall first or not. In other words, files that InnoSetup believes are under its control are going to be clobbered if the user decided to modify them. That behavior has not changed.

sschuberth:
> Let's keep the tone professional.

Would you mind practicing what you preach? I wasn't the one giving this the turn of a gender discussion with totally unrelated and unhelpful remarks.

I'll stop here as I've nothing more to add. I cannot help people understand who do not *want* to understand.

PR #63, opened 2015-06-15
Documented --cd, --no-cd and the HOME environment variable

sukiletxe:


sukiletxe:
Please, could you specify the line numbers, rather than adding comments? I'm totally blind and unfortunately I can't read where your comments belong to.
Also, I'm quite newbie to github, is it okay if I use git commmit --ammend?
Thanks.

sukiletxe:
Probably this clarification is not necessary, but just in case: I mean, could you specify the line numbers that need to be changed in the comment body, rather than adding line comments?

sukiletxe:
Okay, I was looking in the wrong place for the line comments. I can see them and I will fix them in a few hours.
Thanks.

sukiletxe:
I've fixed those issues.
Thanks.

dscho:
> Please, could you specify the line numbers, rather than adding comments? I'm totally blind and unfortunately I can't read where your comments belong to.

Oops, sorry! I will try to quote better in the future, thanks for pointing that out.

> is it okay if I use git commmit --ammend?

Yes, the rule of thumb in the Git for Windows project is that `master` must *never* be rewritten (which `-amend` does, it rewrites commits), but topic branches are fair game.

dscho:
Thanks!

PR #72, opened 2015-08-20
git-extra: add git-prompt check to bash.bashrc

nalla:
If `bash` is called without the `--login` function, the `git-prompt.sh` in
the `profile.d` directory will not get sourced. Lets fix this behavior by
explicitly checking if this bash instance was called without the `--login`
argument and if so call the `git-prompt.sh`.

This fixes git-for-windows/git#290

Signed-off-by: 마누엘 <nalla@hamal.uberspace.de>


dscho:
Awesome contribution. Thanks so much, again!

PR #73, opened 2015-08-21
connect: include new connect package

nalla:
Lets include the newly created package in the installer. Note that the
huble maintainer needs to build, release and install that package so that
it will indeed be picked up.

This fixes git-for-windows/git#293

### DIFF
--- a/make-file-list.sh
+++ b/make-file-list.sh
@@ -29,7 +29,8 @@ pacman_list () {
 pacman_list mingw-w64-$ARCH-git mingw-w64-$ARCH-git-doc-html \
 	git-extra ncurses mintty vim openssh winpty \
 	sed awk less grep gnupg tar findutils coreutils diffutils \
-	dos2unix which subversion mingw-w64-$ARCH-tk "$@" |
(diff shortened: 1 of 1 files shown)


dscho:
Perfect. Since I included the documentation in the package, I guess we now add 79kB + 42kB + 21 kB uncompressed, probably something like ~30kB compressed. Not too bad.

dscho:
The not so perfect thing is that it does not work because Bintray is not working properly. I uploaded the packages, and you can see them: https://bintray.com/git-for-windows/pacman/mingw-w64-x86_64-connect/view#files and https://bintray.com/git-for-windows/pacman/mingw-w64-i686-connect/view#files. However, when trying to download those files, one is greeted by the message: "The requested path was not found." I already reported the bug to Bintray, but I have no idea when I will hear back.

nalla:
Well, it is not listed here: https://dl.bintray.com/git-for-windows/pacman/x86_64/

dscho:
Seems the Bintray issues are resolved now.

PR #75, opened 2015-09-13
Convert pdf, doc and docx files to text by default

rimrul:
Converting PDF and Word files to text before diffing them allows an easier comparison between changed files. This reintroduces some functionality of Git for Windows 1.x.  It was requested by user @asdqwezx to reintroduce this feature in [#355](https://github.com/git-for-windows/git/issues/355). This pull request is not yet ready to be merged, there are still slight Issues with the doc conversion and the pdf conversion is yet to be reintroduced. All "new" files are copied out of the old msysgit repositories, and some have been slightly altered to work with Git 2.x.

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -18,6 +18,8 @@ pkgver() {
 build() {
   test $startdir/$pkgname.install -nt $startdir/$pkgname.install.in ||
   sed -e "/^@@GITCONFIG@@$/r $startdir/gitconfig" -e "/^@@GITCONFIG@@$/d" \
+	-e "/^@@GITATTRIBUTES@@$/r $startdir/gitattributes" \
+	-e "/^@@GITATTRIBUTES@@$/d" \
 	<$startdir/$pkgname.install.in >$startdir/$pkgname.install
 
   gcc -o create-shortcut.exe $startdir/create-shortcut.c -luuid -lole32
@@ -50,4 +52,5 @@ package() {
   install -m755 $startdir/bash_profile.sh $pkgdir/etc/profile.d
   install -m644 $startdir/msys2-32.ico $pkgdir/usr/share/git
   install -m644 $startdir/99-post-install-cleanup.post $pkgdir/etc/post-install
+  install -m755 $startdir/astextplain $pkgdir/$mingwdir/bin
 }
--- a/git-extra/astextplain
+++ b/git-extra/astextplain
@@ -1,0 +1,30 @@
+#!/bin/sh -e
+# minimalistic replacement for `run-mailcap --action=cat <file>`
+
+if test "$#" != 1 ; then
+	echo "Usage: astextplain <file>" 1>&2
+	exit 1
+fi
+
+# XXX output encoding (UTF-8) hardcoded
+case "$1" in
+	*.doc | *.DOC | *.dot | *.DOT)
+		antiword -m UTF-8 "$1" | sed "s/\^M$//" || cat "$1"
+		;;
+	*.docx | *.DOCX)
+		docx2txt.pl "$1" -
+		;;
+	*.pdf | *.PDF)
+		pdftotext -layout "$1" -enc UTF-8 - | sed "s/(\^M$)|(^\^L)//"
+		;;
+	# TODO add rtf support
+	*.rtf | *.RTF)
+		cat "$1"
+		;;
+	*)
+		echo "E: unsupported filetype $1" 1>&2
+		exit 1
+		;;
+esac
+
+exit 0
--- a/git-extra/git-extra.install.in
+++ b/git-extra/git-extra.install.in
@@ -8,6 +8,12 @@ post_install () {
 		cat > /$dir/etc/gitconfig <<\GITCONFIG
 @@GITCONFIG@@
 GITCONFIG
+		test ! -d /$dir ||
+		test -f /$dir/etc/gitattributes ||
+		cat > /$dir/etc/gitattributes <<\GITATTRIBUTES
+@@GITATTRIBUTES@@
+
+GITATTRIBUTES
 	done
 
 	grep -q '^db_home: env windows' /etc/nsswitch.conf ||
--- a/git-extra/gitattributes
+++ b/git-extra/gitattributes
@@ -1,0 +1,10 @@
+*.doc	diff=astextplain
+*.DOC	diff=astextplain
+*.docx	diff=astextplain
+*.DOCX	diff=astextplain
+*.dot	diff=astextplain
+*.DOT	diff=astextplain
+*.pdf	diff=astextplain
+*.PDF	diff=astextplain
+*.rtf	diff=astextplain
+*.RTF	diff=astextplain
--- a/make-file-list.sh
+++ b/make-file-list.sh
@@ -28,14 +28,15 @@ pacman_list () {
 }
 
 # Packages that have been added after Git SDK 1.0.0 was released...
-pacman -S --needed --noconfirm mingw-w64-$ARCH-connect git-flow >&2 ||
+pacman -S --needed --noconfirm mingw-w64-$ARCH-connect git-flow unzip docx2txt \
+	mingw-w64-$ARCH-antiword mingw-w64-$ARCH-xpdf >&2 ||
 die "Could not install required packages"
 
 pacman_list mingw-w64-$ARCH-git mingw-w64-$ARCH-git-doc-html \
 	git-extra ncurses mintty vim openssh winpty \
 	sed awk less grep gnupg tar findutils coreutils diffutils patch \
 	dos2unix which subversion mingw-w64-$ARCH-tk \
-	mingw-w64-$ARCH-connect git-flow "$@" |
+	mingw-w64-$ARCH-connect git-flow docx2txt mingw-w64-$ARCH-antiword "$@" |
 grep -v -e '\.[acho]$' -e '\.l[ao]$' -e '/aclocal/' \
 	-e '/man/' -e '/pkgconfig/' -e '/emacs/' \
 	-e '^/usr/lib/python' -e '^/usr/lib/ruby' \
@@ -71,7 +72,8 @@ grep --perl-regexp -v -e '^/usr/(lib|share)/terminfo/(?!.*/(cygwin|dumb|xterm.*)
 sed 's/^\///'
 
 test -z "$PACKAGE_VERSIONS_FILE" ||
-pacman -Q filesystem dash rebase util-linux >>"$PACKAGE_VERSIONS_FILE"
+pacman -Q filesystem dash rebase util-linux unzip \
+	mingw-w64-$ARCH-xpdf >>"$PACKAGE_VERSIONS_FILE"
 
 cat <<EOF
 etc/profile
@@ -89,4 +91,9 @@ usr/bin/dash.exe
 usr/bin/rebase.exe
 usr/bin/rebaseall
 usr/bin/getopt.exe
+mingw$BITNESS/bin/astextplain
+mingw$BITNESS/etc/gitattributes
+usr/bin/unzip.exe
+mingw$BITNESS/bin/pdftotext.exe
+mingw$BITNESS/bin/libstdc++-6.dll
 EOF


rimrul:
Should we upgrade to the current [docx2txt v1.4](http://sourceforge.net/projects/docx2txt/files/docx2txt/v1.4/docx2txt-1.4.tgz/) while we are changing this?

dscho:
It is a very good start!

The `antiword` program should be packaged as a package, though. Adding it [as a binary without source code](https://github.com/rimrul/build-extra/blob/e5a81441392a749c30b174377decf03b3ff88300/git-extra/antiword.exe) is inappropriate.

You will find excellent documentation on Git for Windows' package management [on the wiki](https://github.com/git-for-windows/git/wiki/Package-management), and there are pretty good examples in both `MSYS2-packages`' and `MINGW-packages`' commit history how a new package can be added, e.g. [`mingw-w64-connect`](https://github.com/dscho/MINGW-packages/commit/b7d8db389d03b001ca16c15dab8807c57d486a45).

I have no idea whether `antiword` can be compiled in default mode (i.e. whether it is a MinGW program), or whether it has to be compiled in an MSys shell (i.e. whether it is an MSys2 program); This would be the indicator whether the package needs to go into `MSYS2-packages` or `MINGW-packages`.

After you have the package, you do not need to list every single of `antiword`'s files in `make-file-list.sh` but instead simply list the package.

Listing single files (and adding the information about the corresponding package's version) is something I would do only for packages that ship way too many things, where we want to cherry-pick only a few files.

I have no idea where the `docx2txt` comes from, but I suspect that we would want to add a new package for that as well, of course using the current version.

rimrul:
Yes, I'll test how I can get `antiword` compiled from source, the author mentions djgpp or cygwin as ways to build it, but hasn't tested it himself. `docx2txt` seems to have its source (it's a perl script) uploaded to sourceforge by the original author. Our versions of `antiword` and `docx2txt` are both from the old [msysgit repository](https://github.com/msysgit/msysgit/tree/master/bin) where antiword was just the same inappropriate binary without source.

dscho:
> Yes, I'll test how I can get antiword compiled from source, the author mentions djgpp or cygwin as ways to build it, but hasn't tested it himself.

It is typically fairly easy to get it to build in MSys2. Just try it, download the source, extract it, run `./configure` (or `make` or whatever is the recommended way to build it).

With that information, the `PKGBUILD` should be really easy to write (or even better: just take the `PKGBUILD` from the `mingw-w64-connect` example I linked earlier, go through it line by line and adjust what looks like it needs to be adjusted).

> Our versions of antiword and docx2txt are both from the old msysgit repository where antiword was just the same inappropriate binary without source.

Yeah, my mistake. I should not have accepted that... :innocent:

dscho:
> docx2txt seems to have its source (it's a perl script) uploaded to sourceforge by the original author.

Okay, let's package that just the same like Markdown (which is also a Perl script): https://github.com/Alexpux/MSYS2-packages/blob/master/markdown/PKGBUILD

rimrul:
I created a kind of working docx2txt package. When built with`makepkg-mingw -s`, it installs files to 
/usr/src/MINGW-packages/mingw-w64-docx2txt/pkg/mingw-w64-x86_64-docx2txt/mingw64/bin and
/usr/src/MINGW-packages/mingw-w64-docx2txt/pkg/mingw-w64-x86_64-docx2txt/mingw64/etc that I intended to install to /mingw64/bin and /mingw64/etc. Is this normal `makepkg-mingw -s`-behaviour or should I remove the `$pkgdir` from the call to `make`? Can I use this package to make an installer already or do I need to create a pull request and wait for it to be merged before testing the new package in the installer? Where should I open the pull request? On the GfW repo or on the base repo?

dscho:
> I created a kind of working docx2txt package.

Could you fork MINGW-packages and push your topic branch?

dscho:
> When built withmakepkg-mingw -s, it installs files to
> /usr/src/MINGW-packages/mingw-w64-docx2txt/pkg/mingw-w64-x86_64-docx2txt/mingw64/bin and
> /usr/src/MINGW-packages/mingw-w64-docx2txt/pkg/mingw-w64-x86_64-docx2txt/mingw64/etc that I intended to install to /mingw64/bin and /mingw64/etc. Is this normal makepkg-mingw -s-behaviour or should I remove the $pkgdir from the call to make?

The idea is that `makepkg-mingw` "installs" it into the `pkg/` subdirectory, indeed, and then builds the `mingw-w64-<arch>-doc2txt-<version>.pkg.tar.xz` package from that subdirectory. That `.pkg.tar.xz` file should then be installed for real, via `pacman -U <file>`.

The only suspicious thing there is the extra `/mingw-w64-x86_64-docx2txt/` level before `/mingw64/bin`; that should not be the case.

> Can I use this package to make an installer already or do I need to create a pull request and wait for it to be merged before testing the new package in the installer? Where should I open the pull request? On the GfW repo or on the base repo?

You could already make the installer, but I think that I would try to focus on building the package first.

BTW since it is a Perl script, i.e. it requires `perl` (which is an MSys2 package), I would put it into the `MSYS2-packages` repository and call it `docx2txt` rather than `mingw-w64-docx2txt`, but those are really only details that are *easily* remedied once you have a working `PKGBUILD`.

rimrul:
>  Could you fork MINGW-packages and push your topic branch?

[Done.](https://github.com/rimrul/MINGW-packages/tree/docx2txt)

>BTW since it is a Perl script, i.e. it requires perl (which is an MSys2 package), I would put it into the MSYS2-packages repository and call it docx2txt rather than mingw-w64-docx2txt, but those are really only details that are easily remedied once you have a working PKGBUILD.

Yes we should do that, sounds like that shouldn't be too hard.

dscho:
> > Could you fork MINGW-packages and push your topic branch?
> 
> Done.

Perfect, thanks. I adjusted it so it is a proper MSys2 package (since it depends on an MSys2 package, it is appropriate):

https://github.com/Alexpux/MSYS2-packages/compare/master...dscho:docx2txt

Please note that you need to open an *MSys2* shell (i.e. launch `C:\git-sdk-64\msys2_shell.bat`) before building, and you need to change directory to `/usr/src/MSYS2-packages/docx2txt/` (after initializing `/usr/src/MSYS2-packages/` and fetching my `docx2txt` branch), and the build is started using `makepkg -s` (i.e. not `makepkg-mingw -s`).

The resulting package looks pretty good to me, but I did not test anything.

The only thing I noticed is that there are now two scripts: `docx2txt.pl` and `docx2txt.sh`... Maybe rename the latter to `docx2txt`?

rimrul:
>The only thing I noticed is that there are now two scripts: docx2txt.pl and docx2txt.sh... Maybe rename the latter to docx2txt?

The .sh file is just a wrapper around the .pl file that always outputs a textfile. Since diff needs stdout output for its textconv we should keep the perl script. If you want to open a pull request to the base repo I would keep the files with their endings, since they are the output of the makefile, which is the recommended way to install `docx2txt`. I would just attend .pl to the call in `astextplain`.

>The resulting package looks pretty good to me, but I did not test anything.

Mine (still a MINGW-Package, but that shouldn't be a big diffference here) doesn't contain a `mingw-w64-x86_64-docx2txt/` folder inside the tar.xz file so that shouldn't cause any problems. I'll test the installation with pacman.

>>docx2txt seems to have its source (it's a perl script) uploaded to sourceforge by the original author.

>Okay, let's package that just the same like Markdown

the only significant difference I see in the Markdown `PKGBUILD` is a `pod2man` call. Since `docx2txt` contains no POD we won't find anything to convert to manpages.

dscho:
Oh, and I just noticed a suspicious line:

```
Setting systemConfigDir to [...<wrong path>...]
```

where the <wrong path> points to *inside* the `pkg/` directory...

> > The only thing I noticed is that there are now two scripts: docx2txt.pl and docx2txt.sh... Maybe rename the latter to docx2txt?
> 
> The .sh file is just a wrapper around the .pl file that always outputs a textfile. Since diff needs stdout output for its textconv we should keep the perl script.

Makes sense, in particular since the `.sh` file seems to require the Perl script anyway.

> If you want to open a pull request to the base repo I would keep the files with their endings, since they are the output of the makefile, which is the recommended way to install docx2txt. I would just attend .pl to the call in astextplain.

Hmm. I still would like to rename it because it would make it easier to use (if a random user installs the `docx2txt` package and sees that there are both a `.pl` and a `.sh` file, they might be confused -- as I was -- which one to use).

> > The resulting package looks pretty good to me, but I did not test anything.
> 
> Mine (still a MINGW-Package, but that shouldn't be a big diffference here) doesn't contain a mingw-w64-x86_64-docx2txt/ folder inside the tar.xz file so that shouldn't cause any problems. I'll test the installation with pacman.

Thank you.

However, I guess that the `CONFIGDIR` needs to be set to `/etc/` in order for the test to be truly successful: it will probably work for you because you have that `pkg/` directory, but if you install the Pacman package on some other machine, it will probably fail.

> > > docx2txt seems to have its source (it's a perl script) uploaded to sourceforge by the original author.
> 
> > Okay, let's package that just the same like Markdown
> 
> the only significant difference I see in the Markdown PKGBUILD is a pod2man call. Since docx2txt contains no POD we won't find anything to convert to manpages.

Yeah, I agree. The only documentation is the usage, and there is no need to convert that into a man page.

dscho:
I just figured out that leaving off the `CONFIGDIR` setting should fix it ;-) Force-pushed.

rimrul:
>However, I guess that the CONFIGDIR needs to be set to /etc/ in order for the test to be truly successful: it will probably work for you because you have that pkg/ directory, but if you install the Pacman package on some other machine, it will probably fail.

Your prediction was right, it configured the CONFIGDIR wrong. 



>>If you want to open a pull request to the base repo I would keep the files with their endings, since they are the output of the makefile, which is the recommended way to install docx2txt. I would just attend .pl to the call in astextplain.

>Hmm. I still would like to rename it because it would make it easier to use (if a random user installs the docx2txt package and sees that there are both a .pl and a .sh file, they might be confused -- as I was -- which one to use).

Since we need neither the shell script nor the config file (just tested the installed v1.4 with `$systemConfigDir = "/";`) how about we drop the make call and just copy the perl script without the extension?

rimrul:
Investigated the antiword build. The following commands where needed:

    rm Makefile
    mv Makefile.Linux Makefile
    make install

antiword built into `$HOME/bin` and intalled its mapping files into `$HOME/.antiword`.
`$home/bin` contains `antiword.exe` and `kantiword`. `kantiword` is some script to make drag and drop work in KDE. I hope there are no plans to support running KDE (or gnome or LXDE etc.) from Git bash. `$HOME/.antiword` contains the mappingfiles I pulled from the msysgit repo plus a `Default` file that seems to map MS Word fonts to RiscOS fonts, an `Example` file that contains a slightly different mapping to RiscOS fonts,  a `fontnames.russian` file that maps cyrillic Word fonts to corresponding PostScript fonts and two files `UnicodeXX` that are variations of the corresponding `8859-XX.txt` files for RiscOS.

dscho:
> Since we need neither the shell script nor the config file (just tested the installed v1.4 with $systemConfigDir = "/";) how about we drop the make call and just copy the perl script without the extension?

In the interest of making a package that is as useful as possible, I would rather install the whole thing; That would make MSys2 benefit most from your work, not just Git for Windows.

> Investigated the antiword build. The following commands where needed:
> 
> rm Makefile
> mv Makefile.Linux Makefile
> make install

Good catch!

You could also call

```sh
make -f Makefile.Linux install
```

instead.

> I hope there are no plans to support running KDE (or gnome or LXDE etc.) from Git bash.

LOL... no, there isn't ;-)

Good work!

rimrul:
>In the interest of making a package that is as useful as possible, I would rather install the whole thing; That would make MSys2 benefit most from your work, not just Git for Windows.

That's what I originally intended to achieve. But removing the extension of the perl script would break the shell script. We could regex search and replace inside the shell script using perl or sed, though.
`/etc` should usually work as the `$systemConfigDir`.

>You could also call
make -f Makefile.Linux install
instead.

So we should end up with something like

    build()
    {
        cd $srcdir
        make -f Makefile.Linux antiword
    }
    prepare()
    {
        install $srcdir/Resources/* $pkgdir/$mingwdir/share/antiword
        install $srcdir/antiword.exe $pkgdir/$mingwdir/bin
    }

dscho:
> In the interest of making a package that is as useful as possible, I would rather install the whole thing; That would make MSys2 benefit most from your work, not just Git for Windows.
> 
> That's what I originally intended to achieve. But removing the extension of the perl script would break the shell script.

Oh sorry! I was unclear. Since the `.sh` script is clearly the one that is intended to be the end-user script (and the Perl script being the work-horse that you can also call from elsewhere), I was only talking about the `.sh` extension to be stripped.

I did that and it looks pretty good, a simple test did exactly what I hoped it would do: OneDrive's "Getting started" document (which is a `.docx`) got translated as I wished.

Are you okay with these changes? https://github.com/Alexpux/MSYS2-packages/commit/4fd937cb7754c9961bea1b36d29246abe407d2e0 If so, I would squash those changes in and submit the build recipe upstream.

> So we should end up with something like
> 
> build()
> {
>     cd $srcdir
>     make -f Makefile.Linux antiword
> }
> prepare()
> {
>     install $srcdir/Resources/* $pkgdir/$mingwdir/share/antiword
>     install $srcdir/antiword.exe $pkgdir/$mingwdir/bin
> }

Yes. You probably also need to pass `-D` to the `install` executable so that the directories are created.

Oh, and I just saw that in other `PKGBUILD` scripts, the common invocation is:

```sh
install -D -m755 <something>.exe "${pkgdir}"${MINGW_PREFIX}/bin/<something>.exe
```

i.e. the `MINGW_PREFIX` environment variable is already available, and I guess upstream likes curly brackets :grinning:

rimrul:
>Are you okay with these changes? Alexpux/MSYS2-packages@4fd937c If so, I would squash those changes in and submit the build recipe upstream.

Seems fine to me. I'll write a `PKGBUILD` for antiword and generate an upstream pull request at [Alexpux/MINGW-packages](https://github.com/Alexpux/MINGW-packages)

dscho:
Cool!

dscho:
Whoopsie! Sorry, hit the wrong button :blush:

rimrul:
Ok, antiword is a mess. On http://www.winfield.demon.nl/linux/ it lists `antiword-0.37.tar.gz` as 310K. The version my Firefox downloads is reproducably 1.7M big. the version `makepkg-mingw` downloads is 310K. 7Zip opens both as valid files. It opens the big version in one step and the small one in two (a gz archive containing a tar archive)... This slight difference causes a checksum mismatch -.-

dscho:
> The version my Firefox downloads is reproducably 1.7M big.

Firefox probably "helpfully" uncompresses it.

dscho:
The `docx2txt` Pull Request was merged: https://github.com/Alexpux/MSYS2-packages/pull/345

rimrul:
I've opened [Alexpux/MINGW-packages#781](https://github.com/Alexpux/MINGW-packages/pull/781) for Antiword. We'll wait for that to be merged, and then I'll introduce those packages here. Enough done for today.

dscho:
> Enough done for today.

You did great! Thank you!

dscho:
And the `mingw-w64-antiword` package now hit MSys2's Pacman repository (but not the `docx2txt` package...)!

rimrul:
I've removed the `unzip`, `docx2txt` and `antiword` files from `git-extra/PKGBUILD` and `make-file-list.sh`. Should I add the `unzip`, `docx2txt` and `antiword` packages to line 30 (`pacman -S --needed`...), line 33(`pacman_list mingw-w64-$ARCH-git`...) or line 73 (`pacman -Q filesystem`) of `make-file-list.sh`? Would you prefer a commit on top of e5a8144, an ammended commit based on e5a8144 or all changes rebased into a single commit on top of d930be0c20? I'd say rebasing onto d930be0c20 would produce more precise results when meassuring the size difference.

dscho:
> I've removed the unzip, docx2txt and antiword files from git-extra/PKGBUILD and make-file-list.sh.

Good!

> Should I add the unzip, docx2txt and antiword packages to line 30 (pacman -S --needed...), line 33(pacman_list mingw-w64-$ARCH-git...) or line 73 (pacman -Q filesystem) of make-file-list.sh?

They need to be on line 30 for sure because the Git for Windows SDK 1.0.0 did not install them, and contributors who were early adopters should not suffer for their early help.

Whether you add them to line 33 or 73 depends on the size it adds. For example, if the `unzip` package contains much more than just the `unzip.exe` it makes sense to add *just* the file to line 91 and list the `unzip` package on line 73. If it does not really matter, installer size-wise, it is better to add the `unzip` package to line 33 and be done with it.

> Would you prefer a commit on top of e5a8144, an ammended commit based on e5a8144 or all changes rebased into a single commit on top of d930be0?

Yeah, a single commit (i.e. an amended e5a8144) rebased on top of current `master` would make most sense.

Thank you!

dscho:
BTW I made a couple of mistakes in the `docx2txt` package definition which @Alexpux has fixed gracefully, and he said that he would upload the packages today...

dscho:
And the `docx2txt` packages were uploaded! Yeah!

rimrul:
Just pushed ~~two~~ three updates: one update and two bugfixes for said update, the third one (c32af21) should be allright. Maybe I'll take a look at pdftotext tomorrow morning.

rimrul:
I've done the [`git rebase --whitespace=fix HEAD^`](https://github.com/rimrul/build-extra/commit/c32af212f327f728e1b5748f3ba05cddceedf307#commitcomment-13261846) and added the pdf function, but I have yet to check for form feed issues, rebase onto master and check  the size differences. We are definitly getting close to a good result.

dscho:
Very good! I am looking forward to merging this!

rimrul:
It seems, that removing `unzip.exe` increases the installer size by 215 bytes. I have no idea, how that works.

rimrul:
>It seems, that removing unzip.exe increases the installer size by 215 bytes. I have no idea, how that works.
Logical error here. Removing unzip from the rebased c6508e0 installer produces different results than adding it to the master installer...

dscho:
:-) Compression algorithms are strange, anyway.

rimrul:
Ok, I think I'm ready. I've solved all remaining problems, switched to xpdf and included the size differences into the commit message.

dscho:
Thank you! In my tests, the size difference was much more favorable, though: just 701kB. I adjusted the commit message (and I also adjusted the `make-file-list.sh` file because the portable Git cannot be built when there are duplicate entries: 7-Zip does not like that at all).

Thank you for your contribution! The bulk of the work was yours.

dscho:
[*When someone submits a pull request into your project.*](https://twitter.com/terminalbreaker/status/644526894549852161):

When someone submits a pull request into your project.

PR #77, opened 2015-09-16
git-extra: include some usefull sdk-aliases

nalla:
For housekeeping the _Git for Windows SDK_ installation it would be useful
to have some common pacman snipplets right there to be used. Lets
introduce those snipplets in the build-extra repository.

### DIFF
--- a/pacman-helper.sh
+++ b/pacman-helper.sh
@@ -1,11 +1,11 @@
 #!/bin/sh
 
 # This script helps Git for Windows developers to manage their Pacman
-# repository.
+# repository and their local pacman package-database.
 #
 # A Pacman repository is like a Git repository, but for binary packages.
 #
-# This script supports three commands:
+# This script supports seven commands:
 #
 # - 'fetch' to initialize (or update) a local mirror of the Pacman repository
 #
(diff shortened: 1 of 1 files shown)


dscho:
Whoops, I missed that you updated the pull request, sorry.

But now I have an additional idea, please bear with me for a moment. I use tab-completion heavily, and having a `pacman-mirror.sh` and a `pacman-housekeeper.sh` in the same directory would require a couple more keystrokes from my side than I would like. How about renaming `pacman-mirror.sh` to `pacman-helper.sh` and then just adding the new subcommands `disowned-dirs`, `disowned-files` and `remove-orphans` to the same script?

What do you think?

nalla:
Sounds like a plan. I'm on it.

dscho:
Thanks!

dscho:
Thanks @nalla! I unfortunately only get notified when a new comment is available, not when the branch was force-pushed. Would you terribly mind giving me a quick ping when you addressed my two last nit-picks?

nalla:
@dscho ping

dscho:
Perfect, thanks!

(Just out of curiosity: what was wrong with `echo` that it needed to be replaced by a `printf`?)

nalla:
mainly because of the `\n` in the usage string.
```
$ echo "test\ntest"
test\ntest

$ printf "test\ntest"
test
test
```

dscho:
Ah! Thanks!

PR #78, opened 2015-09-17
git-extra: add aliases to support interactive consoles in mintty

dscho:
This patch adds support for a couple of interactive consoles that users
invariably want to run in their Git Bash (despite our clear warning in the
release notes and the installer; Who reads text anyways?).

Of course, the list is non-exhaustive. The idea is that we can extend this
whenever needed. Or even better: users can extend this as needed and make
a Pull Request, in true Open Source style facilitated by... *drumroll*...
Git.

This fixes https://github.com/git-for-windows/git/issues/399 and
https://github.com/git-for-windows/git/issues/400

### DIFF
--- a/git-extra/aliases.sh
+++ b/git-extra/aliases.sh
@@ -4,3 +4,15 @@
 # --show-control-chars: help showing Korean or accented characters
 alias ls='ls -F --color=auto --show-control-chars'
 alias ll='ls -l'
+
+case "$TERM" in
+xterm*)
+	# The following programs are known to require a Win32 Console
+	# for interactive usage, therefore let's launch them through winpty
+	# when run inside `mintty`.
+	for name in node php php5 psql python2.7
+	do
+		alias $name="winpty $name.exe"
+	done
+	;;
+esac


dscho:
@harriha @radulemnaru please test.

harriha:
From https://github.com/git-for-windows/git/issues/399

> BTW can you think of more interactive interpreters than the ones I had listed?

Nope, cannot come up with any right now myself. `babel-node` is among the interactive consoles I tend to use, but that seems to work out of the box already.

harriha:
Replaced the `aliases.sh` in my `/etc/profile.d` dir with this one, node and psql working nicely.

dscho:
@harriha :+1: Thanks for being so responsive!

dscho:
I am merging this now, we can easily add more interpreter executable names as needed.

mikofski:
Thanks for these aliases, great idea! Although it caused me a little confusion, since I didn't have `psql` on my path, so I couldn't figure out why `psql` was showing up in the mintty shell and didn't occur to me to check the aliases. Are these aliases documented anywhere? I'm looking through the wiki, but I don't see any references to mintty, winpty or the interactive aliases. Would that be a good place to write something up? I can certainly contribute something if it is useful. Thanks!

nalla:
Well there is a mention in the release notes, see https://github.com/git-for-windows/git/releases/tag/v2.5.3.windows.1

dscho:
> I'm looking through the wiki, but I don't see any references to mintty, winpty or the interactive aliases. Would that be a good place to write something up?

@mikofski absolutely!

gmanfunky:
You can probably add ```docker``` to that list. Docker is interactive when given something like ```docker run -it ubuntu bash``` and the Docker project has recently started including git-for-windows in a quick-start toolbox.

I am suspecting this all kinda goes back to an assumption that STDIN is file-descriptor 0, which doesn't hold try on windows unless using some pty wrapper like ConEmu or winpty. Has anyone else been down in the details to find out?

dscho:
> You can probably add docker to that list

Could you research that a bit and then come back with a definitive statement? Thanks.

so0k:
I'm not sure if this is related, but I got rid of putty/pagent and started using ssh//ssh-agent bundled in git-for-windows. adding this line to my .bashrc fixed my ssh terminal troubles:

```bash
ssh='env TERM=linux ssh'
```

dscho:
@so0k indeed, it is completely unrelated. So you need to open a new ticket for it. With [enough information to allow others to help you](https://github.com/git-for-windows/git/wiki/Issue-reporting-guidelines).

so0k:
@dscho sure, just that it's not an issue and ssh works fine for me now.

It's more of a tip.. which I thought was related as it controls the behavior of the terminal similar to making a `winpty` alias to fix terminal behavior. But it seems each fix different issues.

dscho:
> It's more of a tip

@so0k ah, now that makes sense to me! Let's collect such tips in the [FAQ](https://github.com/git-for-windows/git/wiki/FAQ), okay?

moldstadt:
The auto aliasing for python was working fine for me and then suddenly stopped after aborting a postgresql installation.  I tried reinstalling git bash but it didn't help.  Had a look at this commit and it doesn't include an alias for python.exe, only python2.7.exe.  My executable is just python.exe.  Strangely it has been working fine until now.

Shouldn't python be added to the list of aliases?

dscho:
@moldstadt sure. How about patching it locally, then testing it (essentially `cd /usr/src/build-extra/git-extra && updpkgsums && makepkg-mingw -s && pacman -U $(ls -t git-extra-*.xz | head -n 1)` in the [Git for Windows SDK](https://git-for-windows.github.io/#download-sdk)), then contributing it as a Pull Request?

PR #79, opened 2015-09-18
post-install.bat: give 03-mtab a chance to write to the correct path

dscho:
Since we do not run the post-install scripts in a full login shell
anymore, `/etc/profile` has no chance of defining the `SYSCONFDIR`
before `03-tab.post` wants to use it.

So let's just hard-code it in `post-install.bat`.

This fixes https://github.com/git-for-windows/git/issues/405

### DIFF
--- a/post-install.bat
+++ b/post-install.bat
@@ -38,7 +38,7 @@
 )
 
 @REM Run the post-install scripts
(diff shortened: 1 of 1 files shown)


dscho:
Thanks for testing, @nalla!

rimrul:
@nalla was faster than me, but I can confirm this, too.

dscho:
Thanks @rimrul! It is always better to have two confirmations than one. I appreciate your help!

PR #81, opened 2015-09-22
installer: record version and some paths in the registry

dscho:
The newly introduces registry key

	HKEY_LOCAL_MACHINE\Software\GitForWindows

will now be created and it will contain three values: CurrentVersion,
InstallPath and LibexecPath.

The intention is to support third-party tools' installers so that e.g.
additional remote or credential helpers can easily be installed into the
correct location.

### DIFF
--- a/installer/install.iss.in
+++ b/installer/install.iss.in
@@ -97,6 +97,11 @@ SetupAppTitle={#APP_NAME} {#APP_VERSION} Setup
 SetupWindowTitle={#APP_NAME} {#APP_VERSION} Setup
 
 [Registry]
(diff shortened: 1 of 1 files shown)


dscho:
@linquize why don't you try it out ;-)

`{@MINGW_BITNESS}` refers to [either `mingw32` or `mingw64`](https://github.com/dscho/build-extra/blob/0fd86a3dc4604095a1dd23a6a3aa5ac579adb96e/installer/release.sh#L105), depending on the target CPU architecture.

`{@APP_VERSION}` is the version that is displayed in the installer's window bar, and it is also part of the file name of the installer.

PR #82, opened 2015-09-25
Only modify PATH if the user chose to do so.

brownp2k:
See https://github.com/git-for-windows/git/issues/438

### DIFF
--- a/installer/install.iss.in
+++ b/installer/install.iss.in
@@ -1440,15 +1440,15 @@ begin
     // Get the current user's directories in PATH.
     EnvPath:=GetEnvStrings('PATH',IsAdminLoggedOn);
 
-    // First, remove the installation directory from PATH in any case.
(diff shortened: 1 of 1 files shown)


dscho:
Oh, and if you test the newly-built installer twice, and it does indeed fix the bug then, please describe that in the commit message as well. This is valuable supporting information, therefore it needs to be recorded in the commit history.

brownp2k:
I've amended the commit message, as well as the commit itself to protect against the uninstall case, as you mentioned.  Assuming the previous install logged the PATH modification option, then uninstall can check that and only modify the path if the install option did so too.

This is all untested as I do not have the required build environment.  Since this is an uncompiled and untested commit, feel free to use this commit as a basis for your own changes as you see fit.

dscho:
> This is all untested

:-(

PR #84, opened 2015-10-01
installer (64-bit): remove 2GB pack size limit

tacker66:
In 64-bit setups it does not make sense to limit the pack size to 2GB,
so delete it from the installation defaults (which are not changeable
later anymore by git config).

This fixes https://github.com/git-for-windows/git/issues/288.

Signed-off-by: Thomas Ackermann <th.acker@arcor.de>

### DIFF
--- a/installer/install.iss.in
+++ b/installer/install.iss.in
@@ -1342,6 +1342,15 @@ begin
         end;
     end;
     if FileExists(ProgramData + '\Git\config') then begin
+#ifdef IS64
+        if not Exec(AppDir+'\{#MINGW_BITNESS}\bin\git.exe','config -f config --unset pack.packsizelimit',ProgramData+'\Git',SW_HIDE,ewWaitUntilTerminated,i) then begin
+            Msg:='Unable to unset packsize limit';
+
+            // This is not a critical error, so just notify the user and continue.
+            SuppressibleMsgBox(Msg,mbError,MB_OK,IDOK);
+            Log(Msg);
+        end;
+#endif
         Cmd:='http.sslCAInfo "' + AppDir + '/{#MINGW_BITNESS}/ssl/certs/ca-bundle.crt"';
         StringChangeEx(Cmd,'\','/',True);
         if not Exec(AppDir+'\{#MINGW_BITNESS}\bin\git.exe','config -f config '+Cmd,


dscho:
> > if the user chose a packsizelimit of, say, 16G, you still remove it. That is most distinctly not what I want. If the user changed the default, the value is not fair game any more.
> 
> But how you can tell that it's not the "2g" the user wants?

I can't. But if it is *different* from `2g` I *know* the user changed it. And I leave it alone.

> I tested the patch with a fresh install and several reinstalls on Win7 Prof 64bit. I didn't test different
> old values of packsizelimit but only that the value was correctly unset/deleted from the config file.

Thank you for the clarification.

> > BTW this part of the commit message is wrong:
> 
> > > (which are not changeable later anymore by git config).
> 
> Could you please explain, why this is wrong?

Yes. The patch you submitted in this Pull Request actually already explains the answer: you *can* change the value later via `git config`, and that is *exactly* what the installer will do: by passing the `-f C:\ProgramData\Git\config` parameter.

tacker66:
> Yes. The patch you submitted in this Pull Request actually already explains the answer: you can
> change the value later via git config, and that is exactly what the installer will do: by passing the -f 
> C:\ProgramData\Git\config parameter.

Of course, but there is no way to change it without knowing the name of the config file as in
'git config --system' and this brings me to another question: Why do we have this additional layer
of config data "above" '--system/--global/--local' in Git-for-Windows which is not present in
standard Git (and not documented in the 'git config' man page)?

dscho:
@tacker66 I could guess what you meant. But that's not what the commit message said.

dscho:
In any case, I got the impression that this here Pull Request does not really go into the direction I wanted it to go. Therefore my current plan is to just bite the bullet and finish the work I continued over at https://github.com/git-for-windows/build-extra/compare/2g-64-bit

tacker66:
> In any case, I got the impression that this here Pull Request does not really go into the direction I
> wanted it to go. Therefore my current plan is to just bite the bullet and finish the work I continued over 
> at https://github.com/git-for-windows/build-extra/compare/2g-64-bit.

That's OK for me but would you mind giving an answer to my last question?

dscho:
> Why do we have this additional layer
> of config data "above" '--system/--global/--local' in Git-for-Windows which is not present in
> standard Git (and not documented in the 'git config' man page)?

We do have that additional layer because with Git for Windows, the POSIX root directory is actually *not* a system-wide root directory. You can actually install five hundred different portable Gits on the same machine, and all of them will have *different* POSIX root directories! Therefore, `/etc/gitconfig` is *not actually* system-wide a config. We do need that system-wide config, though. Hence the additional layer.

Please note, though, that hiding this question (and forcing me to hide my answer) in this Pull Request is not a good use of my time. It would have been much superior an idea to ask on [the mailing list](https://groups.google.com/forum/#!forum/git-for-windows).

In any case, I merged my fix after testing it.

PR #86, opened 2015-10-20
installer: Change installer privileges to 'lowest'

kjeremy:
Allows the installer to be run with normal user permissions again.

Fixes git-for-windows/git#497

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -55,11 +55,7 @@ DefaultGroupName={#APP_NAME}
 DisableProgramGroupPage=auto
 DisableReadyPage=yes
 InfoBeforeFile={#SourcePath}\gpl-2.0.rtf
-#ifdef OUTPUT_TO_TEMP
 PrivilegesRequired=lowest
-#else
-PrivilegesRequired=poweruser
-#endif
 UninstallDisplayIcon={app}\{#MINGW_BITNESS}\share\git\git-for-windows.ico
 #ifndef COMPILE_FROM_IDE
 #if Pos('-',APP_VERSION)>0


kjeremy:
I just force pushed this with the suggested changes and did not realize that it would remove the previous discussion.  Sorry :(

dscho:
> I just force pushed this with the suggested changes and did not realize that it would remove the previous discussion. Sorry :

No harm done.

Instead, this is now an excellent Pull Request and it is fun to merge it.

dscho:
> [...force-push...] would remove the previous discussion.

It might have removed the original discussion, but it now lives on in your excellent commit message ;-)

XhmikosR:
FYI, this breaks upgrades if the previous version was installed with admin rights.

White-Tiger:
well, for me the previous uninstaller asked for admin rights and the new installation continued. Though it was stuck for unknown reasons at 100% installation and I had to kill it but that should be unrelated^^ *(some certificate setup sh was "running")*

What was/is your problem in more detail?

dscho:
> this breaks upgrades if the previous version was installed with admin rights.

@XhmikosR would you *please* provide more information so I have a prayer to understand what problem you are talking about?

XhmikosR:
```
---------------------------
Git 2.6.3 Setup
---------------------------
Setup was unable to create the directory "C:\Program Files\Git\tmp".



Error 5: Access is denied.
---------------------------
OK   
---------------------------

```

1) Have 2.6.2 installed elevated in Program Files
2) Run 2,6.3 installer

When it tries to install the new version after uninstalling the old one, the process isn't elevated hence the issue. The warning message doesn't show up in this specific case so the upgrade process isn't smooth. One has to run the installer again to actually see the message that they need to run elevated manually.

paladox:
This may be causing https://github.com/git-for-windows/git/issues/526

XhmikosR:
That is the same issue, yes.

dscho:
So: ouch. What a terrible mistake to merge this. My bad.

dscho:
Bah! I should *have known*!!! The `none` setting is *not* at all the same as the `lowest` one:

https://github.com/jrsoftware/issrc/blob/841a86c32c44a44789e88ad1b74328cfe9842fa7/Projects/Main.pas#L2939-L2940

As you can see, if the setting is `lowest`, InnoSetup will *never* elevate. But if it is `none`, it *will* elevate because then the `AEmulateHighestAvailable` flag is true: https://github.com/jrsoftware/issrc/blob/ea65d280c3157d154c07e653bb5b1d9cb2a11d40/Projects/SpawnServer.pas#L217

So: I will revert this change and put `none` again.

PR #89, opened 2015-11-19
Add script for building .tar.bz2 release

bastianeicher:
Adds an `archive/release.sh` script, similar to `portable/release.sh`. The resulting .tar.bz2 archive has all post-install steps pre-applied so the Git for Windows installation is ready for use right after extraction.

This is for issue git-for-windows/git#513

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@
 /portable/root/etc/
 /portable/root/post-install.bat
 /sdk-installer/root/
+/archive/root/*
+!/archive/root/etc
--- a/archive/release.sh
+++ b/archive/release.sh
@@ -1,0 +1,104 @@
+#!/bin/sh
+
+# Build a .tar.bz2 archive of Git for Windows with post-install steps already applied.
+
+test -z "$1" && {
+	echo "Usage: $0 <version> [optional components]"
+	exit 1
+}
+
+die () {
+	echo "$*" >&1
+	exit 1
+}
+
+ARCH="$(uname -m)"
+case "$ARCH" in
+i686)
+	BITNESS=32
+	;;
+x86_64)
+	BITNESS=64
+	;;
+*)
+	die "Unhandled architecture: $ARCH"
+	;;
+esac
+VERSION=$1
+shift
+TARGET="$HOME"/Git-"$VERSION"-"$BITNESS"-bit.tar.bz2
+SCRIPT_PATH="$(cd "$(dirname "$0")" && pwd)"
+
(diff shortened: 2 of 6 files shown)


dscho:
Thanks!

A couple of comments, though:

- Where do those hard-coded `/etc/...` files from? Should they not rather be copied from the current installation instead?
- How about adding `/archive/root/` wholesale to the `.gitignore` file?
- the commit message is awfully parsimonious. Please spend some time on it, it will make a very nice user story. Please see 23672af723da18e5bc3c679e52106de3c2dec55a for a good example, if I may say so, and https://github.com/git-for-windows/git/wiki/Good-commits for more inspiration.

With regard to that latter point: the commit message should *at least* mention that the post-install is required in `portable/release.sh` for the simple reason that the `.7z` format has no idea about symlinks, and that `portable/release.sh` uses the `.7z` format because it comes with a convenient graphical self-extractor.

Thank you again!

bastianeicher:
* The Git SDK installer copies those `/etc` files from `C:\Windows\System32\drivers\etc`. Therefore, if `archive/release.sh` were to take them from the local installation the resulting archive would contain any custom `hosts` entries the packager may have added to their system. To ensure the generated archive remains "clean" from local influences I took the corresponding files from a vanilla Ubuntu installation.
 * I decided not to add the whole `/archive/root/` directory to the `.gitignore` file because it does contain at least some non-ignore/checked-in files, namely the aforementioned `/etc` stuff.
 * You're right, I was bit lazy there. ;) I've crafted a more comprehensive one. Please let me know if it needs some more tweaking.

dscho:
> To ensure the generated archive remains "clean" from local influences I took the corresponding files from a vanilla Ubuntu installation.

Hmm. We need to make sure that this is allowed by the license. Aren't there BSD equivalents?

> I decided not to add the whole /archive/root/ directory to the .gitignore file because it does contain at least some non-ignore/checked-in files, namely the aforementioned /etc stuff.

That's what the negative patterns (i.e. lines prefixed with an exclamation point) are for ;-)

> I've crafted a more comprehensive [commit message]

Excellent!

bastianeicher:
> Hmm. We need to make sure that this is allowed by the license. Aren't there BSD equivalents?

I've replaced the `/etc` files with copies from https://github.com/freebsd/freebsd/tree/master/etc.

> That's what the negative patterns (i.e. lines prefixed with an exclamation point) are for ;-)

Didn't know about that feature yet, thanks. :) Applied.

dscho:
Perfect, thanks!

PR #91, opened 2015-12-30
git-extra: Source ~/.profile as well in generated ~/.bash_profile

elieux:
Bash will prefer `~/.bash_profile` over `~/.profile`, so if we're generating
`~/.bash_profile` for the user, we now insert a clause to source the original
`~/.profile` to keep it from getting skipped.

As mentioned on Gitter.


dscho:
@elieux just pinging in case this has been lost in the mail traffic of last year.

elieux:
Yeah, it did, but due to the way GitHub send email notifications, the ping got lost as well. :)

I followed your advice, but now the indentation looks horrible. I also added one more change. I'm open to further requests (though it may be with a slight delay again).

dscho:
I decided to do this as a follow-up patch, as well as adjusting the `md5sums`. As of `git-extra-1.1.110.a4c35b9-1`, it can be enjoyed in the SDK! :smile: 

Thanks, @elieux !

elieux:
Oh shoot, I forgot about the checksums.

Glad to help.

PR #93, opened 2016-01-16
Add initial infrastructure for gitsetup.exe

robmen:
This is the first step on the path to move the `robmen/gitsetup` repo
into the official git-for-windows. There is much work remaining but
this provides the base infrastructure necessary to get underway.


dscho:
Great, thank you so much!

dscho:
FYI I just added a script to update the WiX toolset easily: https://github.com/git-for-windows/build-extra/blob/master/msi/update-wix-toolset.sh

msi/GitBundle.wxs
```diff
@@ -0,0 +1,20 @@
+<?xml version='1.0'?>
+<Wix xmlns='http://wixtoolset.org/schemas/v4/wxs'
+     xmlns:bal='http://wixtoolset.org/schemas/v4/wxs/bal'>
+<Bundle Name='!(bind.packageName.git.msi)' Manufacturer='!(bind.packageManufacturer.git.msi)'
+        Version='2.6.4.32' UpgradeCode='4adb169f-377b-4a52-b817-4ecee51fe64d'
+        IconSourceFile='git.ico'>
+
+  <BootstrapperApplicationRef Id='WixStandardBootstrapperApplication.HyperlinkLicense'>
+    <bal:WixStandardBootstrapperApplication LicenseUrl='http://www.opensource.org/licenses/gpl-2.0' LogoFile='git.png' />
```
dscho: Could we reuse https://raw.githubusercontent.com/git-for-windows/build-extra/master/git-for-windows.svg here? Or are only PNGs allowed, no SVGs?

msi/make.cmd
```diff
@@ -0,0 +1,33 @@
+rmdir /s/q build
+
+rem TODO: Where should the git-credential-winstore, gcm and posh-git binaries
+rem       be found/stored?
```
dscho: They should be bundled as Pacman packages first... But let's leave them out for the initial .msi support? We can worry about them once we have something working.

msi/make.cmd
```diff
@@ -0,0 +1,33 @@
...
+
+rem TODO: Generate GitComponents.wxs from ..\make-file-list.sh
```
dscho: See https://github.com/git-for-windows/build-extra/compare/master...dscho:msi :smile:

msi/make.cmd
```diff
@@ -0,0 +1,33 @@
...
+rem          all others use the pattern Directory='INSTALLFOLDER:\path\'.
+rem
+rem       2. The `git.exe` file in `BinFolder` is given the explicit
+rem          Id='GetExe' all other files are anonymous (do not have an Id).
+
+rem Build the .msi.
+rem
+rem TODO: Correctly set the bind paths (change the -b gitfileshere switch)
+rem       to the real source location for the binaries to include in the
+rem       installation package.
```
dscho: What is the bind path?

msi/make.cmd
```diff
@@ -0,0 +1,33 @@
...
+rem
+rem TODO: Set the output paths appropriately.
```
dscho: You mean for temporary files?

msi/make.cmd
```diff
@@ -0,0 +1,33 @@
...
+rem
+wix\candle.exe GitProduct.wxs GitComponents.wxs GitCredStoreComponents.wxs GitPoshComponents.wxs -o build\obj\  -ext WixUtilExtension
+wix\light.exe build\obj\GitProduct.wixobj build\obj\GitComponents.wixobj build\obj\GitCredStoreComponents.wixobj build\obj\GitPoshComponents.wixobj -o build\msi\git.msi -ext WixUtilExtension -b gitfileshere -b ..\installer -sval
+
+rem Build the bundle.
```
dscho: What is the difference between the msi and the bundle?

robmen:
For https://github.com/git-for-windows/build-extra/blob/master/msi/update-wix-toolset.sh you may find it is easier to get the latest build number from our appsyn feed at http://wixtoolset.org/releases/feeds/wix4-0.feed

msi/GitBundle.wxs
robmen: `.bmp`, `.gif`, `.jpg`, `.png` are all supported by GDI+. No SVG support.

msi/make.cmd
robmen: Okay, but I hope that is sooner than later. :)

msi/make.cmd
robmen: Very cool. Not clear where to leave comments on that work. Are you sending a pull request for it soon?

msi/make.cmd
robmen: Think of "bind paths" as search paths for binaries to be included into the package. Relative paths in `@Source` and `@SourceFile` attributes the .wxs files are searched from the set of bind paths (provided with the `-b` switch).

So rather than `cd /` in the msi/release.sh, you can add the `/` path as a `-b` switch and that folder will be used to find files. The `getfileshare` value in the `make.cmd` is the one that should be changed.

You can add as many `-b` switches as useful. They are searched in order.

msi/make.cmd
robmen: No, I simply meant for the output of `candle.exe` to flow into the input of `light.exe` and `light.exe` to put the output where you wanted it.

msi/make.cmd
robmen: The msi is the MSI package. The bundle is an executable, often called a bootstrapper, that can wrap the .msi or the .msi can be external (but when shipping over the internet single executables--under 200MB--are easier to distribute). The bundle does work to cache the .msi package, register a stable source and several other things that improve the installation success rate of installing .msi packages over the internet.

dscho: Ah, so if my only desire is to have a `.msi` to ship (we already have an `.exe` and will continue to have that), then I can ignore the bundle, correct?

dscho:
> For https://github.com/git-for-windows/build-extra/blob/master/msi/update-wix-toolset.sh you may find it is easier to get the latest build number from our appsyn feed at http://wixtoolset.org/releases/feeds/wix4-0.feed

Heh, it is hardly easier (it is the same level of difficulty) but it is probably more robust. So I adjusted it: 081c7c1

msi/GitBundle.wxs
dscho: Bummer.

msi/make.cmd
dscho: Heh... You can help! But first let's have something that works. Please note that current Git for Windows does *not* ship those components, though, so that would add to my maintenance burden. Besides, I have no idea whether e.g. posh-git's author would be okay with us shipping their product (there was no attempt from posh-git's side to contact us so far).

msi/make.cmd
dscho: You can comment on the commit diffs themselves. I plan to just merge this without a Pull Request as soon as I have an `.msi` that works (read: installs Git and runs `post-install.bat` correctly).

msi/make.cmd
dscho: Oh, that's a useful feature!

msi/make.cmd
robmen: Yeah, found that it works better in Chrome that IE (as does much of GitHub). :smirk:

msi/make.cmd
robmen: To be clear, these components (plus the simplified installation) were the reason for gitsetup.exe existed.

msi/make.cmd
robmen: I wouldn't do that. The bundle works around a lot of the issues that can occur when distributing naked MSI packages for the internet. For example, the Windows Installer will register the location the MSI is launched from as the source. If launched straight from the browser that location could be some location deep in the browser cache. If the MSI requires a repair (for example, the `post-install.bat` missing because it deletes itself) then the prompt will be for the MSI in the browser cache.

Additionally, it may make sense to create separate MSI for the posh-git and git-credentials and the bundle creates a single unified experience for installing all of those.

The bundle works around many small issues distributing naked MSIs over the internet and adds many useful features.

Instead, I recommend shipping the bundle as was done in gitsetup.exe. In WiX v4 we'll finish the feature to create an admin image (i.e. extracts the MSIs embedded in the bundle) for those users that want to deploy the naked MSI(s) via GPO. For those users, the bundle is the single distribution container (useful if multiple MSIs are used for Git for Windows).

msi/GitBundle.wxs
robmen: Someone could always do the work to add SVG support to the WiX Toolset... :smile:

msi/make.cmd
dscho: I understand what you're saying, and I will try to help you there, but my main interest was to shut up the many users who want .msi installers, but not badly enough to lift even one finger. :smile:

msi/GitBundle.wxs
dscho: It's not *that* big of a problem for me. My biggest problem is that I currently end up with a `.msi` that is ~900kB and that installs an empty `C:\Program Files\Git`... And now I have to switch tasks ;-)

msi/make.cmd
dscho: Thanks for the explanation.

I agree that it makes sense to work around these problems, but please note that Git for Windows already has an installer that works quite well (and sports a couple of configuration pages that might not be possible with `.msi` files). The *only* reason I see to have support for `.msi` is to be able to push them via GPO...

msi/make.cmd
robmen: Fair enough but we use gitsetup.exe at my company. This effort will be a massive waste of time for me if we lose functionality. I'll have to go back and still maintain gitsetup.exe. My hope was in the end we'd get the best of all worlds in one place.

This is important to me.

msi/GitBundle.wxs
robmen: Can you commit your incremental step and maybe I can poke at what is going awry from there a bit?

msi/make.cmd
dscho: Let's try to get both of our goals met, then.

msi/GitBundle.wxs
dscho: I just force-pushed it. I test with `./release.sh 2.7.0-msi` in `/usr/src/build-extra/` in a Git for Windows SDK with the `msi` branch of https://github.com/dscho/build-extra checked out.

PR #96, opened 2016-01-30
MSI build working

robmen:
Builds on @dscho's changes to get a working git.msi.


dscho:
Thanks!

There are a couple more issues I would like to see addressed:
- `git-bash.exe` and `git-cmd.exe` seem to not be installed
- I would love to have the Start Menu group "Git" with the Git Bash, Git CMD and Git GUI entries
- post-install.bat seems to remain behind
- in my hands, the .msi built from 64-bit installs into `C:\Program Files (x86)`

I will have a look at all that, but any hint is appreciated.

robmen:
Finally caught up on my other work. Is this list still the remainder of work to be complete?

PR #101, opened 2016-03-21
installer: enable fscache by default

dscho:
This feature really stood the test of time by now, and the wish to
enable it by default has been mentioned quite a couple of times to the
Git for Windows maintainer. Let's make it so.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1027,7 +1027,7 @@ begin
     end;
 
     // Restore the settings chosen during a previous install.
-    Data:=ReplayChoice('Performance Tweaks FSCache','Disabled');
+    Data:=ReplayChoice('Performance Tweaks FSCache','Enabled');
 
     if Data='Enabled' then begin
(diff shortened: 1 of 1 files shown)


linquize:
It is still labeled as **experimental** in the installer.

dscho:
Where is the Pull Request?

paladox:
Here https://github.com/git-for-windows/build-extra/pull/106

paladox:
Also is this only enabled by default for new users since when upgrading the option is still unticked.

paladox:
Also done here https://github.com/git-for-windows/build-extra/pull/107

dscho:
> Also is this only enabled by default for new users since when upgrading the option is still unticked.

This is by design: the options that were chosen the last time Git for Windows was installed are retained, even if they were left at their then-defaults.

You can easily uninstall Git and the install it again to get the new defaults.

paladox:
Oh ok thanks for explaning.

PR #102, opened 2016-03-21
installer: switch the default PATH option to 'CMD'

dscho:
Many Windows users are very familiar with `cmd.exe` and not really with
the Bash at all. Requiring them to call `Git CMD` explicitly just to be
able to use Git for Windows from their favorite command-line is not
really helpful, in particular when the command-line is launched from
within other software.

So let's just switch the default to adding the `git.exe` (and gitk.exe
and git-gui.exe`) to the PATH.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -696,7 +696,7 @@ begin
     end;
 
     // Restore the setting chosen during a previous install.
-    Data:=ReplayChoice('Path Option','BashOnly');
+    Data:=ReplayChoice('Path Option','Cmd');
 
     if Data='BashOnly' then begin
         RdbPath[GP_BashOnly].Checked:=True;


fourpastmidnight:
@dscho: If I understand the intent of this correctly, it's so that when a user opens a Windows Command Prompt (after having installed Git for Windows), they can type `git --version`, for example, and git would execute and display the version. Presumably, they could then just start using git from the Windows Command Prompt instead of having to use Git Bash which they may be unfamiliar with.

However, looking at the changeset in this PR, the default option selected would also result in Windows command line tools like `find` and `sort` being "overridden" by the ones provided with Git for Windows.

Was this the behavior you desired by setting the `Cmd` option to be the default selection? I would have though the default should be set to `Both` so that the `%PATH%` environment variable is at least set such that Git can be executed from either Bash or the Command Prompt, but not override the standard Windows command line utilities.

fourpastmidnight:
@dscho I think I answered my own question, but correct me if I'm wrong: Because the `PATH` environment variable is modified such that the Git paths are placed at the _end_ of the variable, the standard Windows utilities will _not_ be overridden.

fourpastmidnight:
Actually, I completey mis-read the options. BTW, I'm really glad about this change, too.

dscho:
Yeah, the `Cmd` option only adds the `cmd/` directory to the `PATH`, i.e. `find` is not overridden. `CmdTools` would do that.

PR #103, opened 2016-03-21
installer: offer to launch "Git Bash" right away

dscho:


dscho:
Not *really* sure whether this is something people would want, so I'll leave this open for discussion for a while or so.

nalla:
Could be missleading if the user choose "cmd" integration to offer a git bash start.

dscho:
True. OTOH the default PATH option is going to change to Cmd, in all likelihood. So we cannot really rely on that setting stating the preference of the current user.

I guess the safest would be to add the option, but leave it unchecked by default.

PR #105, opened 2016-03-22
installer (64-bit): remove 2GB pack size limit

Carina26:
TODO: fix redirection and test!!!

In 64-bit setups it does not make sense to limit the pack size to 2GB.

This fixes https://github.com/git-for-windows/git/issues/288.


8v060htwyc:
version 2.8.1.windows.1

maysusan:
thank

PR #106, opened 2016-03-30
Remove experimental tag from fscache

paladox:
Follows-up: https://github.com/git-for-windows/build-extra/pull/101


dscho:
Thanks.

For future reference, I prefer self-contained commit messages (with a Sign-off) that add a link to the original discussion. But in this case it is a trivial change, hopefully it will not come back to haunt us.

PR #107, opened 2016-03-30
In #101 fscache was enabled by default but was not in with with RdbExtraOptions[GP_FSCache]

paladox:
This patches enables it there.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1016,7 +1016,7 @@ begin
         Height:=ScaleY(17);
         Font.Style:=[fsBold];
         TabOrder:=0;
-        Checked:=False;
+        Checked:=True;
     end;
     LblFSCache:=TLabel.Create(ExtraOptionsPage);
     with LblFSCache do begin


installer/install.iss
```diff
@@ -1016,7 +1016,7 @@ begin
         Height:=ScaleY(17);
         Font.Style:=[fsBold];
         TabOrder:=0;
-        Checked:=False;
+        Checked:=True;
```
dscho: This is overridden in lines 1035ff anyway.

dscho:
This commit message:

```
 In #101 fscache was enabled by default but was not in with RdbExtraOp…

…tions[GP_FSCache] do begin

This patches enables it there.
```

might be technically correct, but violates [our standards](https://github.com/git-for-windows/git/wiki/Good-commits#commit-messages). Remember, the purpose of a commit message is

- to give the reader a good idea what the change is about,
- *not* to repeat what the diff says,
- to provide additional information that might be necessary to convince the reader that the change is good,
- and quite a bit more.

In light of the lines 1035ff (which override the `Check:=False` anyway), I am not certain that it is worth reworking?

dscho:
@paladox actually, after re-reading your patch: with this change, if the user deselected the option previously, there is *no way* to keep it deselected.

I am therefore convinced that this patch is incorrect...

PR #108, opened 2016-04-04
Add arcanist to installer

paladox:
Allows us to download arcanist and install once downloaded.

Adding support for the inno-download-plugin that will allow us to download the arcanist installer I created at https://github.com/paladox/Arcanist-installer-for-windows

Include https://code.google.com/archive/p/inno-download-plugin/ inno setup
plugin. It allows us to download files if the the user selects the check
box that downloads it.

This update then installs the setup.exe once it is downloaded.


dscho:
The commit message still does not conform to our standards.

And the patch just adds an empty line?

paladox:
@dscho yes sorry, I was going to do the patch on my local machine and upload it.

paladox:
@dscho ok done. The commit msg isen't as good as your guidelines butt I have added some details on what the patch does.

dscho:
Wow. This is a *huge* monster commit. And it not only adds the entire Inno Download Plugin (in a manual way that does not imitate [`update-inno-setup.sh`](https://github.com/git-for-windows/build-extra/blob/master/installer/update-inno-setup.sh) and therefore makes upgrading the plugin hard) but also conflates this (already hard-to-review) change with the change to `installer/install.iss` that uses the plugin.

This change would need to be split into three commits to be easier to review:

1. a script like `update-inno-setup.sh` that downloads the Inno Download Plugin and extracts it into the proper place
2. the files that were downloaded by that script (and I would verify locally that they have not been tampered with, after downloading)
3. the change to `install.iss` that makes use of this plugin.

Having said that, I am very reluctant to accept this Pull Request because I do not agree that it would be a good idea for the *Git for Windows* installer to offer installing something that is completely unrelated.

Instead, I would like to encourage you to consider two alternative ways to provide Arcanist to other users:

- either you make your installer a standalone one that piggy-backs onto Git for Windows and has to be reinstalled every time Git for Windows is reinstalled, or
- go fully MSYS2 and provide the necessary packages as MSYS2 (or MINGW) packages.

Personally, I would go with the second option, if I were you. MSYS2 already provides an infrastructure for proper dependency management and all you would have to do is to come up with the build scripts to build the `php`, `libphabricator` and `arcanist` packages.

Then all your users would need to do is install MSYS2 (or the Git for Windows SDK) and call `pacman -Syu arcanist`. And they would automatically keep up-to-date by calling `pacman -Syu` every once in a while.

paladox:
@dscho I already created a standalone installer at https://github.com/paladox/Arcanist-installer-for-windows

I'm wondering if we could add that installer into git for windows meaning execute that setup if the user selects the option to do so meaning it is optional, the box will be unticked by default to make sure users understand that it is optional. Plus a description that says this is arcanist a script for contributing to phabricator projects.

The installer also includes php as an optional dependacy which is also unticked by default.

dscho:
@paladox If I were to accept this change, I would be stuck with maintaining it, and the next thing I know is that Git for Windows' installer offers to install Notepad++, Atom, Ruby on Rails, Apache, etc.

And I would be stuck with maintaining all of that.

This is not a good direction. Git for Windows is about Git for Windows. Add-ons are add-ons, and should not be integrated into Git for Windows' installer.

paladox:
@dscho ok, could we create a page on git for windows that allow users to add links to download the add ons. That would be much better and not turning it into maintenance for you. Since it will allow users to find add ons such as arc quickly.

As long as they are approved by you meaning create a repo that we can host these addons to make sure they are not modified to run code that could compromise security.

dscho:
> could we create a page on git for windows that allow users to add links to download the add ons. That would be much better and not turning it into maintenance for you. Since it will allow users to find add ons such as arc quickly.

Sure!

> As long as they are approved by you

As third-party software, there is no need for me to approve anything.

paladox:
@dscho ok, should we do it so it is approved by you or someone you trust since someone could modify it after a user installs It

PR #109, opened 2016-04-12
Integrate ssh-pageant

landstander668:
@dscho This is my first stab at integrating ssh-pageant support into Git for Windows [[Issue #683]](https://github.com/git-for-windows/git/issues/683). This has been working flawlessly for me under Windows 7 64-bit, provided that the _ssh-pageant_ package was already installed (performed via `pacman -Sy ssh-pageant`) when the installer/portable was built. The reason for request being marked as DO NOT MERGE YET is because I'm unsure if additional action is required, in order to ensure the package is present (by default) when you make a build... my guess is that it needs to somehow be tagged as "install by default" for the SDK, or similar.

The benefit of this integration is that it actually helps us to become _less_ reliant on PuTTY, within the specific context of Git for Windows. For those of us using PuTTY outside of this tool, it removes the need to run and manage multiple agents to manage key-based authentication. In this case, OpenSSH is able to seamlessly handle all of the heavy lifting.

### DIFF
--- a/sdk-installer/setup-git-sdk.bat
+++ b/sdk-installer/setup-git-sdk.bat
@@ -138,7 +138,7 @@
 	mingw-w64-@@ARCH@@-curl mingw-w64-@@ARCH@@-expat ^
 	mingw-w64-@@ARCH@@-openssl mingw-w64-@@ARCH@@-tcl ^
 	mingw-w64-@@ARCH@@-pcre mingw-w64-@@ARCH@@-connect ^
-	git-flow
+	git-flow ssh-pageant
 
 @IF ERRORLEVEL 1 GOTO INSTALL_REST
 
--- a/make-file-list.sh
+++ b/make-file-list.sh
@@ -31,9 +31,10 @@ pacman_list () {
 required=
 for req in mingw-w64-$ARCH-connect git-flow unzip docx2txt \
 	mingw-w64-$ARCH-antiword mingw-w64-$ARCH-xpdf \
-	mingw-w64-$ARCH-git-credential-manager
+	mingw-w64-$ARCH-git-credential-manager ssh-pageant
 do
 	test -d /var/lib/pacman/local/$req-[0-9]* ||
+	test -d /var/lib/pacman/local/$req-git-[0-9]* ||
 	required="$required $req"
 done
 test -z "$required" ||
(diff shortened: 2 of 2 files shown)


landstander668:
@dscho I've added a second commit to ensure that _ssh-pageant_ gets installed with the SDK, and unless you know of additional changes which are required I believe the pull request should now be complete. No changes to the InnoSetup-based installer were required, as it automatically picked up the beastie as long as the package was installed at the time the installer was built. When installing, it was treated exactly the same as ssh-agent.exe... namely that it informs the user that the process must be killed before proceeding.

Here's what I did to test the changes.

1. Built the installer, and verified that _ssh-pageant_ was both included and functioned properly once installed (Listed, added, and removed keys from the agent. Pulled from and pushed to a GitHub repository using OpenSSH, without being prompted for the key passphrase.).
2. Built a portable installer, and verified same.
3. Built a SDK installer, and verified that the _ssh-pageant_ package was downloaded and applied when it was installed. Verified the same functionality as above.

In addition, I've verified that the utility can be used in conjunction with Git CMD (and directly from the Windows Command Prompt), instead of being limited to Git Bash (as I had originally believed). I'll update the [Wiki page](https://github.com/git-for-windows/git/wiki/OpenSSH-Integration-with-Pageant) with the relevant info once I know you're happy with the proposed changes.

sdk-installer/setup-git-sdk.bat
```diff
@@ -138,7 +138,7 @@
 	mingw-w64-@@ARCH@@-curl mingw-w64-@@ARCH@@-expat ^
 	mingw-w64-@@ARCH@@-openssl mingw-w64-@@ARCH@@-tcl ^
 	mingw-w64-@@ARCH@@-pcre mingw-w64-@@ARCH@@-connect ^
-	git-flow
+	git-flow ssh-pageant
```
dscho: You probably also want to add it to [this section](https://github.com/landstander668/build-extra/blob/ssh-pageant/make-file-list.sh#L30-L41).

dscho:
Very nice already.

A couple of suggestions:
- [ ] [Sign off](https://github.com/git/git/blob/v2.8.1/Documentation/SubmittingPatches#L234-L286) your patches,
- [ ] Include a comparison of the installers' sizes before and after the change,
- [ ] Append a link to `ssh-pageant`'s home page to the commit message of the commit that includes it in the installer.

landstander668:
@dscho If you don't mind humoring my ignorance, what's the correct/preferred way to modify commits in this context? I'm assuming that _rebase_ is out, since they've already been pushed to a public repository. Revert the commits and reapply with the necessary changes?

While I've used Git a fair amount internally, so far it's always been small (one or two person) projects where this scenario has never come up. This is my first time working with a public project (using version control, anyway), so I'm not very familiar with the process yet.

Thanx!

dscho:
> I'm assuming that rebase is out, since they've already been pushed to a public repository.

Oh, I think you point to something really important: it is probably not documented anywhere, but I run this project by keeping only `master` force-push free. *All other* branches are fair game for rebasing (upstream maintains a `maint` branch that we do not replicate, but that would also need to be handled as "don't force-push, ever").

So: please use interactive rebase to rewrite the branch, then force-push to update this Pull Request (and then please ping me, I am not notified when there are newly-pushed commits).

Thank you!

landstander668:
@dscho I've made the requested changes, and built / tested the installers again to ensure that everything was still working as expected. No issues were encountered, so I think this pull request should be good to go. (-;

dscho:
Nice! Thank you!

Only one question left: did you want to keep the different author email in the first commit?

landstander668:
Sorry, no... that was inadvertent. I've just re-pushed the branch with that detail corrected (there were no other changes).

dscho:
Perfect, thanks!

landstander668:
Thanx @dscho! (-;

PR #110, opened 2016-04-14
Add support for Git LFS

shiftkey:
There's likely some improvements to do here, but opening this early for feedback:

 - [x] package and test i686 package locally
 - [x] package and test x86_64 package locally
 - [x] confirm compressed file size is accurate 
 - [x] wireup equivalent `git lfs install` and `git lfs uninstall` steps to update config
 - [x] understand the root cause of the issue with this file not being executable after install

```
brendanforster@SHIFTKEYDE8B MINGW32 /usr/src/build-extra/git-lfs (add-support-for-git-lfs)
$ /mingw32/bin/git-lfs.exe
bash: /mingw32/bin/git-lfs.exe: cannot execute binary file: Exec format error

brendanforster@SHIFTKEYDE8B MINGW32 /usr/src/build-extra/git-lfs (add-support-for-git-lfs)
$ file /mingw32/bin/git-lfs.exe
/mingw32/bin/git-lfs.exe: PE32 executable (console) Intel 80386 (stripped to external PDB), for MS Windows
```

 - [x] simplify packaging steps
 - [x] update version to v1.2
 - [x] correct installation to use new rules

cc @dscho as we discussed last week
cc @technoweenie just cause

### DIFF
--- a/git-lfs/PKGBUILD
+++ b/git-lfs/PKGBUILD
@@ -1,0 +1,34 @@
+# Maintainer: Brendan Forster <brendan@github.com>
+
+_realname="git-lfs"
+pkgbase="mingw-w64-${_realname}"
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=1.2.0
+pkgrel=1
+pkgdesc="An open source Git extension for versioning large files"
+install=git-lfs.install
+arch=('any')
+url="https://github.com/github/git-lfs"
+license=('MIT')
+groups=('VCS')
+
+case "$CARCH" in
+i686)
+  folder="git-lfs-windows-386-$pkgver"
+  md5sum=0cea336e4c9d43314ebc692dfbbd1fc3
+  ;;
+x86_64)
+  folder="git-lfs-windows-amd64-$pkgver"
+  md5sum=95296495a70a31dd952de15bf7102ead
+  ;;
+esac
+
+source=("https://github.com/github/git-lfs/releases/download/v$pkgver/$folder.zip")
+
+md5sums=("$md5sum")
+options=('!strip')
+
+package() {
+  install -d -m755 $pkgdir/$MINGW_PREFIX/bin
+  install -m755 $srcdir/$folder/git-lfs.exe $pkgdir/$MINGW_PREFIX/bin/git-lfs.exe
+}
--- a/git-lfs/git-lfs.install
+++ b/git-lfs/git-lfs.install
@@ -1,0 +1,21 @@
+post_install() {
+  for arch in mingw32 mingw64; do
+    if [ -f "${arch}/bin/git-lfs.exe" ]; then
+      git config -f "${arch}/etc/gitconfig" filter.lfs.clean "git-lfs clean -- %f"
+      git config -f "${arch}/etc/gitconfig" filter.lfs.smudge "git-lfs smudge -- %f"
+      git config -f "${arch}/etc/gitconfig" filter.lfs.required true
+    fi
+  done
+}
+
+post_upgrade() {
+  post_install
+}
+
+pre_remove() {
+  for arch in mingw32 mingw64; do
+    if [ -f "${arch}/bin/git-lfs.exe" ]; then
+       git config -f "${arch}/etc/gitconfig" --remove-section filter.lfs
+    fi
+  done
+}
--- a/git-lfs/.gitignore
+++ b/git-lfs/.gitignore
@@ -1,0 +1,4 @@
+pkg
+src
+*.zip
+*.xz


strich:
Hey @shiftkey. I'm one of the primary contributors for the LFS Windows installer.

I'm not sure how much has been discussed with @technoweenie, but just a quick history: When LFS first came out the Win installer made attempts at finding the Git-for-windows install and placing the `git-lfs.exe` exec there. It was a bit hit and miss and as part of upgrading the installer framework itself we decided to install LFS into its own folder under Program Files (By default - The user can choose), and the global `PATH` env var gets updated with the install location. There are other feature now such as being dual-arch and automatically initializing Git LFS for the first time.

Integrating Git LFS with Git-for-windows will help close the gap for non-technical users, so this is really exciting! Might I suggest that instead of directly packaging Git LFS execs in your installer, you instead package the Git LFS installer itself (Or better yet download the latest version at install-time) and run it silently. That way you gain the benefit of the features already in the LFS installer as well as allowing users to independently update/remove LFS outside of git-for-windows.

shiftkey:
@strich thanks for the feedback! I did mention this in passing to @technoweenie, and while I want to avoid duplicating work that the team is doing there's some packaging constraints that make things interesting on the Git for Windows side.

I went with this approach as this makes packaging simpler for the Git for Windows team:

 - no need to run a secondary installer 
 - this supports both PortableGit (bin-deployed) and machine-wide (requires elevation) installs
 - packaging infrastructure supports 32/64-bit packages

I've not thought about scenarios where you might have multiple versions of `git-lfs.exe` on your `PATH` but I'm not sure what I can do here.

> There are other feature now such as being dual-arch and automatically initializing Git LFS for the first time.

Ah yes, I knew there was something that I'd missed. I'd need to ensure that Git LFS is initialized after installation  - by invoking `git lfs init` or equivalent - and I think I need to do that as part of the publishing step - rather than here.

>  Might I suggest that instead of directly packaging Git LFS execs in your installer, you instead package the Git LFS installer itself (Or better yet download the latest version at install-time) and run it silently. 

I'm not sure how this fits with the existing Git for Windows philosophy. I'll let @dscho speak to this but I think users will want this to be as standalone as possible.

> That way you gain the benefit of the features already in the LFS installer as well as allowing users to independently update/remove LFS outside of git-for-windows.

A couple of questions I have:

 - Is managing Git LFS explicitly a common request? If we're publishing frequent updates for Git for Windows (there's been three in the last four weeks) and we're updating this Git LFS package whenever it has new releases, what other sorts of scenarios are we not covering here?

 - Are there uninstall steps that we'd need to consider if we're mimicking the Git LFS installer here? I know the ~/.gitconfig changes might need to be removed, but are there others?

```
[filter "lfs"]
	smudge = git-lfs smudge %f
	required = true
	clean = git-lfs clean %f
```

strich:
> Is managing Git LFS explicitly a common request? 
Unlikely. Its more that since its currently early in its life-cycle there may be quite a few patch releases people might be required to often update for. However if your team release quite often yourselves then maybe its not an issue.

> Are there uninstall steps that we'd need to consider if we're mimicking the Git LFS installer here? 
As you guessed - Only the .gitconfig entries need to be removed, which you can do with `git lfs uninstall` prior to removing it.

> Ah yes, I knew there was something that I'd missed. I'd need to ensure that Git LFS is initialized after installation...
Yep just run `git lfs install`.

> I've not thought about scenarios where you might have multiple versions of git-lfs.exe on your PATH but I'm not sure what I can do here.
If you have `git-lfs.exe` next to `git.exe`, as you propose you'll do, then I believe Windows will first check there before accessing `PATH`. So in the event someone installs git-for-windows as well as Git LFS using our installer, the git-for-windows version of LFS should win.

You guys have a few options here as I see it:
1. You can completely ignore the potential presence of an existing Git LFS installation and given the above your package should win and _should_ work as the user expects.
2. You can package LFS as you propose and also automatically detect and uninstall any existing LFS installation.
3. You can package the LFS installer and run it silently, passing in your directory, and any other options you might wish (I could add any custom functionality you need).

I think I would prefer option 3 if only because it _should_ come with the least amount of edge case issues the user might find themselves in. For example I'd worry that the user would install git-for-windows and then later Git LFS using our installer, either because they weren't aware it was in git-for-windows or because they wanted to update LFS. Such an edge case would not break LFS, but they might find it didn't update as expected.

I do understand you guys have your own ecosystem and requirements however. Happy to continue answering questions, etc.

dscho:
> > Might I suggest that instead of directly packaging Git LFS execs in your installer, you instead package the Git LFS installer itself (Or better yet download the latest version at install-time) and run it silently.
> 
> I'm not sure how this fits with the existing Git for Windows philosophy. I'll let @dscho speak to this but I think users will want this to be as standalone as possible.

Yeah, unfortunately this is a no-go. The idea of Git for Windows (since 2.x) is to have the installer reflect a partial MSYS2 installation, with a MSYS2 setup being the primary development environment. Therefore, everything that gets packages should be bundled in MSYS2 packages.

> Yep just run git lfs install.

What does this do? Just editing the `etc/gitconfig` file?

I am asking because I do *not* want to run `git lfs install` due to concerns about the portable Git.

> If you have git-lfs.exe next to git.exe, as you propose you'll do, then I believe Windows will first check there before accessing PATH.

That is only correct for `.dll` files, not `.exe` files. Whatever comes first in your `PATH` will win.

> I think I would prefer option 3

I do not think this would fly well with portable Git (which is what @shiftkey is mostly concerned, as he packages a modified version of portable Git as the command-line tools of GitHub for Windows). Even having to run a command to install something we would want to avoid: it is too easy to put absolute paths into the gitconfig that are then invalid when moving the portable Git to a new computer.

So I would prefer option 1.

As to releasing new Git for Windows versions when new Git LFS versions become available: we already release new Git for Windows versions when either Git or the Git Credential Manager releases new versions. I could easily add Git LFS to that list.

strich:
Ah okay. Fair enough then. So the only edge case I can see here is if a user installed LFS before installing Git-for-windows, and thereby making LFS appear first in `PATH`. I could add a check in the LFS installer to tell the user to install git first if it cannot find it anywhere in that case.

>What does this do? Just editing the `etc/gitconfig` file?
I am asking because I do not want to run git lfs install due to concerns about the portable Git.

I believe so yes. @technoweenie would better be able to tell you exactly what changes you need to replicate manually in place of `git lfs install`. Though I do not think it adds any kind of paths, absolute or otherwise, so it would be better to use `git lfs install` so you don't have to worry about any future changes to the way that works.

git-lfs/PKGBUILD
```diff
@@ -0,0 +1,41 @@
+# Maintainer: Brendan Forster <brendan@github.com>
+
+pkgname=('git-lfs')
+pkgver=1.1.2
+pkgrel=1
+pkgdesc="An open source Git extension for versioning large files"
+arch=('i686' 'x86_64')
```
dscho: You want to build this with `mingw-makepkg`, right? Then you need to set `arch=('any')`. The reason is that 32-bit and 64-bit MinGW packages (as opposed to MSYS2 packages) can be installed side by side, in both a 32-bit and a 64-bit MSYS2 setup.

So as long as the package does not rely on the MSYS2 runtime (and Git LFS does not), we should use `mingw-makepkg` to generate the packages, and set `arch=('any')`.

To discern one from the other architecture, the `pkgname` needs to reflect the architecture: `pkgname=('${MINGW_PACKAGE_PREFIX}-git-lfs')` (compare with [the Git Credential Manager's `PKGBUILD`](https://github.com/git-for-windows/build-extra/blob/master/mingw-w64-git-credential-manager/PKGBUILD#L5)).

git-lfs/PKGBUILD
```diff
@@ -0,0 +1,41 @@
...
+url="https://github.com/github/git-lfs"
+license=('MIT')
+groups=('VCS')
+
+source=(
+      "https://github.com/github/git-lfs/releases/download/v$pkgver/git-lfs-windows-386-$pkgver.zip"
+      "https://github.com/github/git-lfs/releases/download/v$pkgver/git-lfs-windows-amd64-$pkgver.zip")
```
dscho: I think there should be only one source. The easiest might be something like

```sh
case "$CARCH" in
i686)
   zip="git-lfs-windows-386-$pkgver.zip"
  ;;
x86_64)
  zip="git-lfs-windows-amd64-$pkgver.zip"
  ;;
esac

source=("https://github.com/github/git-lfs/releases/download/v$pkgver/$zip")
```

git-lfs/PKGBUILD
```diff
@@ -0,0 +1,41 @@
...
+
+md5sums=('8f4cc7008042fa88f9d10b257d93d632'
+         '7034f7ea80fb2d12033829a8d9d1c6c4')
```
dscho: This would need to go into the `case` statement, too:

```sh
...
   md5sum=8f4cc7008042fa88f9d10b257d93d632
...

md5sums=("$md5sum")
```

git-lfs/PKGBUILD
```diff
@@ -0,0 +1,41 @@
...
+
+# both files have a git-lfs.exe so we need to extract these manually
+# TODO: perhaps this can be simplified into one expression
+noextract=("git-lfs-windows-386-$pkgver.zip" "git-lfs-windows-amd64-$pkgver.zip")
```
dscho: I think you *want* the single `.zip` extracted, and then copied to `$pkgdir/$MINGW_PREFIX/`.

On another note, I could imagine that MSYS2's `strip.exe` wreaks havoc on the Go executables. So maybe

```sh
options=('!strip')
```

is needed?

dscho:
> wireup git lfs init steps as part of post-install work

This could be done in a post-install step. See https://github.com/git-for-windows/MINGW-packages/blob/master/mingw-w64-qt5-git/qt5.install and https://github.com/git-for-windows/MINGW-packages/blob/master/mingw-w64-qt5-git/PKGBUILD#L11 for an example.

dscho:
> understand the root cause of the issue with this file not being executable after install

I believe that the `options=('!strip')` might help with that.

git-lfs/PKGBUILD
shiftkey: By having one source defined using `case` the `noextract` becomes unnecessary. But thanks for the tip...

git-lfs/PKGBUILD
shiftkey: :thumbsup:

shiftkey:
I'll look at the post-install stuff next and putting that in `/etc/gitconfig`, but that feedback helped out a lot :sparkles:

shiftkey:
One minor issue:

> 
$ git-lfs install --help
git lfs install [options]
>
Perform the following actions to ensure that Git LFS is setup properly:
>
* Set up the clean and smudge filters under the name "lfs" in the **global Git
  config.**
* Install a pre-push hook to run git lfs pre-push for the current repository,
  if run from inside one.
>
> ...

Actually, this might be an issue. @strich would it be possible for us to run this against the system config? Running this at package time means it won't be available to users when they install...

strich:
That was recently discussed here: https://github.com/github/git-lfs/issues/799#issuecomment-200340776.

Basically - There is currently no option to write the smudge config to the system Git config. Though most would agree it should by default.

However as @technoweenie points out in the link above, the config above doesn't affect any repos that aren't managed with LFS so it shouldn't be an issue beyond the principle of it.

technoweenie:
I'm cool with changing it, but I imagine we'd have to worry about system privs. I'm going to add `--system` to the roadmap: https://github.com/github/git-lfs/pull/1145

dscho:
> Total Installed Size: 11.73 MiB

Is that accurate? Is the `git-lfs.exe` really *that* large?

technoweenie:
Yup. Go binaries are statically linked with the entire Go runtime.

shiftkey:
Added the `--system` version of `git lfs install/uninstall` to the package.

Ready for review :metal:

git-lfs/git-lfs.install
```diff
@@ -0,0 +1,21 @@
+post_install() {
+  for arch in mingw32 mingw64; do
+    if [ -f "${arch}/bin/git-lfs.exe" ]; then
+      git config --system filter.lfs.clean "git-lfs clean %f"
+      git config --system filter.lfs.smudge "git-lfs smudge %f"
+      git config --system filter.lfs.required true
```
technoweenie: Ah, this is the danger of duplicating here, instead of relying on LFS: those values aren't completely correct. If you're bundling LFS v1.2 or lower, then keep this. If not, I'll make sure to add `git lfs install --system` to the next LFS release.

You can fix it for now with:

```
git config --system filter.lfs.clean "git-lfs clean -- %f"
git config --system filter.lfs.smudge "git-lfs smudge -- %f"
```

Here's why we need the `--`: https://github.com/github/git-lfs/pull/1083.

shiftkey: @technoweenie thanks - and I missed the fact v1.2 has just shipped so I'll make sure we're including that!

shiftkey:
:apple:

dscho:
> Yup. Go binaries are statically linked with the entire Go runtime.

Urgh. So we would increase the size of the installer by a whopping 35% *just* to include Git LFS? Hrm. Isn't it possible to `strip` this a bit? Or is at least the compressed size *not* that horrible?

technoweenie:
Funny you should ask! [Shrink your Go binaries with this one weird trick](https://blog.filippo.io/shrink-your-go-binaries-with-this-one-weird-trick/) was posted yesterday. 

TLDR:

* [`strip` is untested with Go binaries, and not recommended](https://github.com/docker/docker/blob/2a95488f7843a773de2b541a47d9b971a635bfff/project/PACKAGERS.md#stripping-binaries)
* Work in Go 1.7 is being done to reduce the size, so hopefully it'll get better on its own.
* We can strip DWARF tables with `go build -ldflags="-s -w"`, which leaves stack traces intact. We can investigate this for v1.3.

I don't want to do anything that increases startup time, since the default configuration using git clean/smudge filters can call the `git-lfs` binary many times on large repositories. But, we can probably live without debugging.

Good news is that it does compress really well. Check out the [zip file releases](https://github.com/github/git-lfs/releases/tag/v1.2.0). The Windows Installer is only 4.78MB, while including both the 32bit and 64bit versions in an Inno Setup installer.

dscho:
Okay. I still think it is quite concerning, though, if the Git for Windows installer would increase by more than 10% *just* for LFS support (which not many people need, quite honestly).

git-lfs/git-lfs.install
```diff
@@ -0,0 +1,21 @@
+post_install() {
+  for arch in mingw32 mingw64; do
+    if [ -f "${arch}/bin/git-lfs.exe" ]; then
+      git config --system filter.lfs.clean "git-lfs clean -- %f"
```
dscho: Let's replace the `--system` with `-f "${arch}/etc/gitconfig"`. That way, we can be certain that the *correct* system gitconfig gets updated.

shiftkey:
I'll leave the packaging size discussion for everyone else to debate vigorously. I'm curious why we're not seeing those same compression benefits here @dscho - the compressed packages (which are then installed into the Git SDK environment) are only slightly larger than the upstream Git LFS dual installer:

dscho:
Hey @shiftkey, I'd like to merge this Pull Request (even if I would prefer to build `git lfs` ourselves so we can squeeze the binaries, but MSYS2's `go` package [does not work correctly yet](https://github.com/Alexpux/MINGW-packages/issues/421)). To this end, may I ask you to rebase it interactively, add your Sign-off's to all of the commits, and squash as appropriate?

shiftkey:
@dscho squashed, rebased and signed off

dscho:
Perfect, thanks!

PR #111, opened 2016-04-20
Support generating NuGet packages of Git for Windows

dscho:
We steal heavily from ourselves here...

Please note that the Install.ps1 script is only called if the NuGet
package is installed inside Visual Studio (and not if installed via
nuget.exe; this is a bug in nuget.exe as of time of writing).

Therefore, the user should still call tools\post-install.bat manually
or start tools\git-bash.exe once before doing anything else, if
nuget.exe is used to install this package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,9 @@
 /msi/obj/
 /msi/package-versions.txt
 /msi/wix40-binaries.zip
+/nuget/GitForWindows.nuspec
+/nuget/nuget.exe
+/nuget/package-versions.txt
 /portable/root/bin/
 /portable/root/cmd/
 /portable/root/etc/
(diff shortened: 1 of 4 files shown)


whoisj:
Generally looks good.

I'm surprised there's not a way to do all / most of this via args to nuget.exe.

dscho:
> I'm surprised there's not a way to do all / most of this via args to nuget.exe.

Heh. Would be too easy.

Another question: do you think *GitForWindows* is a good name, or should it be *Git*? (*PortableGit* is unfortunately [taken by somebody else](https://www.nuget.org/packages/PortableGit/))

whoisj:
I'd go with something like `Git.Windows.Portable`. This allows for `Git.Linux`, `Git.Osx`, etc. Additionally, it allows for other future flavors of "Git for Windows". Perhaps `Git.Windows.UberStyle` or what have you.

dscho:
AFAICT on NuGet the names are not really hierarchical. And GitForWindows is not really the same as Git for Linux (because we have to ship *so* much infrastructure *just* to make Git work alright). So I kept the original name.

PR #112, opened 2016-04-21
Fix msi

robmen:
This brings together the work by @dscho and @fourpastmidnight to correctly install all of the files (including those in the root folder which were not being handled properly) and creates the desired shortcuts.

In other words, this PR should address the issues listed in #96 .

### DIFF
--- a/msi/GitProduct.wxs
+++ b/msi/GitProduct.wxs
@@ -30,6 +30,8 @@
       <Component Directory='INSTALLFOLDER'>
         <RegistryValue Root='HKLM' Key='Software\Git' Name='InstallFolder' Value='[INSTALLFOLDER]' Type='string' />
         <Environment Name='PATH' Value='[BinFolder]' Action='set' Part='last' System='yes' />
+
+        <RemoveFolder Directory='GitProgramMenuFolder' On='uninstall' />
       </Component>
 
       <ComponentGroupRef Id='GitComponents'/>
@@ -45,6 +47,12 @@
         <Directory Id='INSTALLFOLDER' Name='Git'>
           <Directory Id='BinFolder' Name='bin' />
         </Directory>
+      </Directory>
+
+      <Directory Id='PersonalFolder' />
+
+      <Directory Id='ProgramMenuFolder'>
+        <Directory Id='GitProgramMenuFolder' Name='Git' />
       </Directory>
     </Directory>
   </Fragment>
--- a/msi/release.sh
+++ b/msi/release.sh
@@ -68,7 +68,10 @@ sed -e 's/\(.*\)\\\(.*\)/            <Component Directory="INSTALLFOLDER:\\\1\\"
             <\/Component>/' \
 	    -e 's/^\([^\\]*\)$/            <Component Directory="INSTALLFOLDER">\
                 <File Source="&" \/>\
-            <\/Component>/'
+            <\/Component>/' \
+	    -e 's/\(<File Source="git-bash.exe"[^>]*\) \/>/\1 \/><Shortcut Name="Git Bash" Icon="git.ico" Directory="GitProgramMenuFolder" WorkingDirectory="PersonalFolder" Advertise="yes" \/>/' \
+	    -e 's/\(<File Source="git-cmd.exe"[^>]*\) \/>/\1 \/><Shortcut Name="Git CMD" Icon="git.ico" Directory="GitProgramMenuFolder" WorkingDirectory="PersonalFolder" Advertise="yes" \/>/' \
+	    -e 's/\(<File Source="cmd\\git-gui.exe"[^>]*\) \/>/\1 \/><Shortcut Name="Git GUI" Icon="git.ico" Directory="GitProgramMenuFolder" WorkingDirectory="PersonalFolder" Advertise="yes" \/>/'
 
 cat <<EOF
         </ComponentGroup>
@@ -87,6 +90,6 @@ wix/light.exe \
 	obj/GitProduct.wixobj \
 	obj/GitComponents.wixobj \
 	-o $TARGET -ext WixUtilExtension \
-	-b / -b ../installer -sval &&
+	-b / -b ../installer &&
 echo "Success! You will find the new .msi at \"$TARGET\"." ||
 die "Could not generate $TARGET"
--- a/msi/release.sh
+++ b/msi/release.sh
@@ -63,9 +63,13 @@ BUILD_EXTRA_WINPATH="$(cd "$SCRIPT_PATH"/.. && pwd -W | tr / \\\\)"
 EOF
 echo "$LIST" |
 tr / \\\\ |
-sed 's/\(.*\)\\\(.*\)/            <Component Directory="INSTALLFOLDER:\\\1\\">\
+sed -e 's/\(.*\)\\\(.*\)/            <Component Directory="INSTALLFOLDER:\\\1\\">\
+                <File Source="&" \/>\
+            <\/Component>/' \
+	    -e 's/^\([^\\]*\)$/            <Component Directory="INSTALLFOLDER">\
                 <File Source="&" \/>\
             <\/Component>/'
+
 cat <<EOF
         </ComponentGroup>
     </Fragment>


robmen:
@dscho if this looks good, would you mind start opening issues ([here?](https://github.com/git-for-windows/git/issues)) so I can track what you want completed so the MSI can be released?

A single place to find and discuss the remaining issues would be tremendously helpful for me.

msi/GitProduct.wxs
```diff
@@ -46,6 +48,12 @@
           <Directory Id='BinFolder' Name='bin' />
         </Directory>
       </Directory>
+
+      <Directory Id='PersonalFolder' />
```
dscho: What is this needed for?

msi/release.sh
```diff
@@ -87,6 +90,6 @@ wix/light.exe \
 	obj/GitProduct.wixobj \
 	obj/GitComponents.wixobj \
 	-o $TARGET -ext WixUtilExtension \
-	-b / -b ../installer -sval &&
+	-b / -b ../installer &&
```
dscho: Would you mind explaining this change in the commit message? I always stumble upon this, and I always wish the commit message was not so parsimonious with extra information that really, really helps reviewing.

msi/release.sh
```diff
@@ -63,9 +63,13 @@ BUILD_EXTRA_WINPATH="$(cd "$SCRIPT_PATH"/.. && pwd -W | tr / \\\\)"
...
 echo "$LIST" |
 tr / \\\\ |
-sed 's/\(.*\)\\\(.*\)/            <Component Directory="INSTALLFOLDER:\\\1\\">\
+sed -e 's/\(.*\)\\\(.*\)/            <Component Directory="INSTALLFOLDER:\\\1\\">\
+                <File Source="&" \/>\
+            <\/Component>/' \
+	    -e 's/^\([^\\]*\)$/            <Component Directory="INSTALLFOLDER">\
                 <File Source="&" \/>\
             <\/Component>/'
+
```
dscho: Why redo this commit, with completely different authorship, when I already did it: dscho/build-extra@306e63f2d6c8b83a79c2614602abad60a7df3231 (I see that you even partially copied the commit message, but you left out the last paragraph that was intended to help reviewers by not forcing them to guess why the `Guid` was skipped).

dscho:
Thanks.

I do take a little exception with being stripped of the authorship, in conjunction with deleting parts of the carefully crafted commit message. Please simply cherry-pick, and *add* info to the commit message, instead of removing authors and parts of the commit message. It was hard enough to figure all of this out, might just as well get some credits. While at it, @fourpastmidnight should get credit, too, because he pointed me into the right direction (although his commit message kind of left me in the dark about the details, I really had to guess a lot).

Once the authorship is reinstated and the commit messages are improved, I'll gladly merge this PR.

msi/GitProduct.wxs
robmen: Working folder for the shortcut. See the ./release.sh

msi/release.sh
robmen: I did.

> Also enables static analysis to catch errors handling per-user folders.

robmen:
Sorry, that was definitely *not* my intention. My goal was to try to get the clean commits that you clearly prefer and it never occurred to me to try to cherry-pick the changes and then rebase more changes back on top of those make the small edits necessary to your changes. I actually did start with your commits and added a third commit that essentially said "Make the good work that @dscho did based on the good work that @fourpastmidnight did to work with minimal code." But I thought you'd tell me to squash that all down into what I did in the end.

Note I removed portions of the commit message because they no longer applied. The only thing that has a Guid at this point is the post-install.bat. That has an empty Guid because post-install.bat's self-destruct behavior messes with the Windows Installer desire to keep the installed files safely installed. It is still foreign to me to put code comments in the commit messages instead of in the code but it seems like that's the convention (my guess is it comes from the stance that comments in code get stale but haven't seen that written down).

You may have to look closely, but you'll see my changes are not exactly what you did. I'm happy to give you all the credit for them since the hard part (IMHO) was the sed line noise in the ./release.sh. But 90% of the lines did have to be touched to get the correct fix.

So, if you look at ./release.sh and see that most of it is not verbatim what your changes were, do you still want me to cherry-pick and add a commit or try to rebase the changes into the cherry-pick (not sure if that's even going to work the way I hope it does)?

Also, how would you like me to give credit for @fourpastmidnight? He's clearly done a lot of work to try to get some of these fixes in and looks to be doing more (thus my desire to get individual issues tracking any remaining problems so we can divide and conquer).

Finally, with all due respect, I spend significantly more time in thi sproject trying to get git commit messages written than I do actually making the code clear, concise and correct. The commit history is clearly very important to you so I'm trying. But, just so you know, it is quite taxing for something I'm just doing on the side.

msi/GitProduct.wxs
dscho: You speak in riddles, sir. Why do we need a working folder? And why is it called `PersonalFolder` when it sounds more like it is a temporary folder?

msi/release.sh
dscho: > I did.

Educate me.

Is

> Create a 'Git' folder in the program menu and install the cmd, bash and GUI shortcuts there.

telling me why `-sval` was removed, and what it does, or is it

> Also enables static analysis to catch errors handling per-user folders.

If you thrive on making other people feel like they are stupid, dumb idiots, I congratulate you for making me another target :smile:

msi/GitProduct.wxs
robmen: Rarely is the default working folder what you want (IIRC): C:\Windows\system32 . Git is not system software, so I didn't think it should start the user running in the system folder.

It's called `PersonalFolder` because that's what the Windows Installer calls it: [PersonalFolder][pf] (I generally find this [MSDN doc][doc] more useful). Personally, I would have named it something else, probably: `DocumentsFolder` but they didn't ask me. :smile_cat: 

[pf]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa370806%28v=vs.85%29.aspx
[doc]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa370905%28v=vs.85%29.aspx#system_folder_properties

dscho:
> Note I removed portions of the commit message because they no longer applied. The only thing that has a Guid at this point is the post-install.bat.

That is exactly what my comment went for. Since my change imitated the post-install.bat entry, I felt it necessary to explain why it was less of a verbatim copy-edit.

> It is still foreign to me to put code comments in the commit messages instead of in the code

This is not a code comment, but a comment on the implementation decision. It needs to go into the commit message for that reason.

This strategy also prevents comments from going stale when the code they refer to is removed or changed beyond recognition.

> You may have to look closely, but you'll see my changes are not exactly what you did.

I did not see that. Maybe this is another good opportunity to describe information in the commit message that is not readily inferred from looking at the diff?

> Also, how would you like me to give credit for @fourpastmidnight?

Personally, what I do in such cases is determine the clear name from the contributor's commits and then say something "Based on outstanding research by [XYZ]" in the commit message, or something similar.

> Finally, with all due respect, I spend significantly more time in thi sproject trying to get git commit messages written than I do actually making the code clear, concise and correct. The commit history is clearly very important to you so I'm trying. But, just so you know, it is quite taxing for something I'm just doing on the side.

Believe me, I do appreciate *any* effort you put into commit messages.

The reason is that I have to juggle quite a bit of different projects. In the course of Git for Windows maintenance, I have to use InnoSetup, soon WiX, C, Make, Perl, Tcl/Tk, Javascript, node.js, npm, NuGet, VirtualBox, Linux' `/proc` pseudo file system, various profiling tools and stuff and things. Of course I do not remember how to use all of them! So I rely heavily on well-written commit messages that I can easily find via `git blame` (or `git log` with various "pickaxe" options) to get up to speed quickly, whenever I have to touch certain tools or APIs or things.

So the more you can help me by describing peculiar aspects of WiX in the commit message, the better.

msi/release.sh
robmen: Sorry, it not my goal to make you feel stupid but it also isn't my goal to create a primer how to use the WiX Toolset via commit messages.

The [linker documentation][doc] shows:

> -sval  Suppress MSI/MSM validation.

Validation is Windows Installer's term for "static analysis" (I tend to use the term static analysis now since more people know what that means). Static analysis is great at catching errors before the code is shipped.

It was disabled.

I enabled it because per-user folders are a great place to introduce mistakes that static validation will catch.

So that's what I was trying to say with:

> Also enables static analysis to catch errors handling per-user folders.

[doc]: http://wixtoolset.org/documentation/manual/v3/overview/light.html

robmen:
> That is exactly what my comment went for. Since my change imitated the post-install.bat entry, I felt it necessary to explain why it was less of a verbatim copy-edit.

Ahh, I didn't see it that way at all. post-install.bat is the outlier. Nothing should be done in relation to it. It is done differently from everything else so all the comments (in my mind) are about why post-install.bat's code is more complex than everything else (aka: self-destruct).

> This is not a code comment, but a comment on the implementation decision. It needs to go into the commit message for that reason.

The changes I made were to keep the code minimal and correct as possible. A simple File element with no adornments (and no need to discuss KeyPath) with Components (with no adornments) is all that is necessary. I found discussion in the commit message about the lack of those attributes confusing since, again, the only Component and File that has adornments is post-install.bat (see above).

> I did not see that. Maybe this is another good opportunity to describe information in the commit message that is not readily inferred from looking at the diff?

I truly believe this is a matter of you seeing several commits made by @fourpastmidnight that were not necessary (all good, it looks like he was learning) in the final commit.

Let's break the changes down:

7fac95c277fbe562dcb16fc063d5c013c78566bf is all about fixing the sed expression to correctly add files in the top-level directory. That's all there is. No need to talk about KeyPath's because they were there before (see line 67). It also makes the added files look identical to 3 of the 4 other files hardcoded in ./release.sh.  Again, post-install.bat is the outlier, not these files.

87297968e14e7e7f4c21385ee08f53e7236cf944 adds the shortcuts to ./release.sh and code necessary to support them in in the Git program menu folder. Then I dis-disabled (aka: re-enabled) validation because per-user folders are easy to mess up (because Windows Installer).

That's it. I thought the commit messages did a proper job not calling out any other confusing side quests that are not relevant.

>  "Based on outstanding research by [XYZ]"

Okay. I can add that to the commit.

> So the more you can help me by describing peculiar aspects of WiX in the commit message, the better.

Right. You'll notice I keep the code very simple to reduce confusion. I often see people adding extraneous attributes to their WiX code then spending time discussing the extraneous attributes which only adds to the confusion. I did none of that and all this discussion is happening is because in history in other forks did.

I'll add the credit message to the commits and repush.

msi/release.sh
robmen: Added the "Based on outstanding..." recognition per discussion elsewhere in PR.

dscho:
Thanks!

@robmen I would still have preferred at least a rudimentary repetition of the MSDN/WiX docs in the commit messages (so that I have a prayer to fix bugs even with flakey or non-existent internet connection).

OTOH I did not want to be the blocker, so I just merged it.

PR #116, opened 2016-04-30
Updating inno setup to 5.5.9 stable

paladox:
This updates to the stable release inno setup 5.5.9


dscho:
Same suggestion as last time you wanted to update InnoSetup: give me a reason why it is a good thing to do, and investigate whether the changes in InnoSetup are even in the least likely to break Git for Windows' release management.

paladox:
Well I have been using innosetup 5.5.9 for another project and has been working well for me.

The reasons we should update is we are currently using inno setup 5.5.9-dev which may include bugs that were fixed in the stable 5.5.9 release.

It includes support for signing your files before storing them.

It doesn't look like anything will break but I carn't be 100% sure.

dscho:
> we are currently using inno setup 5.5.9-dev which may include bugs that were fixed in the stable 5.5.9 release.

Please provide evidence to this claim.

> It includes support for signing your files before storing them.

We do not use this feature, do we?

> It doesn't look like anything will break but I carn't be 100% sure.

Yes, we could be sure. By diligently going through the announcement of the InnoSetup release and verifying that it would not cause a regression, and then at least building one test installer.

Have you done that? If yes, it needs to be mentioned in the commit message (and not as a whimsical "I looked through the release notes"; it needs to be clear from the text that you *actually* looked through the release notes, e.g. by giving the link, and by summarizing the changes and discussing the potentially problematic ones).

If not, you definitely need to do that in order to make this Pull Request complete.

You are touching a very sensitive area of Git for Windows, and so far I have no reason to trust that it will continue to work as intended. Instead, I have reason to believe that it might very well break because nothing has been done to ensure that it won't. Even worse: I have good reason to believe that I would be stuck with any problems that might be incurred by this Pull Request, with no help whatsoever, very close to a critical new Git for Windows version. I really wish that your work would save me time, but currently it does the opposite.

Therefore, for the record, once again: it is not enough to open a Pull Request for something as vitally important as the release engineering setup, just because a new version of InnoSetup is available. To avoid increasingly major frustration on my side, it is really, really important to provide evidence that instills trust that the Pull Request will not cause major hick-ups.

In short: I cannot accept the Pull Request in its current form.

PR #113, opened 2016-04-22
Feature/git for windows msi

fourpastmidnight:
Here's the long-awaited pull request for [git-for-windows/git issue #682](https://github.com/git-for-windows/git/issues/682).

PR #117, opened 2016-04-30
Update wix toolset to v4.0.3922.0

paladox:
http://wixtoolset.org/releases/feeds/wix4-0.feed


dscho:
The very same comments as for your InnoSetup-upgrading PR go here, too. It would be awesome, of course, if you saved me time, but at the moment, the opposite is the case: I am baffled how I am supposed to trust that this update is something we even want. Plus, I have to point out, again, how a commit message that consists of a mere link slapped together with a bogus "Change-Id" (what is *that* supposed to be???) does not at all conform with [our oft-linked guide lines](https://github.com/git-for-windows/git/wiki/Good-commits#commit-messages) (I *will* keep pointing to you until you read them and follow them).

In short: you can do much better than this.

fourpastmidnight:
Please see [PR#113](https://github.com/git-for-windows/build-extra/pull/113) where there's already a commit to update the toolset. The reason this PR has not been merged is that the MSI is still under heavy development. Unless and until an MSI is being built as an official installation vector for Git for Windows, this is not really necessary at this time. The older version of the toolset is sufficient to build the MSI in its current state.

paladox:
Per ^^

PR #119, opened 2016-06-03
WIP: innosetup: support optional packages [DO NOT MERGE]

fourpastmidnight:
This is a first crack at adding support for including additional packages
in the InnoSetup distribution vehicle. It's rough and contains lots of
commented out code and debug output.

All other release vehicles (msi, nuget, portable) allow you to include
additional, optional packages, but InnoSetup release process does not.

This developer is reluctant to make this feature addition lightly because
this is the primary release vehicle for Git for Windows.

So, this developer is putting this commit out there for any early feedback
and to perhaps at least spark a discussion, if nothing else.

This change was instigated via
https://github.com/git-for-windows/git/issues/771.

### DIFF
--- a/installer/release.sh
+++ b/installer/release.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+usage () {
+	die "
+Usage: $0 [-f | --force] [--skip-files] --debug-wizard-page=<page> [<optional-package>...]
+   or: $0 [-f | --force] [--skip-files] <version>[-test] [<optional-package>...]
+"
+}
+
 die () {
 	echo "$*" >&2
 	exit 1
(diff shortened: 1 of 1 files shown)


installer/release.sh
```diff
@@ -9,48 +16,87 @@ die () {
...
 		;;
 	*)
+		if test -z "$version"
+		then
+			if test -n "$test_installer"
+			then
+				version=0-test
+			else
+				version=$1
+				test_installer="$(echo "$version" | grep -o -e "-.*" | sed 's/.*/t/')"
```
fourpastmidnight: The regex I used for `grep` is naive and far from ideal. It's just a start, though.

fourpastmidnight:
After further thoughts regarding this idea, I've decided it's not a good idea after all. When something doesn't work right, undoubtedly people will come here for support for an unsupported custom installation.

nalla:
@fourpastmidnight  isnt  https://github.com/git-for-windows/build-extra/pull/16 just the thing you want?

fourpastmidnight:
@nalla #16 was what led me to create this PR. At one time, I thought it would be nice to have an installer support this (e.g. InnoSetup or MSI). The idea was that instead of having to use the SDK (which is meant for GfW development), someone could pass packages as command line options to the installer to install other packages. But after having worked on the MSI over the last year, and the conversations on PR #131, this was a horribly mis-guided idea. And as you pointed out, if someone really wants this, they already have two good options: Use the SDK or use the portable GfW installation vehicle. This PR would cause much harm to maintaining this project, and so I closed this PR.

PR #123, opened 2016-07-14
Notepad and commit message, proper saving in utf-8.

dimztimz:
Fixes the issue git-for-windows/git#818 .


git-extra/notepad
```diff
@@ -15,7 +15,14 @@ die "Usage: $0 <file>"
...
 then
-	unix2dos.exe "$1"
+	commitencoding="$(git config --get i18n.commitencoding)"
+	if [ $? -ne 0 ]  ||
+		[ "$commitencoding" = "utf-8" ] ||
+		[ "$commitencoding" = "utf8" ]; then
+		unix2dos.exe -m "$1"
+	else
+		unix2dos.exe "$1"
+	fi
```
dscho: Could you rewrite this as a

```sh
case "$(git config ...)" in
utf-8|utf8) ...
...
```

then run `updpkgsums` in `git-extra/` to update the checksums in `PKGBUILD`, and augment the commit message with a description of the problem that is solved?

dscho:
What a nice contribution!

I added some suggestions in a comment; please feel free to force-push after addressing them and leave a note here so I do not miss the updates.

git-extra/notepad
rimrul: Isn't this wrapper also used for other tasks that might be UTF encoded? How well does `unix2dos` handle files that already are in UTF-8BOM with the `-m` flag? Could Notepad also mess up with other UTF encodings? How well does this solution handle edge cases like unexpected crashes? Would something like this, but modified to accomodate more situations than just commit messages be a better solution?

````diff
then
	unix2dos.exe "$1"
fi &&
-notepad.exe "$1" &&
+case "$(git config --get i18n.commitencoding)" in
+utf*)
+		notepad.exe /W "$1"
+*)
+		notepad.exe "$1"
+esac &&
 dos2unix.exe "$1" &&
 case "$1" in
 */COMMIT_EDITMSG|*\\COMMIT_EDITMSG)
````

Sources for the `/W` argument:
[1](http://answers.microsoft.com/en-us/windows/forum/windows_7-files/notepadexe-command-line-options/810760c1-a45a-4013-9544-1c1208e1b389)
[2](http://alinconstantin.blogspot.de/2012/08/notepadexe-command-line-arguments.html)
[3](http://superuser.com/questions/131826/notepad-switches)

**Comment by dimztimz** [comment]
@dscho Ok I updated with force push (i reseted the old commit and it went away)

@rimrul notepad /W is useless as that forces UTF-16. UTF-16 is not even allowed in commit messages. dos2unix wont add a second BOM if one is already present. That handed that case fine.

**Comment by rimrul** [comment]
>@rimrul notepad /W is useless as that forces UTF-16.

Oh, you're right. Sorry, poorly documented flag. One questions remain, though: Do we need to handle other use cases for this wrapper like interactive rebases for example? Does the `commitencoding` apply to more than commit messages regarding the editor or should we introduce a guard clause similar to this one?

````bash
case "$1" in
*/COMMIT_EDITMSG|*\\COMMIT_EDITMSG)
...
*)
...
esac
````

File: git-extra/notepad [hunk 2]
```diff
@@ -15,7 +15,12 @@ die "Usage: $0 <file>"
 
 if test -f "$1"
 then
-	unix2dos.exe "$1"
+	commitencoding="$(git config i18n.commitencoding)" ||
+		commitencoding=""
```
rimrul: Is this fallback needed? Can `git config i18n.commitencoding` return a value indicating failure? If it can, doesn't an output that matches `""|"utf-8"|"utf8")` still probably mean UTF-8 and an output that doesn't match probably mean something that's not UTF-8?

dimztimz: Git config returns error code if the variable is not found. Unset variable (the default) means utf-8.  The OR fallback is for that case

rimrul: But `git config` doesn't write anything to `stdout` in that case, so `"$commitencoding"` should still be an empty string.

dimztimz: I think it's fine this way. The return code is documented. Returning empty string is not as far as I saw.

dimztimz:
@rimrul the bug fix can not introduce new bugs. It may only leave the old behavior unresolved in very rare edge cases if the other subsystems that open editor do not respect the variable.

And that is highly unlikely. My guess is that in all cases the same subroutines are used.

rimrul:
It can introduce new bugs and/or pitfalls for future bugs if it uses the Commit encoding for things that expect some other encoding (wether that's configurable or not).

From `git help config`:

>**i18n.commitEncoding**

>**Character encoding the commit messages are stored in**; Git itself does not care per se, but this information is necessary e.g. when importing commits from emails or in the gitk graphical history browser (and possibly at other places in the future or in other porcelains). See e.g. git-mailinfo(1). Defaults to utf-8.

Using the Commit encoding for anything that is not a commit message is ~~undocumented behaviour~~behaviour that conflicts with the documentation.

git-extra/notepad
rimrul: There is no explicit documentation, that `git config` doesn't output anything to `stdout` or that all errors are written to `stderr`, but [Documentation/technical/api-error-handling.txt](https://github.com/git/git/blob/master/Documentation/technical/api-error-handling.txt) strongly suggests that errors should be written to `stderr`. [`stdio.h`](https://github.com/lattera/glibc/blob/master/libio/stdio.h#L167-L174), [stdio(3)] and [stderr(3)] suggest the same behaviour. I don't have POSIX/SUS/LSB/C89/C90/C95/C99/C11 standards at hand right now, but I'm sure they suggest similar behaviour. And what non-error information should git config output if it doesn't find a key or section?

dimztimz:
Dude you are just splitting hairs now. I never expected this much bureaucracy for a five line bug fix. I have analized enough and I did not find any possibility of new bugs.

If there are new bugs please find it, document it, show example and open an issue.

rimrul:
I'm sorry. I really didn't want to be nitpicking, but I feel like this could really cause issues in the future. I don't have a current bug to show you, but I think it's really not hard to see, that the general scenario I described above is likely to cause issues:

>Using the Commit encoding for anything that is not a commit message

Your fallback that I commented on regarding wether it's needed should be working fine either way.

git-extra/PKGBUILD
```diff
@@ -34,13 +34,13 @@ source=('inputrc'
         '99-post-install-cleanup.post'
         'astextplain'
         'git-sdk.sh')
-md5sums=('3fab57079f0322efe256e95fe29d516f'
-         'e7ad03fffc29e619e402dbd5ec9ef74c'
-         'bfb591886b2a28af3334521e71198b74'
```
dscho: Are you sure about these? Seems to me that you may have checked out the files with CR/LF line endings, when the MD5 checksums apply to the LF-only versions instead.

git-extra/notepad
dscho: I agree that the version without the fallback (and with `2>/dev/null`) reads much easier.

git-extra/notepad
```diff
@@ -15,7 +15,12 @@ die "Usage: $0 <file>"
...
+	case "$commitencoding" in
+		""|"utf-8"|"utf8") unix2dos.exe -m "$1";;
```
dscho: Please lose the quotes around everything in this line except `"$1"`.

dscho:
@dimztimz please realize that @rimrul only tried to help you come up with the best solution.

If you keep in mind that Git for Windows has been downloaded over a million times in the past four weeks, you will have to agree that it is a good idea to be cautious.

I can also help in this discussion by contributing my knowledge and now we have combined the best of three Open Source contributors. I find that really exciting (and not annoying at all).

In this particular case, I would like to add that @rimrul's concern about `COMMIT_EDITMSG` is a very, very valid concern. We introduced a different solution to this problem earlier, though, so all is good: we [skip all special handling for files outside of any `.git/` directory](https://github.com/git-for-windows/build-extra/blob/be72db6bdc4122298307f0561f63b9ebffef5ded/git-extra/notepad#L8-L11).

dscho:
@dimztimz thank you for your contribution! I felt that you got a bit distressed by the discussion (which is not uncommon in the Git project), so I just made the final changes myself and amended the commit.

dimztimz:
Thanks. I see now what was the hustle all about.

Git can launch the core editor for file in the project. I was initially thinking that was used only for commit message, tag message and similar.

Thanks again to everyone.

PR #124, opened 2016-07-18
msi: Fix Git for Windows MSI installer properties implementation [DON'T MERGE]

fourpastmidnight:
Updated the implementation of the Remember Property pattern to correctly support upgrade, repair and reinstall scenarios. The previous implementation of the pattern was based on a simplistic example meant to demonstrate a problem and a potential solution for it; but this installer is quite a bit more complex and the "simple solution" was not appropriate--it just plain didn't work in this instance. This implementation allows the installer to properly support upgrade, repair, and reinstallation scenarios.

This pull request also contains a refinement to the flow of UI dialogs when it comes to choosing the line-ending setting during installation for Git for Windows. The previous flow was a bit unintuitive and in-your-face.

Lastly, I found and corrected a mis-spelled property name that would've caused the user's `PATH` system environment variable to be modified even if the user selected to only use Git from the Bash shell

### DIFF
--- a/msi/GitProduct.wxs
+++ b/msi/GitProduct.wxs
@@ -118,23 +118,25 @@
       UILevel = 5 AND ENABLEFSCACHE="false"
     </SetProperty>
 
-    <SetProperty Id="CMDLINE_ENABLEGCM" Value="0" Before="AppSearch" Sequence="first">
-      ENABLEGCM~="false" OR ENABLEGCM=0
-    </SetProperty>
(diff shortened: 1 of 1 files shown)


fourpastmidnight:
Hold up on this a second--I just realized I need to check something out.

fourpastmidnight:
I changed some things and found something awful that would make this plain not work. Darn it.

dscho:
> I changed some things and found something

Thank you for your diligence!

PR #125, opened 2016-07-25
msi: Fix Git for Windows MSI installer properties implementation [HOLD OFF]

fourpastmidnight:
This pull request updates the implementation of all the MSI properties that have been implemented to date, which record the user's selected configuration options during installation.

Previously, the properties associated with tracking a user's installation configuration choices used the "Remember Property" pattern that @robmen wrote up a long time ago. It's a great pattern, but unfortunately, it doesn't work with properties which have a default value, especially when you need to know from where a property value originated in order to make some determinations about installer data and/or UI flow.

This pull request also contains a refinement to the flow of UI dialogs when it comes to choosing the line-ending setting during installation for Git for Windows. The previous flow was a bit unintuitive, and in some situations, could also lead the user to believing that a default configuration setting was being used when, in fact, a customized setting would be used. This was a bug--and this PR includes a commit to fix that bug.

Lastly, I found and corrected a mis-spelled property name that would've caused the user's `PATH` system environment variable to be modified even if the user selected to only use Git from the Bash shell.

### DIFF
--- a/msi/GitAdditionalOptionsDlg.wxs
+++ b/msi/GitAdditionalOptionsDlg.wxs
@@ -10,11 +10,7 @@
           <Condition Action="disable">NOT WIX_IS_NETFRAMEWORK_451_OR_LATER_INSTALLED</Condition>
         </Control>
         <Control Type="PushButton" Id="Back" Width="56" Height="17" X="180" Y="243" Text="!(loc.WixUIBack)" />
-        <Control Type="PushButton" Id="Next" Width="56" Height="17" X="236" Y="243" Default="yes" Text="!(loc.WixUINext)">
-          <Publish Property="ENABLEFSCACHE" Value="false"><![CDATA[ ENABLEFSCACHE <> "true" ]]></Publish>
-          <Publish Property="ENABLEFSCACHE" Value="true">ENABLEFSCACHE = "true"</Publish>
-          <Publish Property="ENABLEGCM" Value="0"><![CDATA[ ENABLEGCM <> 1 ]]></Publish>
-        </Control>
+        <Control Type="PushButton" Id="Next" Width="56" Height="17" X="236" Y="243" Default="yes" Text="!(loc.WixUINext)" />
         <Control Type="PushButton" Id="Cancel" Width="56" Height="17" X="304" Y="243" Text="!(loc.WixUICancel)" Cancel="yes">
           <Publish Event="SpawnDialog" Value="CancelDlg">1</Publish>
         </Control>
--- a/msi/GitLineEndingsDlg.wxs
+++ b/msi/GitLineEndingsDlg.wxs
@@ -9,8 +9,12 @@
             <RadioButton Text="!(loc.GitLineEndingsDlg_RbCustomize)" Height="30" Value="Yes" Width="310" X="0" Y="15" />
           </RadioButtonGroup>
         </Control>
-        <Control Type="PushButton" Id="Back" Width="56" Height="17" X="180" Y="243" Text="!(loc.WixUIBack)" />
-        <Control Type="PushButton" Id="Next" Width="56" Height="17" X="236" Y="243" Default="yes" Text="!(loc.WixUINext)" />
+        <Control Type="PushButton" Id="Back" Width="56" Height="17" X="180" Y="243" Text="!(loc.WixUIBack)">
+          <Publish Property="LINEENDINGS" Value="true">CustomizeLineEndings = "No"</Publish>
+        </Control>
+        <Control Type="PushButton" Id="Next" Width="56" Height="17" X="236" Y="243" Default="yes" Text="!(loc.WixUINext)">
+          <Publish Property="LINEENDINGS" Value="true">CustomizeLineEndings = "No"</Publish>
+        </Control>
         <Control Type="PushButton" Id="Cancel" Width="56" Height="17" X="304" Y="243" Cancel="yes" Text="!(loc.WixUICancel)">
           <Publish Event="SpawnDialog" Value="CancelDlg">1</Publish>
         </Control>
--- a/msi/GitProduct.wxs
+++ b/msi/GitProduct.wxs
@@ -69,6 +69,7 @@
 
     <Property Id="AUTOCRLF" Value="Unset" />
     <Property Id="GfwLineEndings" Value="true" />
+    <Property Id="SkipGitLineEndingsDlg" Value="No" />
     <Property Id="LINEENDINGS" Value="true" />
     <Property Id="LINEENDINGSSEARCH">
       <RegistrySearch Id="LineEndingsSearch" Root="HKLM" Key="SOFTWARE\GitForWindows" Name="core.autocrlf" Type="raw" />
@@ -83,6 +84,9 @@
     <SetProperty Id="GfwLineEndings" Value="input" Action="NormalizeCaseGfwLineEndingsInput" After="SetGfwLineEndingsFromCommandLine" Sequence="first">GfwLineEndings~="input"</SetProperty>
     <SetProperty Id="GfwLineEndings" Value="false" Action="NormalizeCaseGfwLineEndingsFalse" After="SetGfwLineEndingsFromCommandLine" Sequence="first">GfwLineEndings~="false"</SetProperty>
     <SetProperty Id="LINEENDINGS" Value="[GfwLineEndings]" Before="CostFinalize" Sequence="first" />
+    <SetProperty Id="SkipGitLineEndingsDlg" Value="Yes" Before="CostFinalize" Sequence="ui">
+      NOT GfwLineEndings="true" OR AUTOCRLF~="true" OR AUTOCRLF~="input" OR AUTOCRLF~="false"
+    </SetProperty>
 
     <Property Id="CMDLINE_TERMINAL" Value="MinTTY" />
     <Property Id="TERMINAL" Value="MinTTY" />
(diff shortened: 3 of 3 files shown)


fourpastmidnight:
OK, so I noticed an issue with properties that back a Checkbox UI control.

As I painstakingly noted in the commit message for the `ENABLEFSCACHE` and `ENABLEGCM` options, MSI checkboxes work exactly like HTML checkboxes--if the property backing the checkbox has a value--any value--the checkbox will be checked.

Upon clicking the `Next` button (should also be the `Back` button--but I think I forgot that, so yet another issue with this PR :disappointed: ), the property backing the checkbox is set to `0` if the checkbox is disabled (i.e. the property backing the checkbox is cleared). But, setting the property value to `0` means that if I click the `Back` button, the checkbox will now be checked (because the property has a value--any value.).

I will come up with a fix and amend this PR.

fourpastmidnight:
**UPDATE:** The issue with the checkboxes for `ENABLEFSCACHE` and `ENABLEGCM` is fixed with the 8th commit in this PR.

fourpastmidnight:
**HOLD UP ON THIS**: As I continue to read more and more about Windows Installer, I learn that there is more and more that I don't know. Anyway, I want to check some things out related to what I've done in this series of commits based off of new information I recently came across--before this gets merged in.

fourpastmidnight:
OK, after a long break, I'm reviewing this pull request again. So based on what I found about on 1 Augest 2016, I put this PR on hold. However, after having re-evaluated what I read, I've determined that the changes put forth in this PR are actually OK. What's _not_ OK is the UI. The UI currently does not account for reinstallation scenarios (i.e. running the MSI when Git for Windows is already installed). Here are the two main issues with the UI:

1. The Welcome Dialog will not advance when hitting the Next button during a reinstallation scenario.
    * What should happen is that the UI should advance to the installation Maintenance Dialog where either a repair installation or removal of Git for Windows can be chosen. Currently, there are no plans for supporting the "modification" of the installation of Git for Windows, as there's really only one feature at the moment. This could change--but I don't see a whole lot of value here.

2. Pretending for a moment that the Welcome Dialog is not broken as stated above, the UI would advance through the normal set of dialogs allowing the user to change installation settings. This should not be allowed. Again, during a "reinstallation", the UI should only show the Maintenance Dialog and then the "Ready to Repair" dialog before commencing a repair/reinstallation of Git for Windows using the current installation's chosen installation options. (Of course, I can't control whether or not someone manually fiddles with the registry--but that's their problem. :wink: )

So all in all, as far as "fixing" the MSI installer properties, I believe that this PR is in fine shape.

fourpastmidnight:
@dscho: Not sure if you're interested in reviewing PRs for the MSI installer anymore or not--let me know; this is ready for review for merge.

fourpastmidnight:
FYI: I already have a commit that fixes the Welcome dialog issue mentioned above, as well as the maintenance dialog sequence (for repair/removal of Git for Windows) which is ready for a PR. But since it's "on top" of the commits for this PR, I don't like to "stack up" PRs where one PR is dependent on another. On the other hand, learning from my first PR here, I also don't want to pack PRs with so many changes that it makes it hard to reason about and review the PR.

dscho:
> Not sure if you're interested in reviewing PRs for the MSI installer anymore

Yes! I'll do so tomorrow (unless I forget, in which case I humbly request a gentle prod).

fourpastmidnight:
No worries! I know at one point you said you wanted to be "hands off"--but I was unclear on what exactly you may have meant by that. (I took it to mean you don't want to be dragged into every little "design decision"--and I can totally appreciate that.) I feel much more comfortable having my code reviewed because there's always something--and I don't really want to mess something up for the 2M users of Git for Windows! Mistakes happen, but you know what they say: "An ounce of prevention is worth a pound of cure."  So I really appreciate having the extra set of eyes. :smiley:

dscho:
> at one point you said you wanted to be "hands off"--but I was unclear on what exactly you may have meant by that

I meant that I do not want to tell you what you have to do. You choose what you do, and how. I am quite happy how you run this project, so I think that approach worked just fine.

fourpastmidnight:
So, the reason for all of these changes to the properties for Git for Windows configuration is not so much to "guard against 'pilot errors'," so much as it is to ensure that MSI can operate over an installation of Git for Windows in a predictable manner in order maintain a consistent operating environment for users with respect to their installed software, especially with  respect to repair and upgrade scenarios.

Traditionally, software products are broken into Features and Components (in MSI installer-speak), according to however the software manufacturer elects to allow the user to customize their software installation. And MSI tracks the installation set of Features and Components. Features group Components, and Components are (typically) comprised of a single file or registry entry. During a repair, the components of installed features are checked to see if files or registry entries are missing or corrupt (this is an over-simplification), and if so, those components are repaired by "replaying" what was originally installed. Through all of this, however, the component never becomes "reinstalled" per s&eacute;--in fact, it's installation state never actually changes. This is why you can't decide to install additional (or remove existing) components of a feature during a repair--because MSI would never record these installation changes. It's merely repairing the state of the installed software to the way it is when the installer was originally run.

For the Git for Windows installer, we only really have _one_ feature (made of many components): Git for Windows. All the dialogs asking various questions about line endings, terminal emulator, etc., are as you said, configuration choices the user provides for _after_ Git for Windows is installed. These configuration choices are not features in and of themselves, and so MSI does not track the configuration of the installation--especially since that's "post-installation" (in terms of laying files down on disk).

So, throughout this MSI effort, I've been trying to allow MSI to manage some of this configuration as part of the actual installation process--mostly to support repair and upgrade scenarios. (I see this being important more for the repair scenario than the upgrade scenario, however.)

But perhaps I've been going about this all wrong.... Maybe I should've made these configuration actions come after the installation of the product? In this way, even during repair, I could still offer the user the chance to "reconfigure" their Git for Windows installation without affecting the MSI installation machinery nor the consistency of the installed state of the software in the user's operating environment. In addition, it would allow me the ability to actually support "modifying" the configuration of Git for Windows through the Modify workflow of the installer (but not actually modifying the installed state of the software, but rather, just the configuration, which is more likely what we're after anyway).

It might be worth it for me to branch off of this right now before going too much further and try this out--especially if it leads to the actual installation implementation being simpler. (For the record, I don't even know if this is possible--but I can't see why not--and I won't know until I try.)

fourpastmidnight:
OK, so after some more thought, I don't think this PR is all that bad. This PR ensures repair works as expected. What's noticeably missing is the ability to "modify" an installation (or should I say, an installation configuration) via the MSI. It hasn't been implemented (yet). Typically modification would mean modifying which features are installed--but we only have one (so far). So in our case, this would allow a user to change the configured product for purposes of "recording" their desired installation options--primarily for upgrade and repair scenarios (mostly upgrades in the modification case). By no means would it be _required_ to run the installer to "reconfigure" Git for Windows--as most people would just use the CLI to modify their options to begin with (e.g. file system caching, or new line options notably come to mind), and possibly, during the next upgrade, modify their actual installation settings at that time.

In all of this, there are two particularly troublesome configuration settings in this installer:

- Modifications to be made to the user's `PATH` environment variable
    - Once set by the user during installation, it's not as easy as running some command to change your "configuration" with respect to this setting.
    - I experienced some weirdness with Windows 10 in trying to update my `PATH` environment via the System property pages in the control panel. Basically, they made it much nicer to edit various `PATH` parts, but apparently in the UI, there's a limit to the length of the `PATH`!! Yet, installers work just fine. That's because there's a Registry entry that holds this information, and it contains no such length limitation.
        - To get around this, I created a `GIT` environment variable that stored the paths I wanted in my `PATH` environment variable. Then I modified my `PATH` such that it contained `%GIT%`. However, I still had to manually update the Registry entry to accommodate this--I have a lot installed on my home machine.
    - Perhaps by making each `PATH` option a separate installation "feature", a user can "modify" their installation to quickly and easily change this, while also allowing the installer to "remember" this installation choice during subsequent upgrades and/or repairs.

- Currently, choosing to use Console Host as the terminal for `git-bash.exe` is a one-way operation. Especially where MSI is concerned, as noted back on the main issue.
    - It's on my backlog (in my head) to modify `edit-git-bash.exe` such that it can modify `git-bash.exe` as necessary. (I suppose it's written down now. :wink:)
        - This may have an added benefit of allowing others to experiment and contribute additional terminal hosts.

The rest of the installation (configuration) options are simply running various `git config` commands to modify their git system settings.

To late to be short, but I think it's worth playing around a bit with a few proof of concepts and see what shakes out.

dscho:
> Maybe I should've made these configuration actions come after the installation of the product?

I think that is a valid point.

Will try to find the time to review more, though, as you indicated you'd like to get this (including its strategy) merged first.

fourpastmidnight:
OK, so after doing some more research and starting down a path of a Proof of Concept, I've come to the conclusion that while what I've written _so far_ "works", it's not really correct. MSI Feature and Component rules are _HARD_ to keep straight and there are a lot of ins and outs and gotchas. So let's hold off on merging this still, and see where my latest research takes me.

I'm also fairly excited about all of this because it's going to bring new flexibility for installing Git for Windows (the ability to modify certain aspects of an existing installation) that we haven't had heretofore. All in all, your comments have instigated to really look hard into my current way of doing things--to take a step back and ask: Is this really the best approach?

I'm finding it's not. So thanks for not "rubber-stamping" this PR. :smile: (Not that you would ever "rubber stamp" any PR--you're pretty meticulous--and that's a compliment!!)

dscho:
> Not that you would ever "rubber stamp" any PR--you're pretty meticulous--and that's a compliment!!

:smile:

fourpastmidnight:
Who would've thought one little remark would've led me down such a long, but necessary path?

> Hmm. I would have hoped that the solution could be much, much simpler.
> --- dscho

Thanks!

I'm confident enough in my current approach to improving the MSI installer that I am closing this pull request.

Looking into using MSI Features for many of what were "installer properties" backing radio buttons and check boxes, has lead to a much more robust installer that properly supports installation, change, repair and removal scenarios. In addition, the changes should better support scripted/unattended installation scenarios, such as installation via GPO in corporate environments. Finally, I have been able to complete the rest of the "missing" features from the installer up to this point--and those missing features were very easily added due in part to using MSI features.

I'm still a few days to a week out from submitting a new PR, but this PR is no longer viable.

PR #126, opened 2016-08-11
Fixed the 'sports' typo in the Windows installer

gleerman:
The installer says:

> Git Bash will use MinTTY as terminal emulator, which **sports** a resizable window non-rectangular selections and a Unicode font.

This PR fixes this into "supports".

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1004,7 +1004,7 @@ begin
     with LblMinTTY do begin
         Parent:=BashTerminalPage.Surface;
         Caption:=
-            'Git Bash will use MinTTY as terminal emulator, which sports a resizable window,' + #13 +
+            'Git Bash will use MinTTY as terminal emulator, which supports a resizable window,' + #13 +
             'non-rectangular selections and a Unicode font. Windows console programs (such' + #13 +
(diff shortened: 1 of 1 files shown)


dscho:
This is not a typo. From Wiktionary: [*to sport*](https://en.wiktionary.org/wiki/sport#Verb): To display; to have as a notable feature.

Now, if you do not like the wording, I am open to improvements. However, "to support" is not what I wanted to say there. MinTTY does not merely "support" non-rectangular selections, that sounds as if it would allow other programs to provide that feature. Instead, the feature really is MinTTY's.

In short: I am mildly opposed to changing the sentence. But maybe you surprise me with something I really prefer to the current version?

gleerman:
Okay I see. Apologies. Not being a native English speaker, I perceived this as an error. 
As a suggestion, since others will perhaps feel the same way, we just could use "to feature" instead?

dscho:
> we just could use "to feature" instead?

That would work for me!

I still would need a Sign-off in the commit message, and please force-push to update this PR if you update the language in the installer.

Thanks!

dscho:
I feel there is no life left in this PR. Let's just close it.

PR #128, opened 2016-09-23
git-extra: add support for docm and dotm diffs

atavakoli:
The above file-types were added alongside the existing .doc/docx/dot
Word formats.


dscho:
Very good!

I updated the checksums locally by running `updpkgsums`. I also updated the `git-extra.install.in` file to edit existing `.gitattributes` files.

atavakoli:
Thank you!

PR #129, opened 2016-10-13
explorer integration: use icon from executable

kgybels:
Use the icon from the wrapper executable for the Windows Explorer
integration, this ensures that the executable and context menu show the
same icon, in addition to not depending on a separate file.


kgybels:
@dscho As promised, I verified the changes by building an installer, installing and verifying correct icons in context menu. However, when starting git bash, I get `bash: /etc/msystem: No such file or directory`, I doubt that has anything to do with these changes though.

dscho:
Whoops! I had forgotten about this PR, sorry. I think it is good to go, so I will merge it. If it causes a regression, I will simply revert it.

PR #133, opened 2017-02-25
git-lfs: add config for more efficent filter-process protocol

larsxschneider:
The same config is set via the `git lfs install` command:
https://github.com/git-lfs/git-lfs/blob/e029224251e2d8b134ee92f3cc00e001fdcbfb40/lfs/setup.go#L32-L35

However, I wonder if should rather use [`git lfs install --system`](https://github.com/git-lfs/git-lfs/blob/e029224251e2d8b134ee92f3cc00e001fdcbfb40/commands/command_install.go#L56) ?

--

This change is not tested. I just noticed the file and thought I could update it. I am pretty sure it works, though...

/cc @shiftkey

### DIFF
--- a/mingw-w64-git-lfs/git-lfs.install
+++ b/mingw-w64-git-lfs/git-lfs.install
@@ -3,6 +3,7 @@ post_install() {
     if [ -f "${arch}/bin/git-lfs.exe" ]; then
       git config -f "${arch}/etc/gitconfig" filter.lfs.clean "git-lfs clean -- %f"
       git config -f "${arch}/etc/gitconfig" filter.lfs.smudge "git-lfs smudge -- %f"
+      git config -f "${arch}/etc/gitconfig" filter.lfs.process "git-lfs filter-process"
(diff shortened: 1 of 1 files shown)


shiftkey:
> However, I wonder if should rather use `git lfs install --system`?

At the time I put together the original PR I'm pretty sure `--system` wasn't available at the time https://github.com/git-for-windows/build-extra/pull/110#issuecomment-210140619. So I'm :thumbsup: to using that here if it's now available.

dscho:
The problem is that `git-lfs` may not be installed when `git-extra`'s post-install script is run... so I think this PR is the best we can do.

Thanks for the contribution!

dscho:
whoops, forget my comment about `git-lfs` not being installed... I misread, and thought that this commit is about `git-extra`...

in any case, let's keep this as is for the moment, as it is not obvious what other changes `git lfs install` may want to make in the future.

dscho:
@larsxschneider do you want a 1.5.6-2 to be uploaded to Pacman?

PR #134, opened 2017-02-25
please.sh: add optional "commit" argument to test_remote_branch

larsxschneider:
If test_remote_branch is triggered from TravisCI then the branch name
might not be sufficient to identify the commit that TravisCI is
checking. Teach test_remote_branch an optional argument to specify
the exact commit hash that should be build and test. This commit hash
needs to be on the remote-tracking-branch that is specified as
mandatory argument.

/cc @dscho

### DIFF
--- a/please.sh
+++ b/please.sh
@@ -966,7 +966,7 @@ rebase () { # [--worktree=<dir>] [--test] [--redo] [--abort-previous] [--continu
 	exit
 }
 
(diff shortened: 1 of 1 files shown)


dscho:
Good idea, and excellent implementation (requiring the branch name and checking that the commit is actually part of that branch)!

Thank you for this contribution!

larsxschneider:
Thank you! How can I pass the `commit` parameter to a VSO Build?

PR #135, opened 2017-02-26
[WIP] Install curl-winssl as an alternative to curl built with openssl (#301)

Radrik5:
Implementation of approach suggested in https://github.com/git-for-windows/git/issues/301#issuecomment-282321452

Works as expected on Windows 7 64 bit.

Things to be done/disscussed:
- [ ] Test 32-bit version
- [ ] Write selected option to install-options.txt
- [ ] Set curl-winssl as default for fresh installations (upgrades should preserve previously selected value)
- [ ] Update please.sh to build & release both curl and curl-winssl at the same time
- [ ] Move curl-winssl binaries out of git-extra package to avoid extra releases each time cURL is released, maybe the code that unpacks curl-winssl can be placed in installer/release.sh


Radrik5:
curl-winssl may not work on pre-XP versions: https://github.com/curl/curl/commit/7475c471cf90a272ab0e35e28e842ad2e845a4fe

And lacks HTTP2 and HTTPS-proxy features:
```
curl $ /mingw64/bin/curl-winssl/curl.exe --version
curl 7.53.1 (x86_64-w64-mingw32) libcurl/7.53.1 WinSSL zlib/1.2.11 libssh2/1.8.0 librtmp/2.3
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smtp smtps telnet tftp
Features: IPv6 Largefile SSPI Kerberos SPNEGO NTLM SSL libz Metalink
curl $ /mingw64/bin/curl.exe --version
curl 7.53.0 (x86_64-w64-mingw32) libcurl/7.53.0 OpenSSL/1.0.2k zlib/1.2.11 libssh2/1.8.0 nghttp2/1.19.0 librtmp/2.3
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smtp smtps telnet tftp
Features: IPv6 Largefile SSPI Kerberos SPNEGO NTLM SSL libz TLS-SRP HTTP2 HTTPS-proxy Metalink
curl $ echo openssl && git clone https://github.com/git-for-windows/git.git
```
So maybe we should leave it disabled by default.

dscho:
> curl-winssl may not work on Windows <= XP

Git for Windows does not support XP any longer since v2.10.1...

> So maybe we should leave it disabled by default.

I agree.

dscho:
> git-extra: add curl-winssl

I think we want to skip this commit now. Better install the `mingw-w64-curl-winssl-bin` package as part of the regular installer/package building: https://github.com/git-for-windows/build-extra/blob/52c31c8a5b1839cabc1f3c92d6afc9761dc67cc9/make-file-list.sh#L40-L49 and https://github.com/git-for-windows/build-extra/blob/52c31c8a5b1839cabc1f3c92d6afc9761dc67cc9/make-file-list.sh#L58-L61

dscho:
> and https://github.com/git-for-windows/build-extra/blob/52c31c8a5b1839cabc1f3c92d6afc9761dc67cc9/make-file-list.sh#L58-L61

... actually, I take that back. I would like to include the winssl binaries *only* in the installer, not in PortableGit or anywhere else (because there is no easy way in that case to configure anything).

Instead, I would like to suggest including the files in the same way as [`edit-git-bash.exe`](https://github.com/git-for-windows/build-extra/blob/52c31c8a5b1839cabc1f3c92d6afc9761dc67cc9/installer/install.iss#L112) (i.e. with the `dontcopy` flag) and *overwriting* the existing `curl.exe`/`libcurl-4.dll` with the bundled versions if the user selects WinSSL.

And I agree with @dra27 that a separate page would be nice... It is definitely more of a hassle to add, I agree. See 0f1aba2085c26872c69b2cd8490a2831554e3327 for an example how to add a full new page.

Radrik5:
OK, I agree, having a separate page will be better. I spend an hour on this each morning, so something will probably be ready on Friday.

fourpastmidnight:
See the issue backing this PR for a nice dialog wireframe.

@dscho For MSI, I would create a feature. The user's feature choice would dictate which files get copied (and to where). Does InnoSetup provide a similar ability? This way, all the files are included in the installer, but depending on the user's choice, either one set of files or the other will actually be installed.

dscho:
> OK, I agree, having a separate page will be better. I spend an hour on this each morning, so something will probably be ready on Friday.

Awesome! Thank you.

Radrik5:
The code is ready but I've done only smoke testing on Windows 10 64 bit.
I'll test 32 bit and 64 bit versions tomorrow.

Here is how the new page looks:
image

```
$ curl --version
curl 7.53.1 (x86_64-w64-mingw32) libcurl/7.53.1 WinSSL zlib/1.2.11 libssh2/1.8.0 librtmp/2.3
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smtp smtps telnet tftp
Features: IPv6 Largefile SSPI Kerberos SPNEGO NTLM SSL libz Metalink

$ cat /etc/install-options.txt
Path Option: Cmd
SSH Option: OpenSSH
CURL Option: WinSSL
CRLF Option: CRLFAlways
Bash Terminal Option: MinTTY
Enable Symlinks: Disabled
```

Radrik5:
I've made minor changes to the code and tested 64-installer in the following cases:
* Fresh installation (OpenSSL is selected by default);
* Upgrade from previous version (OpenSSL is selected by default);
* Reinstallation (previously selected option is selected by default).

When OpenSSL version is installed, Git respects `http.sslCAInfo` config and to make it clone our internal repository I have to set `http.sslCAInfo` to a file that contains our internal root certificate. When WinSSL version is installed, Git works with our internal repository "out of the box" but ignores `http.sslCAInfo` config.

I haven't tested 32-bit installer because I haven't found a way how build it on a 64-bit system. Even in mingw32 shell `release.sh` builds 64-bin installer because `uname -m` returns `x86_64`.

I think the code is ready for review and merge, so I'm removing the [WIP] prefix.

PR #139, opened 2017-04-03
[WIP] packaging fixes for git-lfs

shiftkey:
Fixes https://github.com/git-for-windows/git/issues/1120

 - [x] deploy `git-lfs` to correct path
 - [ ] filter configuration values applied on install so that packaging Just Works™


dscho:
Thanks, looks good so far!

As to the configuration: the problematic part is here: https://github.com/git-for-windows/build-extra/blob/78f9f567e06b198ed35a19130c9244b52f4eb780/installer/install.iss#L1856-L1858. We essentially only keep copying `mingw64/etc/gitconfig` into the installer just to be able to populate `C:\ProgramData\Git\config`, unless that exists already.

What we would need to do is to weed out what settings we would like to keep in `mingw64/etc/gitconfig`. For the moment, I think this would be not only `filter.lfs` but also `pack.packSizeLimit`, `http.sslCAinfo`, and `diff.astextplain`.

I *guess* the quickest option would be to replace that `DeleteFile()` call by several `git config --remove-section` calls, namely to remove `core`, `color`, `help` and `rebase`.

A more thorough method would be to teach `installer/release.sh` to split the current `mingw64/etc/gitconfig` into two files based on a white list that says what we will want to keep from getting moved to `C:\ProgramData\Git\config`, include both files in the installer and initialize `C:\ProgramData\Git\config` (unless it exists already) by copying one of the two files.

If you do not get around to work on this until tomorrow, don't worry, I'll simply cherry-pick your two patches and then implement the more involved option of splitting the config into two parts.

shiftkey:
@dscho yeah, I'm kinda distracted for the next few days with other work - feel free to steal these commits if you get to it first...

dscho:
I'll steal them :-)

dscho:
Cherry-picked and squashed into 5f7d44728. Thanks!

zezba9000:
> filter configuration values applied on install so that packaging Just Works™

Was this done? Without this git-lfs clones will fail FYI.

dscho:
> > filter configuration values applied on install so that packaging Just Works™
> 
> This is the .gitconfig git-lfs filter settings stuff correct?

Yes. As I said earlier, in the SDK, `mingw64\etc\gitconfig` has all the `filter.lfs.*` settings. This file even was shipped with the installer, but it was only used to initialize a `C:\ProgramData\Git\config` if none was there yet. Otherwise it was simply thrown away. Now it is no longer thrown away, and the `filter.lfs.*` settings live happily in `mingw64\etc\gitconfig`, where they are picked up by Git for Windows.

zezba9000:
Sry, ic, tnx!

PR #140, opened 2017-04-03
Suppress command echoing at start

coladict:
Instead of telling each and every command not to get echoed to STDOUT, simply disable the echo at start.

### DIFF
--- a/sdk-installer/setup-git-sdk.bat
+++ b/sdk-installer/setup-git-sdk.bat
@@ -1,57 +1,58 @@
-@REM Set up the Git SDK
+@echo off
+REM Set up the Git SDK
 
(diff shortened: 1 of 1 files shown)


dscho:
Thank you for opening this Pull Request, @coladict!

However, we actually made it a habit in the Git for Windows project of *not* using wholesale `@echo off` statements, ever since https://github.com/msysgit/msysgit/commit/229a38207aea845b7bc37d27f13d1db7f45da82e.

Further, IIRC there was at least one part of that script that exposed problems with `@ECHO OFF`.

In any case, I am really reluctant to fix something that ain't broken :smile: I hope you understand!

PR #144, opened 2017-05-20
update git-lfs to address security vulnerability

shiftkey:
Details here: https://github.com/git-lfs/git-lfs/releases/tag/v2.1.1


dscho:
Oh, sorry, I did not look at the bug tracker before tending to my queue (which includes a job looking for new Git LFS versions, which I typically pick up as soon as possible). I merged this nevertheless as it is an additional confirmation of what I did previously.

PR #147, opened 2017-07-31
sdk-installer: shallow clone the git source

j16180339887:
Just clone the latest commit of the source.


dscho:
Good idea.

Would you mind adding your `Signed-off-by:` line to the commit message, and for good measure a size comparison between non-shallow vs shallow clone at the time of writing?

PR #148, opened 2017-08-03
Compare openDocument files

albfan:
Use odt2txt to compare libreoffice files

https://github.com/dstosberg/odt2txt

Should I provide a compiled version or ping the maintainer to do so?

### DIFF
--- a/git-extra/astextplain
+++ b/git-extra/astextplain
@@ -8,6 +8,9 @@ fi
 
 # XXX output encoding (UTF-8) hardcoded
 case "$1" in
+	*.ods | *.ODS | *.odf |*.ODF | *.odt | *.ODT)
+		odt2txt "$1" || cat "$1"
+		;;
 	*.doc | *.DOC | *.dot | *.DOT)
 		antiword -m UTF-8 "$1" | sed "s/\^M$//" || cat "$1"
 		;;
--- a/git-extra/git-extra.install
+++ b/git-extra/git-extra.install
@@ -42,6 +42,12 @@ GITCONFIG
 *.PDF	diff=astextplain
 *.rtf	diff=astextplain
 *.RTF	diff=astextplain
+*.ods	diff=astextplain
+*.ODS	diff=astextplain
+*odf	diff=astextplain
+*.ODF	diff=astextplain
+*odt	diff=astextplain
+*.ODT	diff=astextplain
 
 GITATTRIBUTES
 
--- a/git-extra/gitattributes
+++ b/git-extra/gitattributes
@@ -12,3 +12,9 @@
 *.PDF	diff=astextplain
 *.rtf	diff=astextplain
 *.RTF	diff=astextplain
+*.ods	diff=astextplain
+*.ODS	diff=astextplain
+*odf	diff=astextplain
+*.ODF	diff=astextplain
+*odt	diff=astextplain
+*.ODT	diff=astextplain
--- a/make-file-list.sh
+++ b/make-file-list.sh
@@ -66,8 +66,8 @@ required=
 for req in mingw-w64-$ARCH-git-credential-manager $SH_FOR_REBASE \
 	$(test -n "$MINIMAL_GIT" || echo \
 		mingw-w64-$ARCH-connect git-flow unzip docx2txt \
-		mingw-w64-$ARCH-antiword mingw-w64-$ARCH-xpdf ssh-pageant \
-		mingw-w64-$ARCH-git-lfs tig)
+		mingw-w64-$ARCH-antiword mingw-w64-$ARCH-odt2txt \
+		mingw-w64-$ARCH-xpdf ssh-pageant mingw-w64-$ARCH-git-lfs tig)
 do
 	test -d /var/lib/pacman/local/$req-[0-9]* ||
 	test -d /var/lib/pacman/local/$req-git-[0-9]* ||
@@ -84,7 +84,7 @@ then
 	packages="$packages mingw-w64-$ARCH-git-doc-html ncurses mintty vim
 		winpty less gnupg tar diffutils patch dos2unix which subversion
 		mingw-w64-$ARCH-tk mingw-w64-$ARCH-connect git-flow docx2txt
-		mingw-w64-$ARCH-antiword ssh-pageant mingw-w64-$ARCH-git-lfs tig"
+		mingw-w64-$ARCH-antiword mingw-w64-$ARCH-odt2txt ssh-pageant mingw-w64-$ARCH-git-lfs tig"
 fi
 pacman_list $packages "$@" |
 
@@ -106,7 +106,7 @@ grep -v -e '\.[acho]$' -e '\.l[ao]$' -e '/aclocal/' \
 	-e '^/mingw../itcl/' \
 	-e '^/mingw../t\(cl\|k\)[^/]*/\(demos\|msgs\|encoding\|tzdata\)/' \
 	-e '^/mingw../bin/\(autopoint\|[a-z]*-config\)$' \
-	-e '^/mingw../bin/lib\(asprintf\|gettext\|gnutlsxx\|pcre[0-9a-z]\|quadmath\|stdc++\)[^/]*\.dll$' \
+	-e '^/mingw../bin/lib\(asprintf\|gettext\|gnutlsxx\|pcre[0-9a-oq-z]\|quadmath\|stdc++\)[^/]*\.dll$' \
 	-e '^/mingw../bin/\(asn1\|gnutls\|idn\|mini\|msg\|nettle\|ngettext\|ocsp\|pcre\|rtmp\|xgettext\)[^/]*\.exe$' \
 	-e '^/mingw../.*/git-\(remote-testsvn\|shell\)\.exe$' \
 	-e '^/mingw../.*/git-cvsserver.*$' \
@@ -144,6 +144,7 @@ else
 		-e '^/mingw../bin/lib\(gcc_s_seh\|gmpxx\)-.*\.dll$' \
 		-e '^/mingw../bin/lib\(gomp\|jansson\|minizip\)-.*\.dll$' \
 		-e '^/mingw../bin/libvtv.*\.dll$' \
+		-e '^/mingw../bin/libpcreposix.*\.dll$' \
 		-e '^/mingw../bin/\(.*\.def\|update-ca-trust\)$' \
 		-e '^/mingw../bin/\(openssl\|p11tool\|pkcs1-conv\)\.exe$' \
 		-e '^/mingw../bin/\(psktool\|recode-.*\|sexp.*\|srp.*\)\.exe$' \


git-extra/astextplain
```diff
@@ -27,4 +30,4 @@ case "$1" in
 		;;
 esac
 
-exit 0
\ No newline at end of file
+exit 0
```
dscho: Please lose this hunk.

git-extra/astextplain
```diff
@@ -8,6 +8,9 @@ fi
 
 # XXX output encoding (UTF-8) hardcoded
 case "$1" in
+	*.ods | *.ODS | *odf |*.ODF | *odt | *.ODT)
+		odf2txt "$1" || cat "$1"
```
dscho: The program `odf2txt` does not exist in Git for Windows' SDK. It will need to be provided in a Pacman package before this PR can be merged.

dscho:
Please also add a Signed-off-by: line to the commit message, and it would be really good to add a comment to the commit message whether it affects only LibreOffice or also OpenOffice, and by how much this change increases the installer.

Thank you for working on this and contributing it!

git-extra/astextplain
albfan: I can provide a PKGBUILD for arch linux as I maintain several AUR packages, but need instructions of where to push that file.

An existing example would be great. (Cannot imagine how arch linux relates with git for windows, but really want to known how)

albfan: Do this help? https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/odt2txt

albfan:
Compiled code is 23Kb and man page is 2,3Kb

dscho:
> Compiled code is 23Kb and man page is 2,3Kb

Even the `.exe`? How did you compile it?

git-extra/astextplain
dscho: Pacman is not only for Arch Linux. Yes, it was born there, but the [MSYS2 project](https://msys2.github.io/) uses it, too. And Git for Windows' SDK is a custom version of MSYS2. Git for Windows is essentially a subset of the custom MSYS2 development environment.

You can learn much more about this here: https://github.com/git-for-windows/git/wiki/Package-management

As to an example of a recently-added MINGW package, how about this: https://github.com/git-for-windows/MINGW-packages/commit/898f5ed4bd2c6f5addd60aabb977051d97d0350f

dscho: You can, of course, use the PKGBUILD to which you linked as example, but please note a couple of crucial, required changes, as we are talking about a MINGW package:

- [`arch` must be `any`](https://github.com/git-for-windows/MINGW-packages/commit/898f5ed4bd2c6f5addd60aabb977051d97d0350f#diff-312c6a788fc5432403915ac0aa4090b6R10), as you can install 32-bit and 64-bit MINGW packages side-by-side,
- MINGW executables [install into `$MINGW_PREFIX/bin`](https://github.com/git-for-windows/MINGW-packages/commit/898f5ed4bd2c6f5addd60aabb977051d97d0350f#diff-312c6a788fc5432403915ac0aa4090b6R68), not `/usr/bin/`, and
- It will be `odt2txt.exe`, i.e. [you will have to add the `.exe` extension](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/odt2txt#n25)

Please also note that I'd much rather use a tag than a commit, if at all possible. If not, I'd rather use `master` [with a custom-generated `pkgver` reflecting the number of commits and the short commit name](https://github.com/git-for-windows/MINGW-packages/commit/898f5ed4bd2c6f5addd60aabb977051d97d0350f#diff-312c6a788fc5432403915ac0aa4090b6R19).

dscho: Oh, and there really is a need to use [`_realname`, `pkgbase` and `pkgname`](https://github.com/git-for-windows/MINGW-packages/commit/898f5ed4bd2c6f5addd60aabb977051d97d0350f#diff-312c6a788fc5432403915ac0aa4090b6R3), and you most likely will want to change the dependency to `${MINGW_PACKAGE_PREFIX}-zlib`. Otherwise you would depend on the *MSYS2* version of zlib, which would be totally useless (you cannot link a MINGW executable to an MSYS2 library).

albfan: With this example and the repo where to push I can finish this. Maybe need a couple of iterations for details. Let's see.

albfan:
No, about .exe I don't know size compiled from sources. I grab an exe from a freeware software that uses as  plugin. ExamDiff Pro. It is 959Kb for odt2txt.exe

Now with instructions, I can build and check.

How do I add size to commit, any example in old commits?

dscho:
> How do I add size to commit, any example in old commits?

You can use `git rebase -i` to rewrite previous commits. See e.g. https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

albfan:
Oh yes I had my own alias for that

https://github.com/albfan/dotfiles/blob/master/gitconfig#L51

I mean the exact kind of comment like 

```
78792a81167d34a45bb8629d87b276a6e930d3de
> The added package increases the installer size by 54K.
```

So I just need to work on arch linux PkGBUILD

albfan:
Compiling odt2txt for msys2: Tested on mingw64: https://github.com/Alexpux/MSYS2-packages/pull/964

I tried to improve original project too: https://github.com/dstosberg/odt2txt/pull/11

Any review is appreciated

albfan:
Size of arch linux package compiled is 27K

git-extra/astextplain
dscho: Please note that a MINGW package would be much preferable over an MSYS2 one.

albfan: Ok. What's the difference for those repos? Seems msys2 depends on mingw, am I right?

dscho: I tried to explain the difference here: https://github.com/git-for-windows/git/wiki/The-difference-between-MINGW-and-MSYS2 (sorry for the link, but I found that it is not an easy thing to explain, so I spent something like a dozen hours to try to explain it well, once and for all).

albfan:
Finally it is a MinGW package. Here the 32 bits package http://repo.msys2.org/mingw/i686/mingw-w64-i686-odt2txt-0.5-1-any.pkg.tar.xz

git-extra/astextplain
dscho: Well done, sir!

BTW I just realized that `odt2txt` tries to determine the encoding based on the current Win32 Console. Maybe we want to pass the `--encoding=UTF-8` option here? It seems to work better for me.

Also: we will have to measure the delta of the installer first before updating the commit message accordingly.

dscho:
> Finally it is a MinGW package. Here the 32 bits package http://repo.msys2.org/mingw/i686/mingw-w64-i686-odt2txt-0.5-1-any.pkg.tar.xz

Yeah!

Now only a couple things are left to do, methinks:

- [ ] [Build an installer](https://github.com/git-for-windows/git/wiki/Making-an-installer) to verify that you have a working SDK, and also to compare installer file sizes later.
- [ ] Add the `mingw-w64-$ARCH-odt2txt` packages as a non-MinGit dependency to [`make-file-list.sh`](https://github.com/git-for-windows/build-extra/blob/master/make-file-list.sh): I would simply look for `antiword` and append it there.
- [ ] Exclude the `zipcmp.exe`, `zipmerge.exe` and `ziptool.exe` files from being packages by appending yet another pattern to [the really long `grep` invocation](https://github.com/git-for-windows/build-extra/blob/44b7588e0be71674b3f05da019ce2032a6117c9e/make-file-list.sh#L121).
- [ ] Look into [the `--encoding=UTF-8` idea](https://github.com/git-for-windows/build-extra/pull/148/commits/6626e752f126e3ea1404fd0676fd0f36c4ec6220#r133418154).
- [ ] Build another installer and verify that it works correctly.
- [ ] Compare the installer file sizes and amend the commit message.

Normally, I would try to do at least a couple of these myself, but today is a *really* busy day for me, unfortunately. Maybe you can see how far you can get from here?

albfan:
Please let me do all that myself. Is not gonna be this week, but we all can wait a little more for this.

Feels great to contribute back without adding workload to maintainers.

I will report as soon as possible.

dscho:
Awesome, thanks so much! And please do ping me here should you get stuck with anything.

albfan:
I'll try to finish it this week

dscho:
Thanks for the update!

albfan:
I have some problem with creating the installer.
- What's the supposed version I should use?
- need to disable gpg check for mingw-w64-x86_64-git-lfs dependency. Cannot get the needed keys
- InnoSetup didn't was launched correctly from release.sh (it just hangs and don't do anything) I launched it manually and generates an exe of 45Mb

I will try to add the odt2txt and see if everything works

Next I will try those utf-8, etc...

dscho:
> What's the supposed version I should use?

It does not really matter. I typically use `2.14.1-without-feature` and `2.14.1-with-feature` or some variation thereof. You *may* get a downgrade warning when installing if you chose a bogus version, nothing to worry about.

>     need to disable gpg check for mingw-w64-x86_64-git-lfs dependency. Cannot get the needed keys

This is really vexing. I have no idea what causes this... the signature file is there: https://dl.bintray.com/git-for-windows/pacman/x86_64/mingw-w64-x86_64-git-lfs-2.2.1-1-any.pkg.tar.xz.sig. As a workaround, you may want to download it manually and put it into `var/cache/pacman/pkg/` before installing Git LFS again with `pacman -S mingw-w64-x86_64-git-lfs`.

>     InnoSetup didn't was launched correctly from release.sh (it just hangs and don't do anything)

It *may* look like that, but it works, I assure you. It just suppresses all the output because I got sick and tired of seeing so many lines whizz by.

> I launched it manually and generates an exe of 45Mb

Good. The size is different from the official installers because during the installation, the SDK builds Git (and installs it into the SDK) with debug symbols not stripped out.

If you want to get closer to the official installer, just install the official Pacman package:

```sh
pacman -S mingw-w64-x86_64-git
```

> I will try to add the odt2txt and see if everything works

Nice! Thanks!

albfan:
Nice tips.  Will check all this weekend

halabelhadj:
should i do that too? for Windows?

albfan:
I cannot generate installer

error: destino no encontrado: mingw-w64-x86_64-git-credential-manager
Could not install required packages:  mingw-w64-x86_64-git-credential-manager

I try to look for PKGBUILD here without luck

https://github.com/Microsoft/Git-Credential-Manager-for-Windows

albfan:
I found  mingw-w64-x86_64-git-credential-manager in this repo (my fault)

Now it ask for

```
error: package 'mingw-w64-x86_64-git' not found
error: package 'git-extra' not found
error: package 'mingw-w64-x86_64-git-doc-html' not found
error: package 'winpty' not found
cp: no se puede efectuar `stat' sobre '/mingw64/etc/gitconfig': No such file or directory
Could not split gitconfig
```

albfan:
I started to compile all by hand with `makepkg-mingw` (git-extra, git-for-windows-keyring, its deps) and now it ask for git-for-windows

I think I missed some step here

dscho:
Are you sure that you have an up-to-date Git for Windows SDK? In particular, your `/etc/pacman.conf` should end in these lines:

```
[...]
# The testing repositories are disabled by default. To enable, uncomment the
# repo name header and Include lines. You can add preferred servers immediately
# after the header, and they will be used before the default mirrors.

[git-for-windows]
Server = https://dl.bintray.com/$repo/pacman/$arch

[git-for-windows-mingw32]
Server = https://dl.bintray.com/git-for-windows/pacman/i686

[mingw32]
Include = /etc/pacman.d/mirrorlist.mingw32

[mingw64]
Include = /etc/pacman.d/mirrorlist.mingw64

[msys]
Include = /etc/pacman.d/mirrorlist.msys

# An example of a custom package repository.  See the pacman manpage for
# tips on creating your own repositories.
#[custom]
#SigLevel = Optional TrustAll
#Server = file:///home/custompkgs
```

If the `git-for-windows` or the `git-for-windows-mingw32` (or for 32-bit SDKs, `git-for-windows-mingw64`) repository is missing, you will want to call (a variation of) the steps described in https://github.com/git-for-windows/git/wiki/Install-inside-MSYS2-proper

albfan:
I'm sure is related with that Update sdk is done installing https://git-for-windows.github.io/#download-sdk?

dscho:
@albfan that sdk installer should set you up correctly in the first place... Are you saying that your `/etc/pacman.conf` looks differently?

albfan:
My /etc/pacman.conf is the same. I guess this is related with running the ./installer/release.sh from mingw64 (bash from git).

Running from msys2 it says I have all the necessary packages. At this time I don't know if that is because I did manually or what. Anyway I get progress.

Now the error is:

```
Error on line 4 in installer\file-list.iss: Source file ":cmd\git.exe" does not exist.
Compile aborted
Could not make installer
```

I have a "cmd\git.exe" into "C:\git-sdk-64" or into a "C:\msys64" or into "C:\Program files\Git\" or into  "C:\Program files\Git\mingw64"

should I configure any env variable?

Thanks for all the support

dscho:
You should not run any of this from MSYS2. All of build-extra's scripts are designed to run from Git for Windows' SDK's `git-bash.exe`.

albfan:
Don't have a clue of what is happening

This is a screencast of my problems (both on msys2 or in mingw64)

https://www.youtube.com/watch?v=P_aqocml_B8

dscho:
@albfan hmm. If you are calling Git for Windows' SDK's `git-bash.exe`, then the appropriate `build-extra` should be in `/usr/src/build-extra/`... and in that case, I would not expect either problem to occur.

The first problem seems to be that `please.sh` fails to figure out that MSYS2's pseudo-root directory is not detected correctly. I verified that this is a problem with `build-extra` checked out *outside* of the pseudo-root... I'll look into that, but in the meantime you should be able to use `/usr/src/build-extra/` instead.

The other problem, i.e. the "cygwin1.dll" one (which is actually an `msys-2.0.dll` one) is due to a mismatch between your installed Git's MSYS2 runtime and the MSYS2 runtime of your Git for Windows SDK.

Another thing I noticed: your window title says "SDK-MSYS", but it should say "SDK-64" instead. Did you start it via `git-bash.exe`?

dscho:
> I verified that this is a problem with build-extra checked out outside of the pseudo-root... I'll look into that

Found it: https://github.com/git-for-windows/build-extra/blob/0a1bef2e3c5f930bf163900bb2f0a5f8cf18e9dc/installer/install.iss#L40 (this really hardcodes the assumption that `build-extra` is checked out into `/usr/src/build-extra`, relative to MSYS2's pseudo root directory.

dscho:
> installer: allow running from arbitrary checkout locations

I just fixed the problem that prevented your first run (where you cloned build-extra into your home directory), pushed it to `master` and also cherry-picked it into your branch, as you can see. After pulling from your own `patch-1` branch, it should work in your SDK.

albfan:
Let me check again. Thanks for triaging. Hope all this leads to new contributions

dscho:
> Hope all this leads to new contributions

Oh no worries, we'll get your changes integrated. Just a little patience and a little tenacity will get it done.

albfan:
:( I cannot check in on https://github.com/git-for-windows/build-extra/pull/148#issuecomment-322737447 checklist

I rework patch (there's no odf2txt but odt2txt)

Now I have a problem (git diff says cannot diff files)

trying directly odt2txt on file it says

    error while loading shared libraries: libpcreposix-0.dll: cannot open shared object: No such file or directory

I have other odt2txt.exe and size is really different 62Kb to 959Kb. With that on path everything works. I have to check what's the problem with this one.

I have a repo with odt files with changes, to check feature, but wonder if there's any testsuite so we can cover this feature with a test. Like:

    git diff --no-index fileA.odt fileB.odt > result
    result equals expected_result

albfan:
$ pkgfile libpcreposix-0.dll
    mingw32/mingw-w64-i686-pcre
    mingw64/mingw-w64-x86_64-pcre

I think this line is removing pcre libs:
https://github.com/git-for-windows/build-extra/blob/master/make-file-list.sh#L109

Can I remove pcre from list of removed libs?

dscho:
> I have a repo with odt files with changes, to check feature, but wonder if there's any testsuite so we can cover this feature with a test.

We do not really have any integration tests yet, no. But maybe you want to start that? We could add shell/Powershell/CMD scripts to, say, `integration/` in `build-extra.git`, for example.

> trying directly odt2txt on file it says
> 
> ```
> error while loading shared libraries: libpcreposix-0.dll: cannot open shared object: No such file or [...]
> ```

Booh!

> I think this line is removing pcre libs:
> https://github.com/git-for-windows/build-extra/blob/master/make-file-list.sh#L109

I think you are correct.

> Can I remove pcre from list of removed libs?

Maybe not all of them? Maybe just replace the `a-z` by `a-oq-z`? There are currently many `libpcre*` files in the `bin` directory:

```sh
$ ls /mingw64/bin/libpcre*
/mingw64/bin/libpcre-1.dll*      /mingw64/bin/libpcre2-posix-1.dll*
/mingw64/bin/libpcre16-0.dll*    /mingw64/bin/libpcre32-0.dll*
/mingw64/bin/libpcre2-16-0.dll*  /mingw64/bin/libpcrecpp-0.dll*
/mingw64/bin/libpcre2-32-0.dll*  /mingw64/bin/libpcreposix-0.dll*
/mingw64/bin/libpcre2-8-0.dll*
```

Also, there is a clause for MinGit right after the one that excludes the libpcre[0-9a-z]* files, and I would think that we would want to exclude `libpcreposix.*` for MinGit (because it does not want such helpers as `odt2txt`). I.e. something like this:

```diff
diff --git a/make-file-list.sh b/make-file-list.sh
index 3935c9fb1..8cda6a53e 100755
--- a/make-file-list.sh
+++ b/make-file-list.sh
@@ -106,7 +106,7 @@ grep -v -e '\.[acho]$' -e '\.l[ao]$' -e '/aclocal/' \
 	-e '^/mingw../itcl/' \
 	-e '^/mingw../t\(cl\|k\)[^/]*/\(demos\|msgs\|encoding\|tzdata\)/' \
 	-e '^/mingw../bin/\(autopoint\|[a-z]*-config\)$' \
-	-e '^/mingw../bin/lib\(asprintf\|gettext\|gnutlsxx\|pcre[0-9a-z]\|quadmath\|stdc++\)[^/]*\.dll$' \
+	-e '^/mingw../bin/lib\(asprintf\|gettext\|gnutlsxx\|pcre[0-9a-oq-z]\|quadmath\|stdc++\)[^/]*\.dll$' \
 	-e '^/mingw../bin/\(asn1\|gnutls\|idn\|mini\|msg\|nettle\|ngettext\|ocsp\|pcre\|rtmp\|xgettext\)[^/]*\.exe$' \
 	-e '^/mingw../.*/git-\(remote-testsvn\|shell\)\.exe$' \
 	-e '^/mingw../.*/git-cvsserver.*$' \
@@ -144,6 +144,7 @@ else
 		-e '^/mingw../bin/lib\(gcc_s_seh\|gmpxx\)-.*\.dll$' \
 		-e '^/mingw../bin/lib\(gomp\|jansson\|minizip\)-.*\.dll$' \
 		-e '^/mingw../bin/libvtv.*\.dll$' \
+		-e '^/mingw../bin/libpcreposix.*\.dll$' \
 		-e '^/mingw../bin/\(.*\.def\|update-ca-trust\)$' \
 		-e '^/mingw../bin/\(openssl\|p11tool\|pkcs1-conv\)\.exe$' \
 		-e '^/mingw../bin/\(psktool\|recode-.*\|sexp.*\|srp.*\)\.exe$' \
```

> :( I cannot check in on #148 (comment) checklist

I moved the checklist to the next comment, which you should be able to update: https://github.com/git-for-windows/build-extra/pull/148#issuecomment-322751491

albfan:
Will try again with those fixes

`integration/` is a journey I really want to embark on. Let's start small with this feature and see how it goes (maybe I have to replicate existing CI)

dscho:
> integration/ is a journey I really want to embark on.

Nice!

> Let's start small with this feature and see how it goes

I agree, that's the best way to do it.

If you have any idea how I could automate https://github.com/git-for-windows/build-extra/blob/master/installer/checklist.txt, I would *love* to have that in there, too...

albfan:
Ok, last rebase seems to work

I found this and don't know if I should add to patch

```
$ git diff
diff --git i/git-extra/git-extra.install w/git-extra/git-extra.install
index e743842..2cefba4 100644
--- i/git-extra/git-extra.install
+++ w/git-extra/git-extra.install
@@ -42,6 +42,12 @@ GITCONFIG
 *.PDF  diff=astextplain
 *.rtf  diff=astextplain
 *.RTF  diff=astextplain
+*.ods  diff=astextplain
+*.ODS  diff=astextplain
+*odf   diff=astextplain
+*.ODF  diff=astextplain
+*odt   diff=astextplain
+*.ODT  diff=astextplain

 GITATTRIBUTES

diff --git i/git-extra/gitattributes w/git-extra/gitattributes
index 14637c8..e496ee3 100644
--- i/git-extra/gitattributes
+++ w/git-extra/gitattributes
@@ -12,3 +12,9 @@
 *.PDF  diff=astextplain
 *.rtf  diff=astextplain
 *.RTF  diff=astextplain
+*.ods  diff=astextplain
+*.ODS  diff=astextplain
+*odf   diff=astextplain
+*.ODF  diff=astextplain
+*odt   diff=astextplain
+*.ODT  diff=astextplain
```

Don't know if it is related but seems git push fails with 

    cannot set backend SSL to 'openssl' already set

Can you check if it fails for you

I will create another PR with integration. There'are some fails I cannot fix and some technology decissions to discuss

albfan:
See #159 for integration tests.

dscho:
> Don't know if it is related but seems git push fails with
> 
> cannot set backend SSL to 'openssl' already set

This means that you have an older Git version and somewhere your config sets this twice. Find out where with `git config --get-all --show-origin http.sslbackend`, and if you upgrade your Git for Windows, it should work anyway.

> See #159 for integration tests.

Awesome!

> I found this and don't know if I should add to patch [diff to gitattributes]

Yes, I think this should be part of the commit.

BTW is it really true that this adds 925kB (i.e. almost one megabyte to a ~34MB package)? That is rather a bit larger than I hoped for. If that is true, do you know on what files this huge increase is spent?

albfan:
I did the comparison with your installer. If I made comparison with a installer made from parent commit of this PR, difference is only 86Kb

I've added the gitattributtes changes too

After install with my installer:
```
$ git config --get-all --show-origin http.sslbackend
file:"C:\\Program Files\\Git\\mingw64/etc/gitconfig"    openssl
```

but problem with ssl arises again

After reinstall with your installer, problem vanished

dscho:
> I did the comparison with your installer. If I made comparison with a installer made from parent commit of this PR, difference is only 86Kb

Excellent!

> After reinstall with your installer, problem vanished

Ah, that means that the Git version you have installed in your SDK is a bit old ;-)

Thank you so much for your contribution. I tip my hat for your tenacity. Well done, sir!

albfan:
This story starts with an itch. My pleasure

PR #149, opened 2017-08-04
sdk-installer: shallow clone the git source

j16180339887:
Just clone the latest commit of the source.

.git size of a full clone   : 143 MB (150,062,380 bytes)
.git size of a shallow clone: 7.52 MB (7,886,193 bytes)


dscho:
Thanks!

PR #151, opened 2017-08-10
git-update - Keep Git for Windows up to date via the command-line

between2spaces:
A shell script that makes keeping Git for Windows up to date as simply as # git-update --yes

Compares what version is currently installed against the latest available release.

If versions differ, the bit matched installer is downloaded and run when confirmation to do so is given.

Discussed in Git issue https://github.com/git-for-windows/git/issues/1256

### DIFF
--- a/git-extra/git-update
+++ b/git-extra/git-update
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+# Compares the currently installed Git for Windows against latest available release.
+# If versions differ, the bit matched installer is downloaded and run when confirmation
+# to do so is given.
+
+git_update () {
+    yn="$1"
+    case $yn in
+        -\? | --?\?) echo "Usage: git-update [options]";
+        echo "Options:";
+        echo "  -y, --yes       Automatic yes to download and install prompt";
+        return;;
+    esac
+    bit=$(ls --color=never -d /mingw* | sed 's/\/mingw//')
+    releases=$(curl --silent https://api.github.com/repos/git-for-windows/git/releases/latest)
+    latest=$(echo "$releases" | grep '"tag_name": "v' | sed -E 's/.*"tag_name": "v([^"]*).*/\1/')
+    version=$(git --version | sed "s/git version //")
+    echo "Git for Windows $version (${bit}bit)"
+    if [ "$latest" != "$version" ]
+    then
+        echo "Update $latest is available"
+        download=$(echo "$releases" | grep '"browser_download_url": "' | grep "$bit\-bit\.exe" | sed -E 's/.*": "([^"]*).*/\1/')
+        filename=$(echo "$download" | sed -E 's/.*\/([^\/]*)$/\1/')
+        installer=$(mktemp -t gfw-install-XXXXXXXX.exe)
+        case $yn in
+            -[Yy]* | --[Yy]*) yn="y"; echo "Downloading $filename";;
+            *) read -p "Download $filename and install [N/y]? " yn;;
+        esac
+        case $yn in
+            [Yy]* ) curl -# -L -o $installer $download; start "" "$installer" /SILENT; ps | grep bash | awk '{print "kill -9 " $2 ";" }' | sh;;
+        esac
+        return
+    fi
+    echo "Up to date"
+}
+
+git_update $*
+
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -39,7 +39,8 @@ source=('inputrc'
         '99-post-install-cleanup.post'
         'gitattributes'
         'astextplain'
-        'git-sdk.sh')
+        'git-sdk.sh'
+        'git-update')
 md5sums=('3fab57079f0322efe256e95fe29d516f'
          'a2d3b63eb1c88c2a2b011ccf0b219d60'
          'bfb591886b2a28af3334521e71198b74'
@@ -58,7 +59,8 @@ md5sums=('3fab57079f0322efe256e95fe29d516f'
          'bdeb6046510aa1a6cd76e7f86673aa3f'
          '52c242d0c84d6934ec6a94766f8a70f0'
          'e87a5075629b58fe329094b92c1d13db'
-         '593dbce30c442529b4351650f4422b6e')
+         '593dbce30c442529b4351650f4422b6e'
+         'da5db7644f0aff6567185a5495435415')
 
 prepare() {
   test $startdir/$pkgname.install -nt $startdir/$pkgname.install.in &&
@@ -102,4 +104,5 @@ package() {
   install -m644 msys2-32.ico $pkgdir/usr/share/git
   install -m644 99-post-install-cleanup.post $pkgdir/etc/post-install
   install -m755 astextplain $pkgdir/usr/bin
+  install -m755 git-update $pkgdir/${MINGW_PREFIX}/libexec/git-core
 }


nalla:
If you now add the script to the sources inside the `PKGBUILD` plus the missing `md5` and further down add a line like

```
install -m755 git-update $pkgdir/${MINGW_PREFIX}/libexec/git-core
```

The updater script will automatically be available out of the box in the next release ;)

git-extra/git-update
```diff
@@ -0,0 +1,38 @@
...
+        echo "Options:";
+        echo "  -y, --yes       Automatic yes to download and install prompt";
+        return;;
+    esac
+    local bit=$(ls --color=never -d /mingw* | sed 's/\/mingw//')
+    local releases=$(curl --silent https://api.github.com/repos/git-for-windows/git/releases/latest)
+    local latest=$(echo "$releases" | grep '"tag_name": "v' | sed -E 's/.*"tag_name": "v([^"]*).*/\1/')
+    local version=$(git --version | sed "s/git version //")
+    echo "Git for Windows $version (${bit}bit)"
+    if [ "$latest" == "$version" ]
```
ViceIce: I think the compare is wrong?

between2spaces:
@ViceIce Oops. that was a debug/test setup. I've corrected it now. Thanks.

@nalla Thanks for that info. I've made the additional changes you suggested.

dscho:
Can I ask for two more things?

- could you add a *Signed-off-by:* line to the commit message, and maybe describe the tool a little in the commit message, too?

- could you mark the script with `#!/bin/bash`, as `local` is *not* Unix shell but Bash-specific?

landstander668:
>could you mark the script with #!/bin/bash, as local is not Unix shell but Bash-specific?

Since this has morphed from a user defined shell function to a standalone script, the `local` attribute on variables shouldn't be required. It can be important in the original case, in order to avoid clobbering variables which were set outside the function, but that's not really a concern here since scripts are normally run in a subshell.

dscho:
Makes a ton of sense to me to drop the `local` keyword from `git-update`. I still would like to ask for that [*Signed-off-by* line](https://github.com/git/git/blob/v2.14.1/Documentation/SubmittingPatches#L254-L306)... :smile:

git-extra/git-update
```diff
@@ -0,0 +1,38 @@
...
+    then
+        echo "Update $latest is available"
+        local download=$(echo "$releases" | grep '"browser_download_url": "' | grep "$bit\-bit\.exe" | sed -E 's/.*": "([^"]*).*/\1/')
+        local filename=$(echo "$download" | sed -E 's/.*\/([^\/]*)$/\1/')
+        case $yn in
+            -[Yy]* | --[Yy]*) yn="y"; echo "Downloading $filename";;
+            *) read -p "Download $filename and install [N/y]? " yn;;
+        esac
+        case $yn in
+            [Yy]* ) curl -# -L -o $filename $download; start $filename; ps | grep bash | awk '{print "kill -9 " $2 ";" }' | sh;;
```
larsxschneider: You are downloading the installer into the current directory. I would create a tempfile like this...
```
installer=$(mktemp -t "gfw-install-XXXXXXX.exe")
```
... and then download the file to this location and `start` it.

larsxschneider: Also, you might want to run the installer in silent mode like this:
```
start "" "$filename" /SILENT
```

between2spaces:
Updates merged into a new single commit.
* Added more of a description in commit message and added a Signed-off-by line
* Removed local keyword
* Installer saved to templated temp file
* Installer runs with /SILENT option

dscho:
Very nice, thanks!

git-extra/git-update
```diff
@@ -0,0 +1,39 @@
...
+        echo "Update $latest is available"
+        download=$(echo "$releases" | grep '"browser_download_url": "' | grep "$bit\-bit\.exe" | sed -E 's/.*": "([^"]*).*/\1/')
+        filename=$(echo "$download" | sed -E 's/.*\/([^\/]*)$/\1/')
+        installer=$(mktemp -t gfw-install-XXXXXXXX.exe)
+        case $yn in
+            -[Yy]* | --[Yy]*) yn="y"; echo "Downloading $filename";;
+            *) read -p "Download $filename and install [N/y]? " yn;;
+        esac
+        case $yn in
+            [Yy]* ) curl -# -L -o $installer $download; start "" "$installer" /SILENT; ps | grep bash | awk '{print "kill -9 " $2 ";" }' | sh;;
```
larsxschneider: @between2spaces Do you see an easy way to check the hash of the downloaded installer? Something like ...
```
 echo "$SHA256  $installer" | sha256sum --status --check -
```
... would work. I wonder how much sense it would make to download/parse the hashes that @dscho posts on GitHub and compare them that way.

between2spaces: @larsxschneider The following command seems to isolate the published sha256 sum, where ${bit} is the calculated bitness of your current installation.

SHA256=$(curl --silent --location https://github.com/git-for-windows/git/releases/latest | awk '/<td>/' | sed 'N;s/\n/ /' | grep "${bit}-bit\.exe" | sed -E 's/.*<td>([^<]+)<\/td>/\1/')

i.e.
\# curl --silent --location https://github.com/git-for-windows/git/releases/latest | awk '/<td>/' | sed 'N;s/\n/ /' | grep "64-bit\.exe" | sed -E 's/.*<td>([^<]+)<\/td>/\1/'

Problem is, wouldn't parsing the expected sum from HTML be less reliable than curl successfully downloading and writing an uncorrupted binary to disk?

If the parse ever goes wrong the correct sum won't be stored in $SHA256 and the update script will incorrectly abort.

larsxschneider: You are right. Parsing the data is somewhat fragile. I wonder if you could get the same info using the [release API](https://developer.github.com/v3/repos/releases/#get-the-latest-release)?

Plus, we don't need to abort. We could just warn the user. I wonder what @dscho thinks. Is this worth the effort? A premium solution could be if @dscho would sign the builds and ship his public key with GfW. Then the update mechanism could check that nobody has tempered with the installer. But this would add more work for @dscho and he has probably more important items on his todo list.

dscho: I already sign the installer with my code-signing key. Is that not enough?

larsxschneider: Ah. It took a while until I understood it. You are using the Windows install signing mechanism. I assume this mechanism would also scream if the installer contains corrupted data, right? In that case all that I have written above is irrelevant. Thanks!

dscho: It is not the *install* signing mechanism, it is the mechanism where Windows warns loudly if unsigned `.exe` files are to be executed, or if the signature is incorrect.

As to forging a code-signing certificate so it shows my name? I can testify to the fact that it is a *major* hassle to get a code-signing certificate due to the identity verification requirements by companies like Comodo that are highly incompatible with international law (they expect you to abide by what is possible by US law, no matter where you live). If anybody manages to forge a code-signing certificate in my name, they went through hell, and back, so they may deserve their little success.

AuthorProxy:
Will be better if this command support determining if portable git is installed and will update portable folder, not the global (which not exists)

dscho:
@AuthorProxy yes, the idea was to support official Git for Windows installers first, and then hope for a contributor who extends this to portable Gits.

PR #152, opened 2017-08-15
git-extra: Allow non-ASCII characters from $PWD in window title

elieux:
Seems like this limitation is artificial at this point (it's been there since the beginning of the repository).

image

Not updating `PKGBUILD` since `updpkgsums` showed other changes and I guess there's a different process to it.


dscho:
Thanks!

Indeed, I do not require contributors to update the checksums. I do that as part of the release engineering.

PR #153, opened 2017-08-23
feat(installer): Git for Windows Updater

bufferoverflow:
This is a followup on https://github.com/git-for-windows/git/issues/1256 adding git-update created via https://github.com/git-for-windows/build-extra/pull/151 to the installer with a task that is scheduled daily.

Installer looks like this:
2017-08-23 21_08_20-git 2 14 2 setup

This dialog appears if some updates are available:
2017-08-24 09_24_09-taskeng exe

Tested with the following settings:
2017-08-25 12_15_44-local area network lan settings

Todo:
* verification within multiple environments

### DIFF
--- a/git-extra/git-update
+++ b/git-extra/git-update
@@ -5,6 +5,13 @@
 # when confirmation to do so is given.
 
 git_update () {
+	proxy=$(proxy-lookup https://api.github.com | grep URL | awk '{ print $4}' | awk -F";" '{print $1}')
+	if test "$proxy" != ""
+	then
+		export https_proxy="http://$proxy"
+		echo "Using proxy server $https_proxy detected from Internet Explorer settings"
+	fi
+
 	yn=
 	while test $# -gt 0
 	do
--- a/msi/release.sh
+++ b/msi/release.sh
@@ -124,6 +124,10 @@ die "Could not generate ReleaseNotes.html."
 make -C ../ edit-git-bash.exe ||
 die "Could not build edit-git-bash.exe."
 
+# Compile proxy-lookup.exe
+make -C ../ proxy-lookup.exe ||
+die "Could not build proxy-lookup.exe."
+
 # Make a list of files to include
 LIST="$(ARCH=$ARCH BITNESS=$BITNESS \
 	PACKAGE_VERSIONS_FILE="$SCRIPT_PATH"/package-versions.txt \
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.swp
 /ReleaseNotes.html
 edit-git-bash.exe
+proxy-lookup.exe
 /cached-source-packages/
 /download-stats.ids
 /git-extra/pkg/
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 CFLAGS=-O2 -Wall
 
-all: edit-git-bash.exe
+all: edit-git-bash.exe proxy-lookup.exe
 
 edit-git-bash.exe: edit-git-bash.c
 	gcc -DSTANDALONE_EXE $(CFLAGS) -o $@ $^
+
+proxy-lookup.exe: proxy-lookup.c
+	gcc $(CFLAGS) -Werror -o $@ $^ -lshell32 -lwinhttp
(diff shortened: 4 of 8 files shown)


git-extra/git-update
```diff
@@ -5,6 +5,13 @@
 # when confirmation to do so is given.
 
 git_update () {
+	proxy=$(proxy-lookup https://api.github.com | grep URL | awk '{ print $4}' | awk -F";" '{print $1}')
+	if test "$proxy" != ""
+	then
+		export https_proxy="http://$proxy"
+		echo "Using proxy server $https_proxy detected from Internet Explorer settings"
+	fi
+
```
dscho: Would it work if you set `CURL_SSL_BACKEND=schannel` specifically for the `curl.exe` invocation and *not* try to discover the proxy ourselves?

bufferoverflow: I've tested that configuration. libcurl does not automatically detect proxy when using schannel backend. see also 
https://curl.haxx.se/docs/faq.html#Does_curl_support_Javascript_or

dscho: That is funny, because all the claims I read on the internets (e.g. https://stackoverflow.com/questions/1600743/could-not-create-ssl-tls-secure-channel-could-the-problem-be-a-proxy-server) seem to suggest that Secure Channel respects the Internet Options, including the proxy setting...

bufferoverflow: As far as I understood until now, schannel does only SSL/TLS, see https://msdn.microsoft.com/en-us/library/windows/desktop/aa380123(v=vs.85).aspx

dscho: Ah, you're right, of course, it is cURL's duty to handle the higher-order aspects of the HTTPS protocol (which a proxy kinda is).

I wonder, however, if we can do better than this and integrate that proxy lookup into cURL. And indeed, we talked about this before: https://github.com/git-for-windows/git/issues/387

And of course you know this because you referenced that very issue in the header of the file...

Now, I wonder further how any of your `git fetch` calls work if you have a proxy and not set the  environment variable `https_proxy` globally... And my guess is that you use the Git config variable `http.proxy` to this end.

So naturally, I wonder, whether we should not use that `http.proxy` config setting here, too, instead of adding a helper to look up the system proxy setting *just* for `git update` and for nothing else.

msi/release.sh
```diff
@@ -124,6 +124,10 @@ die "Could not generate ReleaseNotes.html."
 make -C ../ edit-git-bash.exe ||
 die "Could not build edit-git-bash.exe."
 
+# Compile proxy-lookup.exe
+make -C ../ proxy-lookup.exe ||
+die "Could not build proxy-lookup.exe."
```
dscho: I think if you do build it in `msi/release.sh`, you probably also want to bundle it here.

An easier way might be to simply have the file installed into the correct location via the `git-extra` package, [just like `create-shortcut.exe`](https://github.com/git-for-windows/build-extra/blob/master/git-extra/PKGBUILD#L98).

git-extra/git-update
bufferoverflow: Yes, having proxy detect directly within curl and git would be my favorite => no http.proxy setting within git and no proxy variable for curl.

Using `http.proxy` could be an option, however updater will fail if it is not configured...

msi/release.sh
dscho: Could you do that? Preferably as a separate commit that needs to be reordered to be th first co|mit in this PR

git-extra/git-update
dscho: I feel we should go for the `http.proxy`-only way (being careful *not* to set `https_proxy` if it has been already set, or if `http.proxy` is not configured or has an empty value).

We can then revive the effort to get automatic proxy.detection into cURL.

dscho: (part of my rationale: if `http.proxy` is not configured but required, cloning/fetching via HTTPS will already fail)

bufferoverflow: @dscho I've made https://github.com/git-for-windows/build-extra/pull/155 which uses `http.proxy` and no proxy-lookup magic.

dscho: For the record: if you force-push your branch, that updates the original Pull Request. No need to add new ones...

PR #155, opened 2017-09-08
Feat/update scheduler v2

bufferoverflow:
This replaces https://github.com/git-for-windows/build-extra/pull/153 and uses git `http.proxy` settings as suggested by @dscho

### DIFF
--- a/git-extra/git-update
+++ b/git-extra/git-update
@@ -5,6 +5,13 @@
 # when confirmation to do so is given.
 
 git_update () {
+	proxy=$(git config --get http.proxy)
+	if test -n "$proxy"
+	then
+		export https_proxy="$proxy"
+		echo "Using proxy server $https_proxy detected from git http.proxy" >&2
+	fi
+
 	yn=
 	while test $# -gt 0
 	do
@@ -39,7 +46,7 @@ git_update () {
 		return
 	fi
 
-	echo "Update $latest is available"
+	echo "Update $latest is available" >&2
 	download=$(echo "$releases" |
 		grep '"browser_download_url": "' |
 		grep "$bit\-bit\.exe" |
@@ -55,7 +62,7 @@ git_update () {
 		else
 			warn=" (killing $other_bashes other Bash)"
 		fi
-		read -p "Download $filename and install$warn [N/y]? " yn
+		read -p "Download $filename and install$warn [N/y]? " yn >&2
 		case "$yn" in
 		[Yy]*) ;;
 		*) return 1;;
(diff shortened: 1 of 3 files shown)


git-extra/git-update
```diff
@@ -5,6 +5,13 @@
 # when confirmation to do so is given.
 
 git_update () {
+	proxy=$(git config --get http.proxy)
+	if test "$proxy" != ""
```
dscho: Could you use `test -n "$proxy"` instead, please?

git-extra/git-update
```diff
@@ -5,6 +5,13 @@
...
+	then
+		export https_proxy="http://$proxy"
```
dscho: From https://git-scm.com/docs/git-config:

> http.proxy::
>   Override the HTTP proxy, normally configured using the 'http_proxy',
>   'https_proxy', and 'all_proxy' environment variables (see `curl(1)`). In
>   addition to the syntax understood by curl, it is possible to specify a
>   proxy string with a user name but no password, in which case git will
>   attempt to acquire one in the same way it does for other credentials. See
>   linkgit:gitcredentials[7] for more information. The syntax thus is
>   '[protocol://][user[:password]@]proxyhost[:port]'. This can be overridden
>   on a per-remote basis; see remote.<name>.proxy

While I think that we cannot replicate the "user name without password triggers credentials to be acquired" idea, I am fairly certain that we should not prefix the value of `http.proxy` with `http://`. Could you delete that prefix, please?

installer/install.iss
```diff
@@ -102,6 +102,14 @@ Name: consolefont; Description: Use a TrueType font in all console windows
...
 Filename: {app}\git-bash.exe; Parameters: --cd-to-home; Description: Launch Git Bash; Flags: nowait postinstall skipifsilent runasoriginaluser unchecked
 Filename: {app}\ReleaseNotes.html; Description: View Release Notes; Flags: shellexec skipifdoesntexist postinstall skipifsilent
+Filename: "schtasks"; \
+    Description: Daily check for available update using git http.proxy if defined; \
+    Parameters: "/Create /F /SC DAILY /TN ""Git for Windows Updater"" /TR ""'{app}\bin\git.exe' update"""; \
+    Flags: runhidden postinstall
+
+[UninstallRun]
+Filename: "schtasks"; \
+    Parameters: "/Delete /F /TN ""Git for Windows Updater"""; \
```
dscho: Very nice.

dscho:
Thank you!

I actually just realized that I can force-push updates to your branch, so i just went ahead and made the modifications myself. While at it, I also redirected a couple more messages to stderr.

bufferoverflow:
@dscho Thank so much!

PR #156, opened 2017-09-09
Enable Ctrl-Arrow backward forward words

iplaman:
Provide the same command-line experience as in Linux
Ctrl+Left moves cursor a word backward
Ctrl+Right moves cursor a word forward


dscho:
Excellent, thank you so much!

PR #159, opened 2017-10-17
Integration tests

albfan:
Based on https://github.com/chriscool/sharness (testsuite alike git uses) this test suite check the integrity of extras added to git-for-windows

First test check the comparison as text from open document files

Now bad news: This test is failing

you can run it with

    $ cd integration
    $ ./t0001-odt2txt

on failed test sharness keep a trash directory with results. cd into it and use git diff. I can see a text diff of a.odt difference, but from sharness env the result is "binary files differ" and test fails. 

This is the typical problem where testsuite env is different from real env, but hey, that's the overhead that a testsuite brings to a project

All my tries to detect where is the env difference fail

Any decission here can be reworked like:

- use submodules
- use sharness as test suite
- name convention for test


albfan:
Don't know how to configure this to run CI too

dscho:
> from sharness env the result is "binary files differ" and test fails.

I *think* that means that it does not take the Git attributes into account. Git's test suite specifically disables the system-wide gitconfig and gitattributes to allow for reproducible testing, so that's what I think bites you here: the test simply does not use the Git attributes you think it uses.

But taking a step back about integration testing...

Ideally, we would start using a framework which allows us to catch the following scenarios:

- Some Git commands behave differently depending whether they are called in a Powershell/CMD/Git Bash window
- The pager is not called correctly, so that `git log` output just scrolls by
- When quitting `git log`, the pager is actually terminated, but not the `git log` command
- Using `WinMerge` as difftool does not let you write the files in question because `git diff` still maps the files into memory (thereby blocking any writes)
- calling `Git GUI` from the Start Menu does not open any window

These were all issues I had to fix in the past, and some of them are reflected in the [pre-flight checklist](https://github.com/git-for-windows/build-extra/blob/master/installer/checklist.txt) I run through after generating installers and before publishing them.

I do not say that we need to come up with a system that allows from the beginning all of those types of bugs to be caught, but it would be really nice if we picked a framework to work with that *allows* those types of bugs to be caught. The `sharness` thing may be a little too limited to non-interactive, Unix shell use to allow for that...

Sadly, I could not find any good ideas how to go about this in a quick web search...

albfan:
I tried to add .gitattributes but didn't work yet.

Defining requisites is the best way to know what are you looking for.

Maybe sharness can be used as typical testsuite

For your requirements I guess we need something like:

https://www.getautoma.com/

But open source, like

http://pywinauto.readthedocs.io/en/latest/getting_started.html

Can you describe in details a use case to translate into a integration test? By now I will use those descriptions and try to code something

dscho:
> I tried to add .gitattributes but didn't work yet.

Hmm. Have you tried with `git diff --ext-diff`?

> http://pywinauto.readthedocs.io/en/latest/getting_started.html

That is an excellent find! And it should work with the Python package included in Git for Windows' SDK. I will try to find some time to play with it.

albfan:
> I will try to find some time to play with it.
Great

About testing opendocument: After install with my own installler (I update sdk an know size is similar than the Git-2.14.2.3-64-bit.exe

When I run git-bash.exe I see this

```
mkdir: cannot create directory '/dev/shm': Read-only file system

Creating /dev/shm directory failed.
POSIX semaphores and POSIX shared memory will not work

mkdir: cannot create directory '/dev/mqueue': Read-only file system

Creating /dev/mqueue directory failed.
POSIX message queues will not work

ln: failed to create symbolic link '/dev/stdin': Read-only file system
ln: failed to create symbolic link '/dev/stdout': Read-only file system
ln: failed to create symbolic link '/dev/stderr': Read-only file system
ln: failed to create symbolic link '/dev/fd': Read-only file system
ln: failed to create symbolic link '/etc/mtab': Permission denied
'C:\WINDOWS\system32\drivers\etc\hosts' -> '/etc/hosts'
/usr/bin/cp: cannot create regular file '/etc/hosts': Permission denied
'C:\WINDOWS\system32\drivers\etc\protocol' -> '/etc/protocols'
/usr/bin/cp: cannot create regular file '/etc/protocols': Permission denied
'C:\WINDOWS\system32\drivers\etc\services' -> '/etc/services'
/usr/bin/cp: cannot create regular file '/etc/services': Permission denied
'C:\WINDOWS\system32\drivers\etc\networks' -> '/etc/networks'
/usr/bin/cp: cannot create regular file '/etc/networks': Permission denied
rm: cannot remove '/etc/post-install/01-devices.post': Permission denied
rm: cannot remove '/etc/post-install/03-mtab.post': Permission denied
rm: cannot remove '/etc/post-install/06-windows-files.post': Permission denied
rm: cannot remove '/etc/post-install/99-post-install-cleanup.post': Permission denied
```

and same error about openssl when I try to push. Anyway I can wait till you release with opendocument to continue with this

dscho:
> ```
> mkdir: cannot create directory '/dev/shm': Read-only file system
> 
> Creating /dev/shm directory failed.
> POSIX semaphores and POSIX shared memory will not work
> 
> mkdir: cannot create directory '/dev/mqueue': Read-only file system
> 
> Creating /dev/mqueue directory failed.
> POSIX message queues will not work
> 
> ln: failed to create symbolic link '/dev/stdin': Read-only file system
> ln: failed to create symbolic link '/dev/stdout': Read-only file system
> ln: failed to create symbolic link '/dev/stderr': Read-only file system
> ln: failed to create symbolic link '/dev/fd': Read-only file system
> ln: failed to create symbolic link '/etc/mtab': Permission denied
> 'C:\WINDOWS\system32\drivers\etc\hosts' -> '/etc/hosts'
> /usr/bin/cp: cannot create regular file '/etc/hosts': Permission denied
> 'C:\WINDOWS\system32\drivers\etc\protocol' -> '/etc/protocols'
> /usr/bin/cp: cannot create regular file '/etc/protocols': Permission denied
> 'C:\WINDOWS\system32\drivers\etc\services' -> '/etc/services'
> /usr/bin/cp: cannot create regular file '/etc/services': Permission denied
> 'C:\WINDOWS\system32\drivers\etc\networks' -> '/etc/networks'
> /usr/bin/cp: cannot create regular file '/etc/networks': Permission denied
> rm: cannot remove '/etc/post-install/01-devices.post': Permission denied
> rm: cannot remove '/etc/post-install/03-mtab.post': Permission denied
> rm: cannot remove '/etc/post-install/06-windows-files.post': Permission denied
> rm: cannot remove '/etc/post-install/99-post-install-cleanup.post': Permission denied
> ```

These should all have been addressed [as part of the installer run](https://github.com/git-for-windows/build-extra/blob/aaa8969bec4671f094fa1c2ef546d98376227829/installer/install.iss#L2100-L2110)... :-(

dscho:
BTW interestingly, I just stumbled over this article discussing options for Bash script unit tests: https://medium.com/wemake-services/testing-bash-applications-85512e7fe2de

albfan:
I just now partially bats https://github.com/sstephenson/bats. Any choose you make will be ok.

About testing this, I finally open https://github.com/git-for-windows/git/issues/1326 about my problems with building installer. I'm close this time (don't know what is that api-builtin.html faulty file)

albfan:
After all the noise I have a first working test for odt2txt (maybe this would be part of a test suite, more than part of integration test suite)

I tried to configure an appveyor CI, but still now working (any tips?)

https://ci.appveyor.com/project/albfan/build-extra

appveyor.yml
```diff
@@ -0,0 +1,14 @@
...
+    DEPLOY_PROVIDER: bintray
+    BINTRAY_ACCOUNT: alexpux
+    BINTRAY_REPOSITORY: msys2
+    PACMAN_REPOSITORY_NAME: ci.msys
+
+build_script:
+    - C:\msys64\usr\bin\pacman --noconfirm --sync --refresh --refresh --sysupgrade --sysupgrade
+
+test_script:
+    - git submodule init
```
drizzd: This needs to be `git submodule update --init` to actually checkout the submodule. Looks like this is why your appveyor test is failing.

dscho:
So... I thought about this long and hard. And I think it is okay to add shell test scripts, but let's not inherit that unfortunate decision of Git to use *Unix shell scripting* for testing.

So here is what I propose: let's use a much more modern test framework: [`jest`](https://facebook.github.io/jest/). We already have a few node.js scripts that are fundamental to Git for Windows' release flow, Javascript is common enough that pretty much every developer at least saw their share, it is similar enough to C not to deter Windows developers who do not want to learn Unix shell scripting, and it is object-oriented enough to allow for a lot safer development.

Also, could we please rename the directory to `integration-tests/`? That is much more descriptive than `integration/`.

FWIW I would be totally okay with doing this in [Typescript](https://www.typescriptlang.org/) to begin with, as it adds type safetey with minimal cost, and excellent integration into the Open Source, cross platform Visual Studio Code that I use to great pleasure in the [GitGitGadget](https://gitgitgadget.github.io/) project.

So as a quick jot, here is my take how your test would look like:

```typescript
import "jest";

test("compare .odt files as text", async () => {
    const workDir = "odt-test";
    await git(["init", workDir]);
    await git(["config", "diff.mnemonicprefix", "true"], workDir);
    [...]
 });
```

This would require a function `git()` that I would define in a separate library file:

```typescript
import * as child_process from "child_process";
import * as fs from "fs";
import * as util from "util";

const copyFile = util.promisify(fs.copyFile);

async function git(args: string[], workDir?: string, stdin?: string):
    Promise<string> {
    return new Promise<string>((resolve, reject) => {
        const child = child_process.spawn("git", args, {
            cwd: workDir,
            stdio: ["pipe", "pipe", "inherit"],
        });
        if (stdin) {
            child.stdin.write(stdin);
        }
        child.stdin.end();
        let stdout = "";
        child.stdout.on("data", (data) => {
            stdout += data;
        });
        child.on("error", (error) => {
            reject(error);
        });
        child.on("close", (code) => {
            if (code !== 0) {
                const command = "git" +
                    (args.length ? ` '${args.join("', '")}'` : "");
                reject(`${command} exited with ${code}`);
            } else {
                resolve(stdout);
            }
        });
    });
}
```

I imagine that more common functions would need to be implemented in that library file, e.g. to initialize a repository (maybe from a fast-exported dump).

Thoughts?

albfan:
That sounds great to me.  Will try the fix commented by @drizzd and then another approach with jest

dscho:
@albfan awsome! BTW feel free to hop onto https://gitter.im/git-for-windows/git to discuss things, maybe even share a Live Session if you use VS Code...

drizzd:
Node.js seems like a fine choice to me. We use it at work not only for web development but also for scripting, mostly due to its platform independence. I am not convinced of the benefits of typescript, but I have no concrete objections to using it.

On the other hand, for testing command line interfaces nothing is as convenient as writing shell scripts. As far as I know the shell testing is working well enough for Git developers on Linux. The performance and platform dependencies are huge issues for Git for Windows, of course.

For Git integration tests, I suspect that we will have to run a lot of CLI commands. If the tests end up looking like a sequence of CLI command wrappers, then using Node.js may only bring us overhead. But let's try and see what the tests will look like.

drizzd:
Another concern is if Git for Windows is the right place to start deviating from the way testing is done in git.git. Using Node.js *will* make it more difficult to re-use and adapt shell tests from git.git in Git for Windows tests. So in an ideal world, we would start using Node.js in git.git first. But maybe we just have to start doing it because we are hurting the most.

dscho:
> I am not convinced of the benefits of typescript

@drizzd I am, though, there were tons of mistakes I made when I first converted [a lengthy shell script](https://github.com/dscho/mail-patch-series/) to node.js, and they were almost all caught when I converted that script to Typescript (this script was then the basis for GitGitGadget).

> On the other hand, for testing command line interfaces nothing is as convenient as writing shell scripts. As far as I know the shell testing is working well enough for Git developers on Linux. The performance and platform dependencies are huge issues for Git for Windows, of course.

Not to mention all the problems we have due to the Bash being based on the MSYS2 runtime. It causes huge headaches here.

> For Git integration tests, I suspect that we will have to run a lot of CLI commands.

My hope is not. For example, I would love for tests to start with a single `fast-import`. That should accelerate things rather handsomely.

> Another concern is if Git for Windows is the right place to start deviating from the way testing is done in git.git.

Repeat after me: we are doing integration tests here. There is no "deviating from the way testing is done in git.git" here. There are simply no integration tests in git.git.

> So in an ideal world, we would start using Node.js in git.git first.

I will not hold my breath for git.git to accept "new-fangled" tech in favor of old-school stuff like Unix shell scripting.

> If the tests end up looking like a sequence of CLI command wrappers, then using Node.js may only bring us overhead.

More overhead than using a full-blown `fork()` emulation just to spawn command-line programs, as the MSYS2 Bash does, will not be possible. `node.js` is *guaranteed* to be faster. And that's even before the JIT. And before using proper Object-Oriented state rather than having to perform line-based operations by storing state in strings, as you have to do in Unix shell scripts.

In short: I feel much more comfortable with picking a node.js framework to start our integration tests than I would feel picking a Unix shell scripted framework.

PhilipOakley:
For those following the thread, the https://github.com/gitgitgadget/gitgitgadget readme.md link is more complete that the https://gitgitgadget.github.io/  :smile: (mentioned in [comment ](https://github.com/git-for-windows/build-extra/pull/159#issuecomment-402098544))

drizzd:
Could you give more details what you mean with integration tests and what is different about those tests compared to the test-lib.sh based ones in git.git? The example test in this PR looks very similar to a test that would normally be written using test-lib.sh. The only difference seems to be that the test run against an installed Git for Windows. On Linux, it is also possible to run the Git tests against an installed Git executable.

In general, I am worried that we may be taking on more issues in Git for Windows than we can handle. It would be great if we could get the support of the Git community as a whole to handle platform independence issues. I do believe that the Git community will understand the benefits of a testing framework. So if we are willing to put some effort in here, if they get a say from the start maybe they will even accept a new framework for testing in git.git.

Sorry if I am coming in here asking questions out of ignorance. So please don't mind me if you feel this is the right approach. I definitely think it's a cool idea. :-)

dscho:
> the https://github.com/gitgitgadget/gitgitgadget readme.md link is more complete that the https://gitgitgadget.github.io/

@PhilipOakley good point. Care to open a PR to give the latter a facelift?

dscho:
> Could you give more details what you mean with integration tests

The idea of integration tests is to test things in the end product. Unit tests try instead to test things as early in the development cycle as possible, and on a much smaller scope: typically they work on "modules" (in Git, of course, we do not have modules, but we have commands, so that's what the granularity is in Git's test suite).

In particular, Git's test suite is singularly unsuited to run in and end-user installation, as that installation will typically lack even something as fundamental as `make`, and Git's test suite also relies very much on the presence of files like `GIT-VERSION-FILE` that were generated as part of the build (and are definitely not present in the end-user installation).

So the role of unit tests is to verify that a certain function, or a certain set of functions (such as the functionality of a given Git command), do what they are expected to do.

The role of integration tests, in contrast, is to ascertain that the product works in the targeted environment. This would e.g. catch the problem if Git for Windows' installer forgot to bundle a required `.dll` file: the unit tests would never catch that (as the `.dll` file is there).

Further, the integration tests can be rather extensive in their requirements. For example, we could extend those integration tests to spin up a full TFS instance (if available), with a newly-generated self-signed certificate, to verify that Git for Windows handles this case correctly.

Also, GUI tests are something I plan on doing via integration tests. See e.g. my [checklist](https://github.com/git-for-windows/build-extra/blob/master/installer/checklist.txt) that I run through (manually) every time before I publish a new Git for Windows version. It is rather tedious, and error-prone, to do this manually.

So at least here, you will probably agree that we safely left the realm of Git's test suite. As I said, it's more for unit testing. What we are talking about here are integration tests. Pretty much very different beast.

dscho:
The longer I think of this, the more I get convinced that we should use a modern test framework in a language that allows more control than Bash.

My personal preference, based on my experience writing [GitGitGadget](https://gitgitgadget.github.io/) would be to use [Jest](https://jestjs.io/). This would make it possible to use all kinds of really excellent `npm` packages to control the entire flow, much more precisely than shell scripts will ever be able to.

PhilipOakley:
@dscho, While I'd agree that the git test framework in bash is awkward given the intricacies of bash and it's variants, any alternate test frameworks probably needs to be even simpler and more broadly based if it is to get folks involved. 

Too much automation only makes the problems worse in terms of the specialist human input (https://www.zdnet.com/article/beware-the-automation-paradox/). I don't know Jest, and am only just getting to grips with the Git framework, so apply usual caution.

PhilipOakley:
> > the https://github.com/gitgitgadget/gitgitgadget readme.md link is more complete that the https://gitgitgadget.github.io/
> 
> @PhilipOakley good point. Care to open a PR to give the latter a facelift?

So that (update the README.md for the gitgitgadget.github.io repository) will be a #leftoverdocs item. 
(backlog++ ;-)

dscho:
> While I'd agree that the git test framework in bash is awkward given the intricacies of bash and it's variants, any alternate test frameworks probably needs to be even simpler and more broadly based if it is to get folks involved.

For these integration tests, I actually do not really care as much about ease (even if it would be easier to use Jest, mind you, as you can use VS Code with code completion and intellisense for Jest, something that you can definitely not do with Bash scripts) as I care about features.

For example, one of [the tests I perform for every single release](https://github.com/git-for-windows/build-extra/blob/master/installer/checklist.txt) is to start a Git Bash from the start menu and verifying that the MinTTY window appears and that it looks correct (i.e. no weird font or color issues).

Try to do that in Bash.

At least with Javascript, there is a good chance that there exists an `npm` package to take screenshots.

*clicketyclick*

There you go: https://www.npmjs.com/package/desktop-screenshot

> Too much automation only makes the problems worse in terms of the specialist human input (https://www.zdnet.com/article/beware-the-automation-paradox/).

If you believe that Git for Windows could be run by one single person without _massive_ amounts of automation, I hate to tell you that you are sorely mistaken.

Just to give you a _very_ small example: before I started automating Git for Windows releases, it would take me 3-5 days of scrambling after each Git release, just to get everything done and the release out and whatever I have to do after a release. With the automation, the only hard part is the actual rebase of the patches (and with the `shears/*` automation, I am notified early of potential merge conflicts and test suite failures, and I essentially record the merge conflict resolutions while updating the `shears/*` branches when I have to do it manually). The tedious parts are taken over by automation so that I can usually release within a couple of hours after I concluded the rebase.

In short: I call out BS in the article you linked. It essentially suggests that employers should pay for time, not for results. Which I am sure some employees find totally okay.

albfan:
@PhilipOakley This was a PoC. to detect errors. Seing in reverved order:

- Adding CI (to run CI, what else is needed to say here)
- Add integration tests (Run some test. Related to odt as that was my contribution at that time) 
- Installer: Fix post-install (a bug find at that time, cherry picked on master, so I guess this will be removed on a rebase)
- Fixing gitattributes (a fix that makes that test to pass, already on master)

Anyway, seems that using this bash framework is not the desired solution, why you care about this commits or the signed off part?. I try to open this discussion, not to get that working example merged.

I suggest to close this PR and start a test suite in the desired test framework, which seems to be Jest. The parent commit for this PR will allow you to implement the same test I did, but any parent of a commit that fix anything could work too.

PhilipOakley:
hi @albfan 
> why you care about this commits or the signed off part?. I try to open this discussion, not to get that working example merged

I had failed to check the original dates and was just responding to the marking that the integration tests had failed. 

Even for a Proof of Concept (PoC) it's worth at least adding a short, concise explanation of what each commit is doing, and a sign off. The latter at least avoids the simple DCO check failure, and allows the PoC to show as 'passed' if it works. And if it works, the explanations help it be taken forward.

albfan:
As you can see this PR use some commits that fix something in installer and in gitattributtes that was already merged.

```
*   18c340d Merge pull request #160 from albfan/fix-post-install
|\  
* | 787c656 please.sh release: the installers are actually in $USERPROFILE
* | 6a90492 nuget: fix release notes generation
* | 70e958e nuget: ensure that only Windows paths are passed to nuget.exe
* | 6124eac please.sh upgrade --force git: force re-tagging
* | 971230f please.sh upgrade git: optionally overwrite tags and package files
* | cf41bfd archive/release.sh: try to work around `tar: file changed as we read it`
* | 3091010 please.sh: do not try to install git-test-artifacts
* | 7479493 Mention New Feature in release notes
* | fea2075 Mention Bug Fix in release notes
| | * 2a305b1 (origin/integration-tests) Adding CI
| | * 1e6f17b Add integration tests
| |/  
| * 368ad82 (origin/fix-post-install) installer: Fix post-install
| * 191c28a (origin/fix-odt2txt) Fixing gitattributes
|/  
* f511d41 (origin/master) please.sh upgrade git: also upload *-test-artifacts packages
```

So after rebase those commits are gone.

Not sure what else you expect as explanation on the commits, anyway I tried to add something, but I think it's out of place. The explanation of what they are should be solved previously (what CI to use, what test framework to use, what kind of test we need).

Again I don't think this is even near to be merged, nor it's the desired tecnology to use.

I signed the commits too.

PhilipOakley:
> If you believe that Git for Windows could be run by one single person without _massive_ amounts of automation, I hate to tell you that you are sorely mistaken.



> In short: I call out BS in the article you linked. It essentially suggests that employers should pay for time, not for results. Which I am sure some employees find totally okay.

I think you have picked up the wrong side of the coin here. 

The problem with too much automation is that it stops those who are supporting the system from actually helping the isolated designers, because the supporters are not told (in a learning and understanding way) about the details of the automation, potentially leading to unexpected consequences.

The Max737 had one or two software coders removing control of the aircraft from the pilots, who then were unable to fly the plane. Similar issues happen with Bank automation where faults cascade to very long outages. It is the broad 'removal' of operational knowledge, through automation, from the operators that creates the risk. It happens across many industries. 

Here it means that you often end up three steps or four steps ahead of the potential contributors, hence my comment to make sure that any proposed techniques are easier to understand and catch up on for the average Jane/Joe, even if some of it is just providing more entry points into the documentation. It is an awkward Catch-22, needing to make haste slowly, sort of problem.

albfan:
@PhilipOakley I just recommend to trust in the maintainer who actually does the work instead of adding some examples you hear on the news.

Everything can be do done better, and that takes time and effort. We don't need to hear that again (I'm maintainer too). Just good contributions to review, modify and merge.

dscho:
@albfan I agree that this was a fine PoC without Sign-off. It got us to think about the problem, and while I think that Jest would make for a fine framework, I am not yet convinced ;-)

As to the argument of "less automation is better": I wholeheartedly disagree. _Especially_ in a project like Git for Windows.

albfan:
So apart from the technology, one start use case to test would be great. I see you define one previously. If there're other use cases (the simpler the better), maybe they can show the path to implement this (how and with what).

PhilipOakley:
@albfan 
> examples you hear on the news

The US Max737 report is at https://www.documentcloud.org/documents/6467999-Final-JATR-Submittal-to-FAA-Oct-2019-FINAL.html with the _broad recommendations_ regarding automated systems being on page 1 of the cover letter.

A useful reference for the way systems tend to progress toward failure is given in "Cognitive Systems Engineering" by Rasmussen et al. (1994), (see fig 6.3 for the trajectory toward the failure boundary). Alternately, "Human Error" by Reason (1990) is a good read. Human Failure effects are not a widely studied area as it is not seen as 'progress', so ends up as a niche area (see also Hollnagel, Dekker, etc and Thimbleby in the CS arena). It is something I have been engaged with as part of my Systems Engineer activities for over 30 years.

The right automation can be beneficial when it eliminates a whole range of errors, especially as part of a community 'dog-fooding' approach but does need that level of engagement so that the real causes are _completely_ eliminated. Many automation attempts merely remove symptoms not causes.

Shall we close this?

albfan:
You keep open new discussion in an issue where the maintainer states what should be done. Whatever works or fails in other projects means nothing. @dscho says what he wants to see here.

> Shall we close this?

What else you need if maintainer says he wants this feature?

What is left here is implementations and discuss about them until we have something that works for the one in charge of this project.

dscho:
> The US Max737 report is at https://www.documentcloud.org/documents/6467999-Final-JATR-Submittal-to-FAA-Oct-2019-FINAL.html with the _broad recommendations_ regarding automated systems being on page 1 of the cover letter.

Wow, this is _quite_ the tangent, I must say.

> What is left here is implementations and discuss about them until we have something that works for the one in charge of this project.

Indeed. I would love to have those integration tests, it would definitely make my maintainer duties a lot less burdensome.

dscho:
I have not forgotten about this PR at all.

In the meantime, two things happened that I think will rekindle the fire over here:

- A [`node.js` framework called `nut.js`](https://github.com/nut-tree/nut.js) looks like it might soon provide the functionality for automated UI testing (I opened two PRs there to support Windows' Scaling factor correctly, which is needed to make this work).
- I created [a new GitHub Action to set up Git for Windows' SDK](https://github.com/git-for-windows/setup-git-for-windows-sdk/) in GitHub workflows. This could be used to set up the `build-installers` flavor, creating a fresh MinGit (or Portable Git) from it and then running those integration tests.

I also got a lot more experience with the `node.js` framework [`jest`](https://jestjs.io/), which would integrate _real_ nicely into the GitHub workflow paradigm (there is [`actions/setup-node`](https://github.com/marketplace/actions/setup-node-js-environment) to create a `node.js` environment really conveniently).

@albfan are you still interested in this? Assuming you are, how about converting this PR to use `jest` instead of `sharness`? For example, [`integration/t0001-odt2txt.sh`](https://github.com/git-for-windows/build-extra/pull/159/commits/a2a7d781090b8342ed53484f907f3791fbcf9186#diff-907d900f3a09f8b6488c21f3f317853cb5315e29b6abaf949b8607455c3ced21) could become something like this:

```js
test('Compare odt files as text', async () => {
  expect(await git('-c diff.mnemonicprefix=true diff --no-index res/a.odt res/b.odt'))
    .toEqual(await fs.promises.readFile('res/result')
})
```

where the `git()` function could be defined somewhat like this:

```js
const exec = util.promisify(require('child_process').exec)
const execFile = util.promisify(require('child_process').execFile)

const git = async (command) => {
  const {stdout, stderr} = typeof command === 'string'
    ? await exec(`git ${command}`, {cwd: __dirname})
    : await execFile('git', command, {cwd: __dirname})
  stderr === '' || console.log(`command ${command}'s stderr:\n${stderr}`)
  return stdout
```

albfan:
Looks nice but I'm not skilled to code on node or javascript.

I'm not using windows machines these days.

Hope you're able to add some implementation and ease the maintenance of this project

dscho:
@albfan sounds good! I'll close this PR.

PR #160, opened 2017-10-21
installer: Fix post-install

albfan:
Fix post install script due to error in  variable expansion

### DIFF
--- a/git-extra/git-extra.install
+++ b/git-extra/git-extra.install
@@ -44,9 +44,9 @@ GITCONFIG
 *.RTF	diff=astextplain
 *.ods	diff=astextplain
 *.ODS	diff=astextplain
-*odf	diff=astextplain
+*.odf	diff=astextplain
 *.ODF	diff=astextplain
-*odt	diff=astextplain
+*.odt	diff=astextplain
 *.ODT	diff=astextplain
 
 GITATTRIBUTES
--- a/git-extra/gitattributes
+++ b/git-extra/gitattributes
@@ -14,7 +14,7 @@
 *.RTF	diff=astextplain
 *.ods	diff=astextplain
 *.ODS	diff=astextplain
-*odf	diff=astextplain
+*.odf	diff=astextplain
 *.ODF	diff=astextplain
-*odt	diff=astextplain
+*.odt	diff=astextplain
 *.ODT	diff=astextplain
--- a/post-install.bat
+++ b/post-install.bat
@@ -11,13 +11,14 @@
 @REM <percent>~fI Expands <percent>I to a fully qualified path name.
 @FOR /F "delims=" %%D IN ("%~dp0") DO @CD %%~fD
 
+@FOR /F "tokens=4 delims=.[XP " %%i IN ('ver') DO @SET ver=%%i
+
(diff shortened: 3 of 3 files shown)


plastikat:
Oh no, @albfan, this [commit](https://github.com/git-for-windows/build-extra/pull/160/commits/99626050c9e20cd0c72878e10857d2c185f59d0d) would break the test again! In ```cmd``` you _must not_ enclose any part of a **numeric** comparison in quotes, doing so turns the comparison into **lexicographic** one! The correct solution in a situation like this should be a _guarantee_ for ```ver``` variable to get a value. Basically, only two options exist:
* assigning variable a default value (inappropriate in this particular case)
* fixing version parsing

Second option is the one to concentrate on.

plastikat:
As I see in [#1326](https://github.com/git-for-windows/git/issues/1326) the string your ```ver``` command prints is ```Microsoft Windows [Versión 10.0.14393]``` (it is Spanish Windows, right?). At first I thought the root cause could be those non-ASCII characters, but this tiny test does not confirm it:
```
@echo off
set version=Microsoft Windows [Versión 10.0.14393]
FOR /F "tokens=4 delims=.[XP " %%i IN ("%version%") DO @SET ver=%%i
echo %ver%
```
It prints ```10```, as expected.

Don't forget that the script file encoding should be ```cp850``` for this code to work. Although this requirement _does not_ apply to the subject installer scripts (until we don't put non-ASCII characters there, and we currently do not).
Also note that I've modified the parsing line a bit: I use ```IN ("%version%")``` instead of ```IN ('ver')``` (note different quotes). This was on purpose, since in this test I parse a _string_, not a _command output_.

So, to summarize. I don't have a Spanish Windows at hand, so can't test it with the ```ver``` command directly, but it seems the parser line in question **does** its job with the _output_ of your ```ver``` command. The problem you encounter must be something else...

dscho:
Good catch. I just found the same error to happen with the tentative v2.14.3 installer... I'll merge your fix and restart the entire release engineering.

albfan:
Glad to help

dscho:
> Glad to help

You have no idea how much you helped me. So let me describe the situation in a bit of a detail: on Monday, I was struggling to get the entire release engineering process automated better (previously, I had already moved all those steps into the [`please.sh`](https://github.com/git-for-windows/build-extra/blob/master/please.sh) script, but I still performed them in a dedicated VM which was busy updating to Fall Creators Update, so I tried to perform those steps in Visual Studio Team Services build agents, but that required a lot of fiddling, as I had custom settings/credentials installed in that dedicated VM). Around midnight, I finally got the first round of installers/portable Gits/MinGits/etc, and my pre-publication test showed the very problem fixed in this Pull Request.

So after about a dozen hours of fighting the release engineering, I encountered a bug that you had already fixed, and all I had to do was to merge it (and then re-create the installers/portable Gits/MinGits/etc which was now easy, but then I continued to struggle with moving the publication step into VSTS build agents for almost four more hours).

I am really, really, really grateful for your help. It helped me stay (relatively) sane.

albfan:
I would say that have git-for-windows help me to stay sane developing on windows, So this is just contribute back

PR #166, opened 2018-01-11
Change to git-update behaviour when Git is up-to-date

david-russo:
As discussed in [Issue #1363](https://github.com/git-for-windows/git/issues/1363#issuecomment-356616531): the existing `git-update` behaviour, where no response is given if Git has already been validated as up-to-date in a previous check, has been moved behind a "--quiet" flag and added to the auto-update process.


dscho:
Thank you!

PR #168, opened 2018-01-19
installer: Change app url used by installer

nalla:
Since _Git for Windows_ propagated the new url
`https://gitforwindows.org/` with the last release we sould also use that
new url inside the installer.


dscho:
Thank you!

PR #169, opened 2018-01-26
git-extra: Install nanorc and git.nanorc file

nalla:
Quite a while now _Git for Windows_ is given the users a choice to
select the `nano` editor over `vim` as the default one. So it should
feel natural to us to improve the user experience with this new default
editor.

This commit will add the `git.nanorc` file that is available on github
(https://github.com/scopatz/nanorc/blob/master/git.nanorc) to the
git-extra package. git-extra will install that syntax-highlighting file
to `/usr/share/nano` so that it will be globally available.

This leads us to the second improvement. `nano` comes with quite a lot
syntax-highlighting files already. But they are not enabled by default.
So this commit also adds a slightly modifed `nanorc` file to the
git-extra package. The file will be installed into the `/etc` folder
just like the `vimrc` file. The following list shows all settings
that will be uncommented in this file. The default `nanorc` has no
settings enabled.

    set titlecolor brightwhite,blue
    set statuscolor brightwhite,green
    set selectedcolor brightwhite,magenta
    set numbercolor cyan
    set keycolor cyan
    set functioncolor green
    include "/usr/share/nano/*.nanorc"

This first group of settings will enable the default nano ui colors.
The last one will enable all those pretty syntax-highlighting files (
including the git based one).

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -25,6 +25,8 @@ pkgver() {
 source=('inputrc'
         'vimrc'
         'vi'
+        'nanorc'
+        'git.nanorc'
         'notepad'
         'wordpad'
         'create-shortcut.c'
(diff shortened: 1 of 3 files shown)


dscho:
Thanks!

PR #172, opened 2018-02-13
Don't set http.sslCAInfo for schannel

asalwa:
This addresses #1409

One strange thing in my fix may be the setting and unsetting of http.dummy variable. This is a work-around for Git config problem described here:
https://stackoverflow.com/questions/37147475/gitconfig-section-duplicates-everytime-i-set-a-preference
Without this trick the installer was leaving one empty [http] section. Everything else was working fine, but system config file looked ugly.

This is my first contribution to Git for Windows, please be forgiving :)

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1853,6 +1853,20 @@ begin
             Log('Line {#__LINE__}: Creating initial "' + ProgramData + '\Git\config" failed.');
         end;
     end;
+
+    {
+        Configure http.sslBackend according to the user's choice.
+    }
+
+    if RdbCurlVariant[GC_WinSSL].Checked then begin
+        Cmd:='schannel';
+    end else begin
+        Cmd:='openssl';
+    end;
+    if not Exec(AppDir+'\{#MINGW_BITNESS}\bin\git.exe','config --system http.sslBackend '+Cmd,
+                AppDir,SW_HIDE,ewWaitUntilTerminated,i) then
+        LogError('Unable to configure the HTTPS backend: '+Cmd);
+
     if FileExists(ProgramData+'\Git\config') then begin
         if not Exec(AppDir+'\bin\bash.exe','-c "value=\"$(git config -f config pack.packsizelimit)\" && if test 2g = \"$value\"; then git config -f config --unset pack.packsizelimit; fi"',ProgramData+'\Git',SW_HIDE,ewWaitUntilTerminated,i) then
             LogError('Unable to remove packsize limit from ProgramData config');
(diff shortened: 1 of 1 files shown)


dscho:
> This is my first contribution to Git for Windows, please be forgiving :)

I always try ;-)

> One strange thing in my fix may be the setting and unsetting of http.dummy variable. This is a work-around for Git config problem described here:
> https://stackoverflow.com/questions/37147475/gitconfig-section-duplicates-everytime-i-set-a-preference
> Without this trick the installer was leaving one empty [http] section. Everything else was working fine, but system config file looked ugly.

If you don't mind, I would prefer a different approach:

1. Change the order of the sslCAInfo and the sslBackend blocks in the first commit, explaining that this is a simple reordering so that the sslCAInfo can be unset in the next patch without introducing an ugly empty `[http]` section.

2. In the second commit, guard the sslCAInfo call as you did, behind `if not RdbCurlVariant[GC_WinSSL].Checked then begin [...] end;`

What do you think?

dscho:
(Oh, and if you agree, and do it that way, simply force-push to the same branch; This will update the PR automatically.)

asalwa:
Yes, I'm willing to follow your suggestion. Then nasty details come to the surface: I'd like to reorder sslCAInfo and sslBackend, but sslCAInfo is inside "if FileExists(ProgramData+'\Git\config')".
Can I just move sslBackend handling inside that "if"? Or should I first do some cleanup - maybe only commands touching ProgramData+'\Git\config' should be inside that "if"? And setting sslCAInfo in system config file should be outside?
Alternatively, we may also remove this check for FileExists(ProgramData+'\Git\config') at all. This file was just created a few lines above. And if creating the file failed, then it should be reported as error (LogError instead of Log if FileCopy(ExpandConstant('{tmp}\programdata-config.template'), ProgramData + '\Git\config', True) has failed.

dscho:
> Then nasty details come to the surface: I'd like to reorder sslCAInfo and sslBackend, but sslCAInfo is inside if FileExists(ProgramData+'\Git\config').

Personally, I would prefer the block

```pascal
    {
        Configure http.sslBackend according to the user's choice.
    }

    if RdbCurlVariant[GC_WinSSL].Checked then begin
        Cmd:='schannel';
    end else begin
        Cmd:='openssl';
    end;
    if not Exec(AppDir+'\{#MINGW_BITNESS}\bin\git.exe','config --system http.sslBackend '+Cmd,
                AppDir,SW_HIDE,ewWaitUntilTerminated,i) then
        LogError('Unable to configure the HTTPS backend: '+Cmd);
    if HttpDummySet then begin
        if not Exec(AppDir+'\{#MINGW_BITNESS}\bin\git.exe','config --system --unset http.dummy',
                    AppDir,SW_HIDE,ewWaitUntilTerminated,i) then
            LogError('Unable to unset http.dummy');
    end;
```

to be moved just before the line `if FileExists(ProgramData+'\Git\config') then begin` (and add a comment before that).

What do you think?

asalwa:
I've updated and pushed my branch according to your suggestions.

dscho:
Very nice! Thank you!

PR #173, opened 2018-02-21
installer: add the git-lfs wrapper to /cmd

stffabi:
Support for git-lfs wrapping in `/cmd`, see git-for-windows/git#1503

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -208,6 +208,9 @@ Type: files; Name: {app}\{#MINGW_BITNESS}\bin\git-*.exe
 Type: files; Name: {app}\{#MINGW_BITNESS}\libexec\git-core\git-*.exe
 Type: files; Name: {app}\{#MINGW_BITNESS}\libexec\git-core\git.exe
 
+; Delete git-lfs wrapper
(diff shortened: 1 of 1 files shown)


installer/install.iss
```diff
@@ -208,6 +208,9 @@ Type: files; Name: {app}\{#MINGW_BITNESS}\bin\git-*.exe
 Type: files; Name: {app}\{#MINGW_BITNESS}\libexec\git-core\git-*.exe
 Type: files; Name: {app}\{#MINGW_BITNESS}\libexec\git-core\git.exe
 
+; Delete git-lfs wrapper
+Type: files; Name: {app}\cmd\git-lfs.exe
```
stffabi: Does this work also if the file was not created, because `IsComponentSelected('gitlfs')` was false during the installation?

stffabi: Doesn't seem to be a problem.

dscho: Yes, sorry, those files are deleted gracefully (i.e. if they are already absent, it is not a big deal, I verified that when I added the part where copied .dll files are deleted from `libexec\git-core`).

dscho:
Excellent! Thank you!

PR #174, opened 2018-02-21
installer: always use the Git wrapper for cmd\git-lfs.exe

dscho:

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1576,21 +1576,26 @@ end;
 
 // Procedure to create hardlinks for builtins. This procedure relies upon that
 // git-wrapper.exe is already copied to {app}\tmp.
-procedure CopyBuiltin(FileName:String);
+procedure HardlinkOrCopyGit(FileName:String;Builtin:Boolean);
 var
-    AppDir:String;
+    AppDir,GitTarget:String;
     LinkCreated:Boolean;
 begin
-    if (not DeleteFile(FileName)) then begin
+    if FileExists(FileName) and (not DeleteFile(FileName)) then begin
         Log('Line {#__LINE__}: Unable to delete existing built-in "'+FileName+'", skipping.');
         Exit;
     end;
 
     AppDir:=ExpandConstant('{app}');
+    if Builtin then
+        GitTarget:=AppDir+'\{#MINGW_BITNESS}\bin\git.exe'
+    else
+        // For non-builtins, we want to use the Git wrapper in cmd
+        GitTarget:=AppDir+'\cmd\git.exe';
 
     try
         // This will throw an exception on pre-Win2k systems.
-        LinkCreated:=CreateHardLink(FileName,AppDir+'\{#MINGW_BITNESS}\bin\git.exe',0);
+        LinkCreated:=CreateHardLink(FileName,GitTarget,0);
     except
         LinkCreated:=False;
         Log('Line {#__LINE__}: Creating hardlink "'+FileName+'" failed, will try a copy.');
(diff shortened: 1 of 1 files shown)


installer/install.iss
```diff
@@ -1576,9 +1576,9 @@ end;
 
 // Procedure to create hardlinks for builtins. This procedure relies upon that
 // git-wrapper.exe is already copied to {app}\tmp.
-procedure CopyBuiltin(FileName:String);
+procedure HardlinkOrCopyGit(FileName:String,Builtin:Boolean);
```
stffabi: Shouldn't this be a `;` to separate arguments?

installer/install.iss
```diff
@@ -1587,10 +1587,14 @@ begin
     end;
 
     AppDir:=ExpandConstant('{app}');
+    if Builtin then
```
stffabi: I've received a compile error here, adding a `begin` to the end of the line fixed it.

installer/install.iss
```diff
@@ -1587,10 +1587,14 @@ begin
...
+        GitTarget:=AppDir+'\{#MINGW_BITNESS}\bin\git.exe'
+    else
```
stffabi: I had to change this to `end else begin` for building to succeed.

installer/install.iss
```diff
@@ -1587,10 +1587,14 @@ begin
...
+        GitTarget:=AppDir+'\bin\git.exe'
```
stffabi: Another `end;` was needed here.

installer/install.iss
stffabi: Don't we need to switch the then and else parts? Because if `Builtin` is set to true, we need to target the wrapper which is found under `AppDir+'\bin\git.exe'`

installer/install.iss
```diff
@@ -1576,9 +1576,9 @@ end;
...
 // Procedure to create hardlinks for builtins. This procedure relies upon that
 // git-wrapper.exe is already copied to {app}\tmp.
-procedure CopyBuiltin(FileName:String);
+procedure HardlinkOrCopyGit(FileName:String,Builtin:Boolean);
 var
-    AppDir:String;
+    AppDir,GitTarget:String;
     LinkCreated:Boolean;
 begin
     if (not DeleteFile(FileName)) then begin
```
stffabi: Wouldn't it be more robust to execute `DeleteFile` only if the file exists? This would also fix the `git-lfs` case, where `DeleteFile` fails because there's no existing file under `/cmd/` and therefore no hardlink would be created.

installer/install.iss
dscho: Uh oh, you're correct! I *always* forget...

installer/install.iss
dscho: No, the builtins are the Git builtins in `mingw64\libexec\git-core`. They need to be hard-linked to the real `mingw64\bin\git.exe`, really.

Git LFS, in contrast, is not a builtin, and needs to be hard-linked to the Git wrapper. I should make it more obvious that the `bin\git.exe` is the Git wrapper, really... and I guess I make that more obvious by using `cmd\git.exe` instead...

installer/install.iss
dscho: Oh! Absolutely!

installer/install.iss
stffabi: Uff I'm so confused about what needs to be hard-linked to the wrapper and what to real git...

This would then mean that in the Git LFS case we have to pass `False` to HardlinkOrCopyGit on Line [1845](https://github.com/git-for-windows/build-extra/pull/174/commits/40143f2ad4d311bc6caccba6ecffd623831cb197#diff-9db0b81cc5d2dc9d49e4e8634ac624b6R1845) whereas we currently pass `True`

dscho: Oy. You know what? I should totally learn how to read my own patches. Sorry about that.

stffabi:
I've retested your latest changes. LGTM and everything works as expected. Thanks so much.

dscho:
Thank you, @stffabi! I managed to test this myself (using the opportunity to add some VSTS build job to test things like this a lot more easily than before) and I confirm that it is working, too.

PR #177, opened 2018-03-15
Install xz

orgads:
It is widely used nowadays

Signed-off-by: Orgad Shaneh <orgads@gmail.com>


dscho:
Thank you, this is a straight-forward and useful change.

Two questions:

1. Does this mean that `tar xf <file>.tar.xz` works when this is merged? If so, let's state that in the commit message.
2. By how much does this increase the installer size? I spent a ton of time to hammer it down again, and it would be a shame to make it heavier again to include a tool that Git itself does not need or use. If the installer size delta is small enough, let's merge this, but please do include the numbers in the commit message.

orgads:
I placed it in the wrong list :)

Now did it the right way, tested and updated the commit message.

dscho:
Excellent! Thank you!

PR #179, opened 2018-04-12
diffutils dep for git-extra

AeliusSaionji:
There's a call to `cmp ` in the .install script.

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -9,7 +9,7 @@ arch=('i686' 'x86_64')
 url="https://github.com/git-for-windows/build-extra"
 license=('GPL')
 groups=('VCS')
-depends=('git-for-windows-keyring')
+depends=('git-for-windows-keyring' 'diffutils')
 optdepends=('vim' 'filesystem')
 install='git-extra.install'
 pkgver() {


PhilipOakley:
The commit message needs a little work to clarify 'why' this is the right solution to the (unstated) problem.
Also, no abbrevs. ;-)

And wwill need a sign off. The request can be uspdated by a forced push.

dscho:
@AeliusSaionji have you seen @PhilipOakley's excellent suggestions?

AeliusSaionji:
I have made no abbreviations, the problem is as stated: there's literally a call to the command `cmp` in the associated install script. It therefore depends on diffutils.

Sorry for the delay, I have been busy. I will look into signing off the PR soon.

dscho:
If there is no more activity on this Pull Request, I will close it soon.

drizzd:
Could we use `diff -q` instead to avoid the extra dependency?

drizzd:
Never mind, both diff and cmp are in diffutils, and diffutils has been removed from the base group:
https://github.com/Alexpux/MSYS2-packages/commit/1994b55167e25d1396a6b8aef33fcccfcfddbe17#diff-a27aecab7197c6b19362658016bf0313R10

Since it is a runtime dependency, we need to add it here.

I think Philip was referring to your abbrevation of *dependency* in your commit message.

dscho:
Thanks, @drizzd for reviving this. I simply reworded the commit as requested, and while I am not enthusiastic of having a "sign off" from a bogus identity, the change itself seems to be too trivial to fight over. Let's just move on.

PR #180, opened 2018-04-19
Added Visual Studio Code Insiders Support

justalemon:
This fixes git-for-windows/git#1513.

The result is quite good:
imagen

I installed the build with this settings:
imagen

And it was completed without problems:
imagen

As you can see in the first picture, I added a message about the differences between Code Stable (blue) and Insiders (green).


dscho:
Thank you so much for this contribution! It is exciting to see support for the Insiders version of VS Code.

I would like to ask for three changes, though:

- In general, I like to have every revision in a topic branch consistent. In particular, no commit should introduce a bug that is resolved in a subsequent commit *in the same topic branch*. For example, https://github.com/git-for-windows/build-extra/pull/180/commits/4c0faba520e3b8411b5220f68b480989cdbb886f should be squashed into https://github.com/git-for-windows/build-extra/pull/180/commits/a19aa9ee98b452dc19666b4507a02eba0c00c796, should be squashed into https://github.com/git-for-windows/build-extra/pull/180/commits/9987f213cdd61c6cd04086bcfa57ec4608273a8b, and other commits really do not make sense on their own, such as replaying previous choices not from the start, but only from the middle of the topic branch. In fact, I could imagine that all these commits should be squashed into one, consistent commit that adds support for Visual Studio Code Insiders.

-  The commit messages should be self-contained. Ideally, the reader who stumbles across one commit should have all the information necessary to understand the "Why?" of the diff from the commit message. See e.g. 51961c66715443570827792b2f450da55811493a which describes the purpose very well, and concisely. Something along these lines:

  ```
  installer: offer to use VS Code - Insiders as Git's default editor

  VS Code comes in the "Insiders" flavor, i.e. updated every night from the then-current `master`.
  For users who like this bleeding edge, let's offer the Insiders flavor alongside the regular VS
  Code version, separately autodetecting the presence of either.
  ```
- could you please add your [Sign-off](https://github.com/probot/dco#how-it-works) to the commit message(s)? This will please the DCO bot and avoid marking the checks as failed.

justalemon:
@dscho I'm going to send a new quick PR.

PR #182, opened 2018-04-28
git-extra: added support for dotx diffs

belidzs:
dotx is the file format for Word templates which is already supported by
docx2txt.pl

This pull request was based on pull request #128 and issue #900
(git-for-windows/git). They added compatibility with docm and dotm files


dscho:
Perfect, thank you so much!

PR #186, opened 2018-05-08
Some changes to the vimrc

chrisbra:
Hi this branch changes some of the Vim settings. I tried to make the commits as granular as possible to make reviewing easier.

Currently some settings can be improved a bit, so I slightly changed them. I tried not to change them according to my personal taste and if I did, I mentioned it in the commit message (cursor block style, command line completion, ...) and those could then reverted if needed.

The rest should keep Vim more closely to Vims upstream repository, especially since it sources the defaults.vim file anyhow (currently overwriting some settings). 

Also I found comments beneficial, so I left a couple of more comments. One thing I did not know for sure is the usage of EDIT_MERGEMSG and EDIT_MSG. I am not sure whether both of them are needed or only one of them (and the other one was a typo). For now I left it as is (and added the other one to the settings as well. This can easily be reverted if needed. Also if this is correct, it might be beneficial to add some of this to upstreams filetype detection logic.

### DIFF
--- a/git-extra/vimrc
+++ b/git-extra/vimrc
@@ -20,17 +20,15 @@ set statusline=%<%F%h%m%r\ [%{&ff}]\ (%{strftime(\"%H:%M\ %d/%m/%Y\",getftime(ex
 "------------------------------------------------------------------------------
 " Only do this part when compiled with support for autocommands.
 if has("autocmd")
-    "Set UTF-8 as the default encoding for commit messages
+    " Set UTF-8 as the default encoding for commit messages
     autocmd BufReadPre COMMIT_EDITMSG,MERGE_MSG,git-rebase-todo setlocal fileencodings=utf-8
 
-    "Remember the positions in files with some git-specific exceptions"
+    " Remember the positions in files with some git-specific exceptions"
     autocmd BufReadPost *
       \ if line("'\"") > 0 && line("'\"") <= line("$")
-      \           && expand("%") !~ "COMMIT_EDITMSG"
-      \           && expand("%") !~ "MERGE_EDITMSG"
+      \           && &filetype !~# 'commit\|gitrebase'
       \           && expand("%") !~ "ADD_EDIT.patch"
-      \           && expand("%") !~ "addp-hunk-edit.diff"
-      \           && expand("%") !~ "git-rebase-todo" |
+      \           && expand("%") !~ "addp-hunk-edit.diff" |
       \   exe "normal g`\"" |
       \ endif
 
--- a/git-extra/vimrc
+++ b/git-extra/vimrc
@@ -4,7 +4,6 @@
 set ai                          " set auto-indenting on for programming
 set showmatch                   " automatically show matching brackets. works like it does in bbedit.
 set vb                          " turn on the "visual bell" - which is much quieter than the "audio blink"
-set ruler                       " show the cursor position all the time
 set laststatus=2                " make the last line where the status is two lines deep so you can see status always
 set backspace=indent,eol,start  " make that backspace key work the way it should
 set nocompatible                " vi compatible is LAME
--- a/git-extra/vimrc
+++ b/git-extra/vimrc
@@ -12,6 +12,7 @@ set vb                          " turn on the "visual bell" - which is much quie
 set laststatus=2                " make the last line where the status is two lines deep so you can see status always
 set showmode                    " show the current mode
 set clipboard=unnamed           " set clipboard to unnamed to access the system clipboard under windows
+set wildmode=list:longest,longest:full   " Better command line completion
 
 " Show EOL type and last modified timestamp, right after the filename
 set statusline=%<%F%h%m%r\ [%{&ff}]\ (%{strftime(\"%H:%M\ %d/%m/%Y\",getftime(expand(\"%:p\")))})%=%l,%c%V\ %P
--- a/git-extra/vimrc
+++ b/git-extra/vimrc
@@ -26,6 +26,12 @@ set statusline+=%=              " Rest: right align
 set statusline+=%l,%c%V         " Position in buffer: linenumber, column, virtual column
 set statusline+=\ %P            " Position in buffer: Percentage
 
+if &term =~ 'xterm'             " mintty identifies itsself as xterm-compatible
+  if &t_Co == 8
+    set t_Co = 256              " Use at lease 256 colors
+  endif
+  " set termguicolors           " Uncomment, to allow trucolors on mintty
+endif
 "------------------------------------------------------------------------------
 " Only do this part when compiled with support for autocommands.
 if has("autocmd")
--- a/git-extra/vimrc
+++ b/git-extra/vimrc
@@ -33,7 +33,6 @@ if has("autocmd")
       \ endif
 
       autocmd BufNewFile,BufRead *.patch set filetype=diff
-      autocmd BufNewFile,BufRead *.diff set filetype=diff
 
       autocmd Syntax diff
       \ highlight WhiteSpaceEOL ctermbg=red |
--- a/git-extra/vimrc
+++ b/git-extra/vimrc
@@ -1,16 +1,17 @@
 " Setting some decent VIM settings for programming
 " This source file comes from git-for-windows build-extra repository (git-extra/vimrc)
 
+ru! defaults.vim                " Use Enhanced Vim defaults
+set mouse=                      " Reset the mouse setting from defaults
+aug vimStartup | au! | aug END  " Revert last positioned jump, as it is defined below
+let g:skip_defaults_vim = 1     " Do not source defaults.vim again (after loading this system vimrc)
+
 set ai                          " set auto-indenting on for programming
 set showmatch                   " automatically show matching brackets. works like it does in bbedit.
 set vb                          " turn on the "visual bell" - which is much quieter than the "audio blink"
 set laststatus=2                " make the last line where the status is two lines deep so you can see status always
-set backspace=indent,eol,start  " make that backspace key work the way it should
-set nocompatible                " vi compatible is LAME
-set background=dark             " Use colours that work well on a dark background (Console is usually black)
 set showmode                    " show the current mode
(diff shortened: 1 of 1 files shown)


git-extra/vimrc
```diff
@@ -19,17 +19,16 @@ set statusline=%<%F%h%m%r\ [%{&ff}]\ (%{strftime(\"%H:%M\ %d/%m/%Y\",getftime(ex
...
+    " Remember the positions in files with some git-specific exceptions"
     autocmd BufReadPost *
       \ if line("'\"") > 0 && line("'\"") <= line("$")
-      \           && expand("%") !~ "COMMIT_EDITMSG"
+      \           && &ft !~# 'commit\|gitrebase'
       \           && expand("%") !~ "MERGE_EDITMSG"
       \           && expand("%") !~ "ADD_EDIT.patch"
-      \           && expand("%") !~ "addp-hunk-edit.diff"
-      \           && expand("%") !~ "git-rebase-todo" |
+      \           && expand("%") !~ "addp-hunk-edit.diff" |
```
dscho: For somebody versed in vim's command language, this might all look very clear. To everybody else, this is just magic.

That includes me, and I need to understand those changes before I can merge the PR. Please take a little bit of time to explain in the commit message what you did there, at least enough for somebody e.g. stuck in an airplane without internet access (and therefore unable to look up documentation and things) to  be able to change the `&ft` line to their needs.

git-extra/vimrc
```diff
@@ -20,13 +20,13 @@ set statusline=%<%F%h%m%r\ [%{&ff}]\ (%{strftime(\"%H:%M\ %d/%m/%Y\",getftime(ex
 " Only do this part when compiled with support for autocommands.
 if has("autocmd")
     " Set UTF-8 as the default encoding for commit messages
-    autocmd BufReadPre COMMIT_EDITMSG,MERGE_MSG,git-rebase-todo setlocal fileencodings=utf-8
+    autocmd BufReadPre COMMIT_EDITMSG,MERGE_EDITMSG,MERGE_MSG,git-rebase-todo setlocal fileencodings=utf-8
```
dscho: This comes from https://github.com/git-for-windows/build-extra/commit/f5cc31f85522b92d0ee1e7ea11de2234cd7cc37d, which references https://github.com/git-for-windows/git/issues/1207 as reason for introducing the `MERGE_EDITMSG` file name here.

As you can see easily from reading that discussion, I simply made a mistake, and `MERGE_MSG` was really intended.

Maybe take this opportunity to fix it?

git-extra/vimrc
```diff
@@ -37,6 +37,4 @@ if has("autocmd")
       autocmd Filetype diff
       \ highlight WhiteSpaceEOL ctermbg=red |
       \ match WhiteSpaceEOL /\(^+.*\)\@<=\s\+$/
-
-      autocmd Syntax gitcommit setlocal textwidth=74
```
dscho: Yep, this stems from a time when this was *not* configured automatically.

git-extra/vimrc
```diff
@@ -14,7 +14,16 @@ set clipboard=unnamed           " set clipboard to unnamed to access the system
...
+" Set the statusline
+set statusline=%f               " filename relative to current $PWD
+set statusline+=%h              " help file flag
+set statusline+=%m              " modified flag
+set statusline+=%r              " readonly flag
+set statusline+=\ [%{&ff}]      " Fileformat [unix]/[dos] etc...
+set statusline+=\ (%{strftime(\"%H:%M\ %d/%m/%Y\",getftime(expand(\"%:p\")))})  " last modified timestamp
+set statusline+=%=              " Rest: right align
+set statusline+=%l,%c%V         " Position in buffer: linenumber, column, virtual column
+set statusline+=\ %P            " Position in buffer: Percentage
```
dscho: Nicely done!

Since this is *expanding* rather than *simplifying*, let's call it that in the commit message?

Also, I see that `%<` was removed. Can you describe in the commit message, please, what this does and why it is okay to remove it?

git-extra/vimrc
```diff
@@ -25,6 +25,12 @@ set statusline+=%=              " Rest: right align
 set statusline+=%l,%c%V         " Position in buffer: linenumber, column, virtual column
 set statusline+=\ %P            " Position in buffer: Percentage
 
+if &term =~ 'xterm'
+  let &t_ti.="\e[1 q"           " Use a blinking block cursor when entering termcap mode
```
dscho: Does this let the cursor blink by default?

I wonder how that interacts with MinTTY. I configured a block cursor there, and I could imagine that other users did so, too. Yet other users may have configured a line cursor specifically and could possibly be pretty annoyed by this here patch.

What's your take on this?

git-extra/vimrc
```diff
@@ -30,6 +30,12 @@ if &term =~ 'xterm'
   let &t_SI.="\e[5 q"           " A thin blinking bar for insert/replace mode
   let &t_EI.="\e[1 q"           " block cursor after leaving insert/replace mode
   let &t_te.="\e[0 q"           " Blinking block cursor when leaving termcap mode
+  if &t_Co == 8
+    set t_Co = 256              " Use at lease 256 colors
+  endif
+  if 0
+    set termguicolors           " Allow trucolors on mintty 
+  endif
```
dscho: What is the effect of this? The commit message stops short of explaining that.

Also, the commit message starts with an unnecessarily controversial "better". Let's avoid that. Let's just call it "Use more colors".

dscho:
I reviewed all patches, and left a few comments and suggestions.

git-extra/vimrc
```diff
@@ -1,16 +1,16 @@
 " Setting some decent VIM settings for programming
 " This source file comes from git-for-windows build-extra repository (git-extra/vimrc)
 
+ru! defaults.vim                " Use Enhanced Vim defaults
+set mouse=                      " Reset the mouse setting from defaults
+aug vimStartup | au! | aug END  " Revert last positioned jump, as it is defined below
```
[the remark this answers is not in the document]
chrisbra: Of course, this is Vim :)

git-extra/vimrc
```diff
@@ -19,17 +19,16 @@ set statusline=%<%F%h%m%r\ [%{&ff}]\ (%{strftime(\"%H:%M\ %d/%m/%Y\",getftime(ex
 "------------------------------------------------------------------------------
 " Only do this part when compiled with support for autocommands.
 if has("autocmd")
-    "Set UTF-8 as the default encoding for commit messages
+    " Set UTF-8 as the default encoding for commit messages
     autocmd BufReadPre COMMIT_EDITMSG,MERGE_MSG,git-rebase-todo setlocal fileencodings=utf-8
 
-    "Remember the positions in files with some git-specific exceptions"
+    " Remember the positions in files with some git-specific exceptions"
```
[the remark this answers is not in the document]
chrisbra: correct

git-extra/vimrc
chrisbra: Sure, will do.

git-extra/vimrc
chrisbra: I am pretty sure, Vim has support for the diff filetype for more than 15 years however. But most likely, filetype detection was not enabled back then by default.

git-extra/vimrc
chrisbra: Sure, will do.

git-extra/vimrc
chrisbra: will do

chrisbra:
> Could you please add "Signed-off-by:" footers to the commit messages (that's why the DCO bot marks this PR with a red X)?

Yeah I saw that, once I created the PR. If I had been a bit more careful before creating the PR, I would have Signed-off all those commits. I usually do not follow the git practice in this regard.

But since I had to reword (and slightly change) some of the commits anyhow, I also added Signed-off-by header. 

Also I reworded the commits as required to explain a bit, why something was changed (and dropped the use blinking cursor commit, that was unneccessary and I should have configured my terminal directly instead of trying to fix this in Vim.)

git-extra/vimrc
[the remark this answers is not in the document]
dscho: In that case, could I ask to split them out? I was really confused because I am not familiar with the Vim syntax, so I thought you introduced functional changes here that I do not understand.

git-extra/vimrc
dscho: Thanks!

git-extra/vimrc
dscho: Ah! Thanks for the clarification. I misunderstood.

git-extra/vimrc
```diff
@@ -11,6 +11,7 @@ set vb                          " turn on the "visual bell" - which is much quie
 set laststatus=2                " make the last line where the status is two lines deep so you can see status always
 set showmode                    " show the current mode
 set clipboard=unnamed           " set clipboard to unnamed to access the system clipboard under windows
+set wildmode=list:longest,longest:full   " Better command line completion
```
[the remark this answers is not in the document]
dscho: Thank you, the new commit message does a really nice job of explaining it!

git-extra/vimrc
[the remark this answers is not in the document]
dscho: Ah, I see that you explained that part in the commit message. No need to split anything anymore!

dscho:
Thank you for your contribution, and for your patience with me!

chrisbra:
Thanks for merging. One minor comment

> I try to make a point in this project to be inclusive. Next time, please include the roughly 50% of humanity that happens to have a different sex than you do.

Hm I always thought he refers to `the user` which in my understanding would be neither masculine nor feminin. In German I don't think of 'männlich' wenn talking about "der Anwender". But it's not my monther tongue as you might guess, so I beg for pardon for the misunderstanding.

dscho:
> I always thought he refers to the user which in my understanding would be neither masculine nor feminin

No, in English "he" refers to a male being. Imagine how you would feel if you were addressed as a "she" 100% of the time? That's exactly why this should be avoided. Hence the "singular they". Example:

> Whenever a user wants to open a file using Git's default editor, they only need to call `git config -e -f <path>`.

PR #188, opened 2018-05-30
Fix wrong links and a duplicate entry in 2.17.1 release notes

orgads:
Fixes git-for-windows/git#1694

Signed-off-by: Orgad Shaneh <orgads@gmail.com>


dscho:
Thank you!

In this case, I guess the patch is descriptive enough that the commit message can be essentially empty. In general, I do prefer commit messages that stand on their own, rather than just linking elsewhere (links can go dead, web pages can be changed, and the internet connection may be failing *just* when a developer needs to understand what a commit message should have talked about).

As I said, not a big deal in this case.

PR #189, opened 2018-06-08
Fix astextplain for pdf files

CrOrc:
Fixed order of command line arguments for pdftotext.
With current order it prints help message to stderr.


dscho:
Thank you so much!

PR #190, opened 2018-06-13
Exclude PCRE2 documentation from releases

rimrul:
As Git for Windows did not ship PCRE documentation before switching to the PCRE2 dependency, there
should be no need too start shipping PCRE2 documentation now.


dscho:
Thanks!

PR #191, opened 2018-06-19
enable GCM and Chacha20 ciphers

drizzd:
These are offered by ssh by default, but they are not included in the
default list of Ciphers in ssh_config.

Closes https://github.com/git-for-windows/git/issues/1723.


drizzd:
We should instead consider the  alternatives suggested here https://github.com/openssh/openssh-portable/pull/94#issuecomment-398570669.

dscho:
> We should instead consider the alternatives suggested here openssh/openssh-portable#94 (comment).

But we can do that here, too, right? I mean, we can edit the very same code you edited in this first draft, and make it do what we need it to do instead, right?

drizzd:
Yes, here is what we could do instead: https://github.com/git-for-windows/build-extra/pull/192

PR #192, opened 2018-06-20
enable GCM and Chacha20 ciphers

drizzd:
These are offered by ssh by default, but they are not included in the
default list of Ciphers in ssh_config.

Instead of setting the entire list of ciphers, selectively re-enable the
CBC ciphers using the "Cipher +somecipher" notation.

Closes https://github.com/git-for-windows/git/issues/1723.


dscho:
Thank you!

I do have two requests, still, before merging time:

1. could you Sign-off on that patch?
2. existing Git for Windows SDKs will already have an `/etc/ssh_config` that was patched by the current `git-extra` (or an earlier one). Could you make it so that this case is handled, too, by reverting the previous edit (something like `! grep '^Cipher .*,aes256-cbc,aes192-cbc' /etc/ssh/ssh_config || sed -i -e 's|^\(Cipher .*\),aes256-cbc,aes192-cbc|# \1|' /etc/ssh/ssh_config`)?

drizzd:
Good point. I can do this, but if a user modifies /etc/ssh/ssh_config, then we may overwrite their modifications too. I think the right way is to patch the openssh package instead. Since we already have some patches there, that's no extra maintenance overhead. I created a pull request there:
https://github.com/git-for-windows/MSYS2-packages/pull/24

The only downside is that due to the modification by git-extra, the new ssh_config will be installed as /etc/ssh/ssh_config.new with a warning.

dscho:
> The only downside is that due to the modification by git-extra, the new ssh_config will be installed as /etc/ssh/ssh_config.new with a warning.

That's not the only downside. It would make it quite a bit harder to revert the supposedly temporary workaround of re-enabling weak ciphers to accommodate a few hosters (all of which have updated their SSH to handle strong ciphers in the meantime AFAICT).

git-extra/git-extra.install
```diff
@@ -219,9 +219,14 @@ test -d "$TMPDIR" || test ! -d "$TMP" || {\
 	# Re-enable at least AES256-cbc and AES128-cbc, as they are still used
 	# e.g. in VSTS
 	test ! -f /etc/ssh/ssh_config ||
-	grep -q '^Ciphers .*aes256-cbc,aes192-cbc' /etc/ssh/ssh_config ||
-	sed -i -e '/^[# ]*Ciphers /{s/^# *//;s/$/,aes256-cbc,aes192-cbc/}' \
-		/etc/ssh/ssh_config
+	{
+		# Revert change by prior versions of git-extra.
+		sed -i -e 's/^Ciphers aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes256-cbc,aes192-cbc$/#   Ciphers aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc/' \
+			/etc/ssh/ssh_config
```
dscho: I tried very hard not to touch files unless needed, whereas this unconditional `sed` call would always be called.

Does this alternative snippet do the job, too?

```sh
! grep -q '^Ciphers .*cbc' ||
sed -i -e 's/^Ciphers .*cbc/# &/' /etc/ssh/ssh_config
```

?

dscho: Actually, better test for `! grep -q ^Ciphers [a-z].*cbc`, to avoid catching the edit below.

git-extra/git-extra.install
```diff
@@ -219,9 +219,14 @@ test -d "$TMPDIR" || test ! -d "$TMP" || {\
 	# Re-enable at least AES256-cbc and AES128-cbc, as they are still used
 	# e.g. in VSTS
 	test ! -f /etc/ssh/ssh_config ||
-	grep -q '^Ciphers .*aes256-cbc,aes192-cbc' /etc/ssh/ssh_config ||
-	sed -i -e '/^[# ]*Ciphers /{s/^# *//;s/$/,aes256-cbc,aes192-cbc/}' \
-		/etc/ssh/ssh_config
+	{
```
dscho: This curly bracket (and the corresponding indentation) is not really needed, is it?

drizzd: We have a similar situation compared to above in https://github.com/git-for-windows/build-extra/pull/192/files#diff-5a39db150c328ad0b5127f77720de3edR161. If /etc/ssh/ssh_config exists, then we want to do two things: revert change by prioer versions of git-extra and append the new incremental ciphers list.

But we can do this instead, if you prefer:

```
test ! -f /etc/ssh/ssh_config ||
! grep -q '^Ciphers [a-z].*cbc' ||
sed -i -e 's/^Ciphers [a-z].*cbc/# &/' /etc/ssh/ssh_config

test ! -f /etc/ssh/ssh_config ||
grep -q '^Ciphers\s' /etc/ssh/ssh_config ||
sed -i -e '$a# Added by git-extra\nCiphers +aes128-cbc,3des-cbc,aes256-cbc,aes192-cbc' \
    /etc/ssh/ssh_config
```

git-extra/git-extra.install
drizzd: Ok.

git-extra/git-extra.install
dscho: Oooops! I totally missed that there was still a `test ! -f /etc/ssh/ssh_config ||` in front of the curlies! Please accept my apologies, my review was incorrect.

So maybe

```sh
test ! -f /etc/ssh/ssh_config || {
    ! grep -q '^Ciphers [a-z].*cbc' ||
    sed -i -e 's/^Ciphers [a-z].*cbc/# &/' /etc/ssh/ssh_config

    grep -q '^Ciphers\s+' /etc/ssh/ssh_config ||
    printf '%s\n%s\n' \
        '# Added by git-extra'
        'Ciphers +aes128-cbc,3des-cbc,aes256-cbc,aes192-cbc' \
        >/etc/ssh/ssh_config
}
```

drizzd: Not at all. Thank you for the review. Indeed, the extra blank line makes this more readable, and I aligned the formatting of the curly braces as suggested.

dscho:
Thank you, @drizzd!

PR #198, opened 2018-08-14
Skip gitweb in Git for Windows

RandomSort:
Long time user, first time contibutor :)

I've solved the ticket here: https://github.com/git-for-windows/git/issues/1362
I have not done anything extra not covered in the ticket. This seems to solve the issue, I have not located any GitWeb content that is not excluded by the simple pattern.

I reference the issue in the body of the commit, let me know if that is considered bad practice.

Thank you for your great work!

Feel free to add feedback, and I will happily rework the patch.


dscho:
Thank you so much! This is perfect.

PR #199, opened 2018-08-15
Exclude xsltproc.exe from installer/portable Git/etc

dscho:
It was reported in https://github.com/git-for-windows/git/issues/1775 that xsltproc does not work due to a missing .dll. Keeping in mind that xsltproc is not at all required for Git to work, let's exclude that executable.


jamill:
This looks good to me - thanks!

PR #200, opened 2018-08-16
VS Code, Sublime Text, Atom as text editors

ungps:
In response to: https://github.com/git-for-windows/git/issues/1795

P.S: "VS Code - Insiders" does not work for me (it is not even detected), but I hope I will be able to fix this too.


ungps:
Ok, I thought I would only fix the problem I reported [here](https://github.com/git-for-windows/git/issues/1795), but since I was still here, I pushed some more commits to:

* Add Sublime & Atom as options when chosing the default text editor. 

* Fix the problem when VSCode is not detected (https://github.com/git-for-windows/git/issues/1741#issuecomment-413834357) by looking in registry at both locations.

* Add an option to allow the user to select any editor he wants.

ungps:
Indeed, I will resend a fix for this as soon as possible.

installer/install.iss
```diff
@@ -1133,7 +1133,11 @@ begin
 
     EditorAvailable[GE_NotepadPlusPlus]:=RegQueryStringValue(HKEY_LOCAL_MACHINE,'SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\notepad++.exe','',NotepadPlusPlusPath);
     EditorAvailable[GE_VisualStudioCode]:=RegQueryStringValue(HKEY_LOCAL_MACHINE,'SOFTWARE\Classes\Applications\Code.exe\shell\open\command','',VisualStudioCodePath);
+    if (not EditorAvailable[GE_VisualStudioCode]) then
+        EditorAvailable[GE_VisualStudioCode]:=RegQueryStringValue(HKEY_CURRENT_USER,'Software\Classes\Applications\Code.exe\shell\open\command','',VisualStudioCodePath);
```
dscho: Question 1: does this even work? Git for Windows' installer runs elevated, so I could imagine that this change makes the installer look in the *administrator's* `HKEY_CURRENT_USER` namespace. And that would not catch the original user's VS Code installation, right?

Question 2 (or actually, more of a very strong suggestion): isn't it inappropriate for a system-wide installation of Git for Windows to depend on a user-specific VS Code installation? In general, other users would not even be able to access that VS Code, and even then, it would wreak havoc if the original user wants to update their VS Code and another user still uses it. So if we go the route at all to detect a user setup of VS Code, I am fairly convinced that we should *not* install that via `git config --system core.editor <path>` but instead via `git config --global core.editor <path>` (of course, launched via [`ExecAsOriginalUser()`](http://www.innosetup.com/ishelp/topic_isxfunc_execasoriginaluser.htm)). This will complicate the code quite a bit, though, as we now have to record not only the path but also the fact that VS Code is to be configured for the current user only.

dakotahawkins:
@dscho Would it be possible to set `core.editor --system` to a "meta" value that triggers a runtime check followed by setting `core.editor --global` on first use?

ungps:
I force pushed some changes and I hope that the first three commits (all, but "installer: add option to allow unlisted editors") are now ok. I tested it by using the command ` ./installer/release.sh --debug-wizard-page="Editor"`. It takes some time to build the installer. I will come back later (after I actaully test the installer, not by just debugging the wizard page) with a reply to confirm if it actually works ok.

shiftkey:
I think the VSCode team is going to try and improve their installer so that they can set the _global_ Git config to VSCode, meaning it might be unnecessary to figure out how to support these new "user install" modes in a system config: https://github.com/Microsoft/vscode/issues/56973

cc @joaomoreno for any other initial details

joaomoreno:
@shiftkey I would actually update the git's installer to also detect Code's user install. Is there any way I can help here?

shiftkey:
@joaomoreno I think the big questions I have were in previous comments from @dscho - https://github.com/git-for-windows/build-extra/pull/200#discussion_r211560985

installer/install.iss
joaomoreno: Answer 1: Yes, the installer needs to look into the user's HKCU.

Answer 2: It does seem a bit backwards. But if done, it should definitely go into the user's `--global`.

shiftkey:
> Bottom line is: Github for desktop team's installer does not work ...

@anthcool I'm a maintainer of [GitHub Desktop](https://desktop.github.com) and we should support both user and admin VSCode installers since 1.3.4 last month. Please ensure you're up to date, or log an issue over on our repository with more details: https://github.com/desktop/desktop.

This issue is for the Git for Windows installer.

anthcool:
@shiftkey Thanks for clearing this up for me.  I feel like a dork.  The whole time I was trying to update Git itself to version 2.19.0 from 2.18.0.windows.1 and am having this same problem.  If I try to update Git as described, using the installer downloaded from [https://git-scm.com/downloads](https://git-scm.com/downloads), and get to the default editor picker part, as soon as I go to choose VS Code as the default, or anything other than Vim, the Next button grays out/disables....  Ugh..  Okay, where do I submit an issue to the Git team then?  Thanks again though!

shiftkey:
@anthcool it's this bit here that I believe is the blocker on any real solution https://github.com/git-for-windows/build-extra/pull/200#discussion_r211560985

dscho:
>     * VSCodeInsiders 1.27 - doesn't work. (`core.editor` is not set). This looks into HKCU.

I tried with VS Code Insiders 1.28 and it did work...

Also, I verified that, for some reason, InnoSetup can see my user registry even if I run the installer elevated (as asked for in non-debugging mode). I have a couple of patches on top that I will push in a moment.

dscho:
Actually, I had tried only the debug mode to see the editor page, and replicated the issue reported by @ungps. I figured it out and only need to clean up the changes, but I have to be away from the keyboard for a few hours first.

dscho:
There is only one remaining problem I can see with this PR: when a user chose VS Code that was installer for the current user only, and next time chooses an editor that is not only for the current user, the `~/.gitconfig` entry will not be removed. However, this entry might have been changed in the meantime, too, in which case we should *not* touch it anyway! Not sure what to do about it... I guess we could store (in the PreviousData) the editor path installed via `git config --global`, and the next time, just before we are about to configure an editor option, test whether the user config still refers to the same path and if so, unset that setting.

Any thoughts?

installer/install.iss
dscho: The problem with answer 2: we now have no longer any way to determine whether the setting in `~/.gitconfig` was configured by the installer (and should hence be removed upon uninstall), or by the user. Do you have any ideas how to solve that conundrum?

PR #202, opened 2018-08-22
Exclude git-archimport from installer/portable Git/etc

jamill:
This is a component that we expect does not get much use. Instead of distributing this component with Git for Windows, remove it. Removing this component will reduce the work associated the release process.

Signed-off-by: Jameson Miller <jamill@microsoft.com>


dscho:
Thanks!

PR #203, opened 2018-08-25
installer: offer the builtin stash/rebase as experimental options

dscho:
As nice as the speed-ups are, the patches in question are still in flux, and they are not battle-tested at all.

Let's use the scripted commands by default, and offer the new, fast, experimental builtins as options.

This gives us the best of both worlds: users who want the raw speed improvement we got through three Google Summer of Code projects working in parallel can have that, while others who are reluctant to play guinea pig by running only well-tested code can stay on the safe side.

Plus, it gives us a chance of battle-testing those patches :-)

Note: the code added in this PR will only kick in if bundling a Git built including the latest patches of https://github.com/git-for-windows/git/pull/1800.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -338,8 +338,22 @@ const
     GP_Symlinks       = 3;
 
 #ifdef WITH_EXPERIMENTAL_BUILTIN_DIFFTOOL
+#define HAVE_EXPERIMENTAL_OPTIONS 1
+#endif
+
+#ifdef WITH_EXPERIMENTAL_BUILTIN_REBASE
+#define HAVE_EXPERIMENTAL_OPTIONS 1
+#endif
+
+#ifdef WITH_EXPERIMENTAL_BUILTIN_STASH
+#define HAVE_EXPERIMENTAL_OPTIONS 1
+#endif
+
+#ifdef HAVE_EXPERIMENTAL_OPTIONS
     // Experimental options
     GP_BuiltinDifftool = 1;
+    GP_BuiltinRebase   = 2;
+    GP_BuiltinStash    = 3;
 #endif
 
 var
@@ -385,10 +399,10 @@ var
     ExtraOptionsPage:TWizardPage;
     RdbExtraOptions:array[GP_FSCache..GP_Symlinks] of TCheckBox;
 
-#ifdef WITH_EXPERIMENTAL_BUILTIN_DIFFTOOL
+#ifdef HAVE_EXPERIMENTAL_OPTIONS
     // Wizard page and variables for the experimental options.
     ExperimentalOptionsPage:TWizardPage;
-    RdbExperimentalOptions:array[GP_BuiltinDifftool..GP_BuiltinDifftool] of TCheckBox;
+    RdbExperimentalOptions:array[GP_BuiltinDifftool..GP_BuiltinStash] of TCheckBox;
 #endif
 
     // Mapping controls to hyperlinks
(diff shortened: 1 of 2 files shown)


dscho:
> My concern is that we are adding more options to a setup that already exposes a lot of options to users. If I wasn't already familiar with the options, I am not sure how I would make an informed decision about what option I should choose based on the information presented in this dialog.

Any suggestions how to word it?

dscho:
Screenshot:

image

bviktor:
[Nevermind](https://github.com/git-for-windows/build-extra/commit/263f64366b886908b898e85eccfd77cf88e31307) :)

```
config --global rebase.useBuiltin true
config --global stash.useBuiltin true
```

Qartar:
Is it too late to change the wording on the setup page? "90% faster" is [incredibly ambiguous](https://randomascii.wordpress.com/2018/02/04/what-we-talk-about-when-we-talk-about-performance/).

shiftkey:
@Qartar the source for the wording is here if you'd like to submit some better wording:

https://github.com/git-for-windows/build-extra/blob/398071c90575299a25e4d0ddea6582c2e4ecbc2e/installer/install.iss#L1414-L1428

Qartar:
Thanks, but I'm not sure what the intended value is, e.g. is it almost twice as fast or is it ten times faster?

dscho:
@Qartar I guess you're right. Have you tested the performance? What I meant was a 10x speed-up. But now that you mention it, I would like to have a bit more confirmation on those numbers from other people.

randomascii:
Uggh. Is it too late to change the "90% faster" text to say "10x as fast"?

90% faster is hopelessly ambiguous, and it hugely undersells the value of the optimizations if they really are supposed to be 10x as fast.

PhilipOakley:
> Is it too late to change the "90% faster" text to say "10x as fast"?

No. Pull requests welcome. (though it won't show until the next build... but it will still be valuable)

palver123:
I agree with @jamill , the installer is already a bit jam-packed with technical decisions. Even as a software developer I am somewhat daunted when I see it.

I am frequently asked by non-IT colleagues to help with the installation of Git on their (Windows) computer because they are afraid of this installer and the questions asked in it.

coachmullen22:
> > My concern is that we are adding more options to a setup that already exposes a lot of options to users. If I wasn't already familiar with the options, I am not sure how I would make an informed decision about what option I should choose based on the information presented in this dialog.
> 
> Any suggestions how to word it?

I am someone who doesn't know which "bleeding edge features" I would like to enable. It's frustrating to slog through setting up my computer so I can learn to code, only to come across two options that may not mean much to me yet...but I still want to make a semi-informed choice. What do they do differently from one another except for running 70% faster or 90% faster than a completely undefined processing speed?

I think I prefer the word 'stash' over 'rebase'. Besides it's (1.9x-1.7x) faster than just x.

Seriously though, I know you're all working on a level I can only aspire to reach someday. Just keep us newbies in mind...

Rutix:
@coachmullen22 'stash' is a git feature though so 'stash' and 'rebase' are different things. So you don't have to choose between them, but it's that both of them received an experimental option

dscho:
Well, the entire discussion is moot now, as v2.19.2 no longer marks those built-in commands as experimental.

Sarkie:
@dscho Just to be clear 

`git config --global rebase.useBuiltin true
git config --global stash.useBuiltin true`

Are enabled by default as of 2.19.2, so not needed to do manually run the commands?

dscho:
@Sarkie it is even "worse": those config settings' *default* value has changed. Read: if you do not configure `stash.useBuiltin` *at all*, it will use the built-in command.

PR #206, opened 2018-09-28
Added support for user-level Visual Studio Code

phistuck:
Since this is the default now and they recommend system level users to switch with a notification, it is going to become more widespread and the current installer does not detect it.

Signed-off-by: PhistucK <phistuck@gmail.com>


dscho:

@ungps already tackled this in https://github.com/git-for-windows/build-extra/pull/200, though, and took even care of configuring the editor setting only for the installing user if the editor was installed only for the installing user.

There was a tiny bug in that, but I managed to figure it out and will fix it probably today, still.

In light of that, I hope you don't mind that I will fix #200 and merge it instead of this PR?

phistuck:
Sure. :)

PR #207, opened 2018-10-11
Make default editor message clear

aakriti-jain:
This fixes the issue/feature request mentioned in git-for-windows/git#1872.
As suggested by @PhilipOakley , I have added an extra note to make it more clear what the Vim option actually does.

capture

Signed-off-by: aakriti-jain <aakritijain1102@gmail.com>

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1324,7 +1324,7 @@ begin
     // 2nd choice
     Top:=TopOfLabels;
     CbbEditor.Items.Add('Use Vim (the ubiquitous text editor) as Git'+#39+'s default editor');
-    CreateItemDescription(EditorPage,'The <A HREF=http://www.vim.org/>Vim editor</A>, while powerful, <A HREF=https://stackoverflow.blog/2017/05/23/stack-overflow-helping-one-million-developers-exit-vim/>can be hard to use</A>. Its user interface is'+#13+'unintuitive and its key bindings are awkward.'+#13+#13+'<RED>Note:</RED> Vim is the default editor of Git for Windows only for historical reasons, and'+#13+'it is highly recommended to switch to a modern GUI editor instead.',Top,Left,LblEditor[GE_VIM],False);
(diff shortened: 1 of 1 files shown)


aakriti-jain:
> * can you check what the installer does if the user had previously set the EDITOR variable - does it leave it alone, or deliberately unset it?
No it doesn't changes the EDITOR variable. Just checked it.

In this case, I think we should make two options for user, one for the default and one for vim because it could be misleading for the user otherwise. In the default option, we could mention that default editor in the Bash environment is Vim.

aakriti-jain:
The requested changes have been done @PhilipOakley . Heres a screenshot -
capture1

Waiting for your review.

dscho:
Excellent, thank you so much @aakriti-jain!

> The only improvement that I could recommend would be including a link to [the git-var documentation](https://git-scm.com/docs/git-var#git-var-GITEDITOR), because it really spells out exactly what is going on with the different configuration options (there are five!) that can influence which editor Git uses.

@zwilliams feel free to work on that. It should be as easy as adding a `<A HREF="<url>">...</A>` construct (which we emulate in the installer script).

dscho:
Excellent, thank you so much @aakriti-jain!

> The only improvement that I could recommend would be including a link to [the git-var documentation](https://git-scm.com/docs/git-var#git-var-GITEDITOR), because it really spells out exactly what is going on with the different configuration options (there are five!) that can influence which editor Git uses.

@zwilliams feel free to work on that. It should be as easy as adding a `<A HREF="<url>">...</A>` construct (which we emulate in the installer script).

dscho:
> Hard, indeed, unfamiliar, yes, but neither awkward nor obsolete.

The installer doesn't state that it is obsolete. But awkward it is, the entire vim experience. I am a long-time vim user, and it is still my go-to editor on the console. And even for me, the handling is awkward at best.

In any case, @jeromedoucet if you truly care, you will open a new Pull Request and not only add a comment to an already-closed Pull Request...

PR #209, opened 2018-10-25
Add perl JSON module

orgads:
It is rather small, and comes in handy for scripts


orgads:
That's +19K to the installer.

dscho:
Thank you!

PR #213, opened 2018-11-06
please.sh: script to fix duplicate relnote entry

tanushree27:
when packages (like 'Git Credential Manager') are upgraded twice (or
more times) between Git for Windows releases, duplicate entries are
created in 'ReleaseNotes.md'.

Subcommand `mention` writes/edits the 'ReleaseNotes.md'.

Added script to delete every match for 'Comes with [<package> v' before
appending new entry 'Comes with [<package> <version>](<url>)'.

this closes issue https://github.com/git-for-windows/git/issues/1898

### DIFF
--- a/please.sh
+++ b/please.sh
@@ -3453,6 +3453,7 @@ mention () { # <what, e.g. bug-fix, new-feature> <release-notes-item>
 		sed -i -e "/^## Changes since/{s/^/$quoted\n\n/;:1;n;b1}" \
 			"$relnotes"
 	else
+		search=$(echo $quoted | sed -r -e 's#.*Comes with \[(.* v|patch level).*#\1#')
 		sed -i -e '/^## Changes since/{
 			:1;n;
 			/^### '"$what"'/b3;
@@ -3460,12 +3461,15 @@ mention () { # <what, e.g. bug-fix, new-feature> <release-notes-item>
 			/^## Changes since/b2;
 			b1;
 
-			:2;s/^/### '"$what"'\n\n'"$quoted"'\n\n/;b5;
+			:2;s/^/### '"$what"'\n\n'"$quoted"'\n\n/;b7;
 
-			:3;/^\*/b4;n;b3;:4;n;/^\*/b4;
-			s/^/'"$quoted"'\n/;b5;
+			:3;/^\*/b4;n;b3;
 
-			:5;n;b5}' "$relnotes"
+			:4;/'"$search"'/b5;n;b6;:5;N;s/^.*\n//;:6;/^\*/b4;
+
+			s/^/'"$quoted"'\n/;
+
+			:7;n;b7}' "$relnotes"
 	fi ||
 	die "Could not edit release notes\n"
 


please.sh
```diff
@@ -3453,6 +3453,7 @@ mention () { # <what, e.g. bug-fix, new-feature> <release-notes-item>
 		sed -i -e "/^## Changes since/{s/^/$quoted\n\n/;:1;n;b1}" \
 			"$relnotes"
 	else
+		search=$(echo $quoted | sed -r -e 's|.*Comes with \[(.* v).*|\1|')
```
dscho: I think this breaks for the MSYS2 runtime upgrades... (see bd6aa98e43744625c0027c945726c30c9449e001 for an example outcome of the new feature).

please.sh
```diff
@@ -3462,8 +3463,13 @@ mention () { # <what, e.g. bug-fix, new-feature> <release-notes-item>
...
 
-			:3;/^\*/b4;n;b3;:4;n;/^\*/b4;
-			s/^/'"$quoted"'\n/;b5;
+			:3;/^\*/b4;n;b3;
+			
+			:4;/.*'"$search"'.*/b6;n;/^\*/b4;
+
+            :7;s/^/'"$quoted"'\n/;b5;
+
+			:6;N;s/.*\n//;/^\*/b4;b7;
```
dscho: Hehe... it is tempting to extend the sed script, but it does look more convoluted yet than before. I guess I regret my choices, and I should just have gone with a node.js or Perl snippet here?

If you move this *before* the `:7` line, you can avoid the `b7` altogether (and hence also the label `:7`). Then also that `b5` in the line starting with `:7` would become unnecessary again (I just realized that I left that in by mistake).

The `s/.*\n//` is correct, though: the `.` does not match a newline, so this really only deletes the (just added) first line of the buffer. To make this a bit more explicit, the anchor `^` could be used.

Oh, and could you adjust the labels so that they are in increasing order again?

From what I can tell, this code works correctly, though.

please.sh
```diff
@@ -3462,8 +3463,13 @@ mention () { # <what, e.g. bug-fix, new-feature> <release-notes-item>
 
 			:2;s/^/### '"$what"'\n\n'"$quoted"'\n\n/;b5;
 
-			:3;/^\*/b4;n;b3;:4;n;/^\*/b4;
-			s/^/'"$quoted"'\n/;b5;
+			:3;/^\*/b4;n;b3;
+			
+			:4;/.*'"$search"'.*/b6;n;/^\*/b4;
```
dscho: Those `.*` should not be necessary: sed's regexes are by default unanchored.

please.sh
```diff
@@ -3462,8 +3463,13 @@ mention () { # <what, e.g. bug-fix, new-feature> <release-notes-item>
...
+
+            :7;s/^/'"$quoted"'\n/;b5;
```
dscho: Could you please adjust the indentation to match the other lines'?

please.sh
dscho: Trying to be a bit more clear here: to me, the `:6` seems like an `else` clause. So I could see it being the `/<delete-condition>/b5;n;b6:5;N;:6` dance below (where `b5` was renamed to `b7`):

```
:1;n;
/^### '"$what"'/b3;
/^### Bug Fixes/b2;
/^## Changes since/b2;
b1;

:2;s/^/### '"$what"'\n\n'"$quoted"'\n\n/;b7;
:3;/^\*/b4;n;b3;

:4;/.*'"$search"'.*/b5;n;b6:5;N;s/^.*\n//:6;/^\*/b4;
s/^/'"$quoted"'\n/;
:7;n;b7
```

There is still one caveat, in that the `search` variable might contain too ambiguous a needle, and that it will be interpreted as regex, so special characters should probably be escaped by a backslash.

dscho:
Thank you! I adjusted the branch target from `b5` to `b7` in the line

```
:2;s/^/### '"$what"'\n\n'"$quoted"'\n\n/;b5;
```

because the final loop that simply copies the rest of the file now starts with `:7`, no longer with `:5`. Otherwise the patch is very fine, I tested it and it works as far as I can tell.

Well done!

PR #214, opened 2018-11-15
Deprecate Git CMD

dscho:
We talked about this recently, and I think it is time. There is one particular issue with CMD that opens a particular attack vector that is unavailable for Git Bash and Powershell users: when calling `git log` in a directory in which there is a `git.exe`, that executable is used, no matter how you configure `PATH` or CMD. So if some creative user talks another user into cloning a repository with `git.exe` in the top-level directory, and that other user is unsuspecting of this behavior, they might be talked into calling `git log` and actually run untrusted code.

To raise awareness of this issue, and at the same time to diminish chances of such social-engineering attacks, let's deprecate Git CMD.

Note: this behavior is neither new, nor limited to Git. It just so happens that we can at least deprecate Git CMD, and that's what we should do.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -35,6 +35,8 @@ This package contains software from a number of other projects including Bash, z
 
 ## Changes since Git for Windows v2.19.1 (Oct 5th 2018)
 
+Please note: _Git CMD_ is deprecated as of this Git for Windows version. The default is to have `git.exe` in the `PATH` anyway, so there is no noticeable difference between CMD and Git CMD. It is impossible to turn off CMD's behavior where it picks up any `git.exe` in the current directory, so let's discourage the use of Git CMD. Users who dislike Git Bash should switch to Powershell instead.
+
 ### New Features
 
 * Comes with [OpenSSH v7.9p1](https://www.openssh.com/txt/release-7.9).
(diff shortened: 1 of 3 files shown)


dscho:
@shiftkey your reviewer comments are the best. THE BEST!

jamill:
Is this deprecating the use of Git from the cmd prompt in general? or just a the cmd prompt that is set up specifically to point to the installed version of Git?

The advice to use Powershell if you don't like Git Bash makes it sound like we are depracating the use of git from the cmd prompt.

shiftkey:
@jamill I believe it is this case that @dscho wants to actively discourage:

> or just a the cmd prompt that is set up specifically to point to the installed version of Git?

dscho:
> Is this deprecating the use of Git from the cmd prompt in general?

No, it is just trying to discourage users from using `Git CMD`. If there is a CMD power user who knows fully well what the security implications of CMD are, they are more than welcome to use it.

In other words: we will keep our default setting where `git.exe` is added to the `PATH` (or more precisely, `C:\Program Files\cmd` is added to the `PATH`).

Please let me know of any improvements to the wording in the commit message and/or Release Notes!

dscho:
> The advice to use Powershell if you don't like Git Bash makes it sound like we are depracating the use of git from the cmd prompt.

I do not want to deprecate calling `git.exe` from CMD. I want to discourage that. What I want to deprecate is the `Git CMD` start menu entry.

bbolli:
I've never used Git CMD, but then I'm very happy with bash and VIM :-)

gistofj:
Yes, o'gawd kill it! Kill it with fire :fire: 👍

dscho:
That's all the encouragement I needed 😆

PR #216, opened 2018-11-27
Use the Command Prompt as own name

mloskot:
Clarifies the Command Prompt is an own name, and qualifying it with Windows, i.e. Windows Command Prompt, is redundant. Similarly, full name of PowerShell on Windows is Windows PowerShell.

Mention the Windows PowerShell in the 2nd choice description for even more clarity.

## References

- Follow-up to the ML thread [Git for Windows v2.20.0-rc0](https://groups.google.com/forum/#!msg/git-for-windows/SFqyASXbGxE/v1nSwrOhCQAJ), where I argued use of the "Windows Command Prompt" in

    > That setting "Use Git from the Windows Command Prompt" affected PowerShell
    > already: by "Windows Command Prompt", I refer to both CMD and PowerShell:
    > both are Windows Command Prompts.

    is confusing, since

    > Shortly, there are two native command prompts in Windows with distinct names:
    > - Command Prompt
    > - Windows PowerShell

    If I'm still trying to be anal about the naming, trying to find a rationale of it, I'd have considered the "Windows Command Prompt" as shortcut of "Windows' Command Prompt", "In Windows, Command Prompt" or "Command Prompt on Windows" :-)

    Possibly, "Windows Command-line" may be better name for all command-line flavours on Windows, than "Windows Command Prompt".

- https://en.wikipedia.org/wiki/PowerShell although now portable, for GfW case, it is clearer to state the old name:

   > Initially a Windows component only, known as Windows PowerShell

- https://en.wikipedia.org/wiki/Cmd.exe uses name "Command Prompt"
    
    > In Windows, Command Prompt

- MSDN and docs.microsoft.com use "The Developer Command Prompt", not "The Developer Windows Command Prompt", to refer to the Command Prompt configured for Visual Studio toolsets.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1436,10 +1436,10 @@ begin
     RdbPath[GP_BashOnly]:=CreateRadioButton(PathPage,'Use Git from Git Bash only','This is the safest choice as your PATH will not be modified at all. You will only be'+#13+'able to use the Git command line tools from Git Bash.',TabOrder,Top,Left);
 
     // 2nd choice
-    RdbPath[GP_Cmd]:=CreateRadioButton(PathPage,'Use Git from the Windows Command Prompt','This option is considered safe as it only adds some minimal Git wrappers to your'+#13+'PATH to avoid cluttering your environment with optional Unix tools. You will be'+#13+'able to use Git from both Git Bash and the Windows Command Prompt.',TabOrder,Top,Left);
+    RdbPath[GP_Cmd]:=CreateRadioButton(PathPage,'Git from the command line and also from 3rd-party software','This option is considered safe as it only adds some minimal Git wrappers to your'+#13+'PATH to avoid cluttering your environment with optional Unix tools.'+#13+'You will be able to use Git from Git Bash, the Command Prompt and the Windows PowerShell as well as any third-party software looking for Git in PATH.',TabOrder,Top,Left);
 
     // 3rd choice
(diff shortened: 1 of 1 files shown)


installer/install.iss
```diff
@@ -1436,7 +1436,7 @@ begin
     RdbPath[GP_BashOnly]:=CreateRadioButton(PathPage,'Use Git from Git Bash only','This is the safest choice as your PATH will not be modified at all. You will only be'+#13+'able to use the Git command line tools from Git Bash.',TabOrder,Top,Left);
 
     // 2nd choice
-    RdbPath[GP_Cmd]:=CreateRadioButton(PathPage,'Use Git from the Windows Command Prompt','This option is considered safe as it only adds some minimal Git wrappers to your'+#13+'PATH to avoid cluttering your environment with optional Unix tools. You will be'+#13+'able to use Git from both Git Bash and the Windows Command Prompt.',TabOrder,Top,Left);
+    RdbPath[GP_Cmd]:=CreateRadioButton(PathPage,'Use Git from the Command Prompt','This option is considered safe as it only adds some minimal Git wrappers to your'+#13+'PATH to avoid cluttering your environment with optional Unix tools. You will be'+#13+'able to use Git from Git Bash, the Command Prompt and the Windows PowerShell.',TabOrder,Top,Left);
```
dscho: Do we want to say "Use Git from the Command Prompt/PowerShell" here?

Or maybe it would be even better to phrase it yet another way: we are not only making it possible to use Git from the command line, but also from 3rd-party software.

Maybe "Expose git.exe to the command line and other applications"? Dunno.

dscho:
@mloskot thank you for opening this PR! I hope that we can come to a consensus about the best phrasing soon and merge it.

installer/install.iss
dscho: > perhaps ", but" could read "and", better?

👍

dscho: Can you make it so and force-push an update?

mloskot: Done

dscho:
Thank you!!!

mloskot:
I'm glad we've worked it out :-) Thanks

PR #217, opened 2018-12-03
Update 2.19.1 release notes for clearer cmd.exe deprecation

mjcheetham:
Some people may be confused about the meaning of deprecation
of "Git CMD". Rewording the release notes to call out that:

- using _Command Prompt (cmd.exe)_ is deprecated,
- the _Git CMD_ start menu shortcut (which configures a cmd.exe
  instance) is really the same as a vanilla cmd.exe instance.


dscho:
Thank you!

PR #218, opened 2018-12-05
notepad: don't break comments in COMMIT_EDITMSG

rimrul:
Wrapping lines with fmt could cause comments in COMMIT_EDITMSG to turn into a
comment line and one or more non-comment lines. Wrapping just the comments with
the -p option before wrapping the other lines prevents this from happening.

This fixes git-for-windows/git#1967

I've marked this as "don't merge yet", because I haven't tested it yet. Will update this PR soon.


dscho:
Very neat, thanks!

PR #220, opened 2018-12-06
Fix the SDK installer so it respects target directories with spaces in their path

dscho:
It was reported in https://github.com/git-for-windows/git-sdk-64/issues/6 that installing into a directory whose path contains a space does not work.

While at it, I had to fix the problem where `.dll` files required by `git-remote-https.exe` were not picked up if they existed in `libexec/git-core/`, and I also activated the FSCache for the checkout.

### DIFF
--- a/sdk-installer/release.sh
+++ b/sdk-installer/release.sh
@@ -48,10 +48,13 @@ do
 	exes_and_dlls="$exes_and_dlls$file "
 
         for dll in $(ldd "$BIN_DIR/$file" |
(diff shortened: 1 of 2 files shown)


aakriti-jain:
Yes sure @dscho ! I will review this at once!

aakriti-jain:
@dscho The installer now perfectly installs sdk in paths with spaces! This solves the original issue.
But I am not much used to regex, so I cannot review your first commit as I may miss some pointers.

Thanks!

PR #221, opened 2018-12-07
Fixes “Not a valid executable” error in installer

aakriti-jain:
This issue was addressed in git-for-windows/git#1951.
Currently, I have only edited the error message as it only showed the first 3 characters if the path was not of a valid .exe file.

I need a suggestion: Should I also make it support CMD commands using %PATH% enviroment?  Currently the message is apt for its working.

Signed-off-by: Aakriti Jain <aakritijain1102@gmail.com>


installer/install.iss
```diff
@@ -1215,21 +1215,26 @@ begin
...
     Path:=EditorPage.Values[0]+' ';
-    if not WildcardMatch(Lowercase(Path), '"*.exe" *') then
-        PathLength:=Pos('.exe ',Lowercase(Path))+3
-    else begin
-        PathLength:=Pos('.exe" ',Lowercase(Path))+2;
-        Path:=Copy(Path,2,PathLength)+Copy(Path,PathLength+3,Length(Path))
-    end;
+    if Pos('.exe',Lowercase(Path))=0 then begin
+    	CustomEditorPath:=EditorPage.Values[0];
+        CustomEditorOptions:='';
```
dscho: This would neatly fit in the previous code's `else` clause:

```pascal
    if not WildcardMatch(Lowercase(Path), '"*.exe" *') then
        PathLength:=Pos('.exe ',Lowercase(Path))+3
    else if Pos('.exe',Lowercase(Path))=0 then begin
        CustomEditorPath:=EditorPage.Values[0];
        CustomEditorOptions:='';
    else begin
        PathLength:=Pos('.exe" ',Lowercase(Path))+2;
        Path:=Copy(Path,2,PathLength)+Copy(Path,PathLength+3,Length(Path))
    end;
```

But what about `code --wait`? It would fall into that new clause, and assume that the entire string is the path to the executable.

I would think, however, that it would maybe make more sense to replace the `"*.exe" *` pattern by `"*" *`, and then just search for `" ` instead of `.exe ` in line 1221.

But I guess the entire code is confusing, because the first clause is about *not* matching that pattern. So how about this:

```pascal
    if WildcardMatch(Lowercase(Path), '"*" *') then
        PathLength:=Pos('" ',Lowercase(Path));
        CustomEditorPath:=Copy(Path,2,PathLength);
        CustomEditorOptions:=Copy(Path,PathLength+2,Length(Path))
    else if Pos(' ',Lowercase(Path))>0 then begin
        PathLength:=Pos(' ',Lowercase(Path));
        CustomEditorPath:=Copy(Path,1,PathLength);
        CustomEditorOptions:=Copy(Path,PathLength+2,Length(Path))
    else begin
        CustomEditorPath:=Path;
        CustomEditorOptions:=''
    end;
```

aakriti-jain: Currently the text field supports path with spaces (Like - C:\Program Files\editor.exe -w) but in your implementation, it would fall in the second clause making things go wrong.

I can surely make this nested if to if..elseif..else way you suggested. I will make this change in next commit.

aakriti-jain:
The code is currently such that after every change, it checks weather the variable `CustomEditorPath` exists and is an .exe file, and enables/disables the Next button accordingly. That means the status of button changes after every hit of keyboard.
I may be wrong, but wont this approach make a lot of overhead?

But still I am working on the solution for it. I will let you know if I manage to get it to work.

installer/install.iss
dscho: > Currently the text field supports path with spaces (Like - C:\Program Files\editor.exe -w) but in your implementation, it would fall in the second clause making things go wrong.

That's a good point. So we should first check for quoted `.exe`, then unquoted `.exe`, then quoted path without extension, then unquoted path without extension, right?

dscho:
> The code is currently such that after every change, it checks weather the variable `CustomEditorPath` exists and is an .exe file, and enables/disables the Next button accordingly. That means the status of button changes after every hit of keyboard.
> I may be wrong, but wont this approach make a lot of overhead?

You're right!

I was wrong to mention that function. What I meant was the validation that would bug the user with that error message unless the path exists and ends in `.exe`.

dscho:
Please note that Git v2.20.0 has already been released, and therefore I want to publish Git for Windows v2.20.0 as soon as possible.

Unfortunately, this means that this PR won't make it in before that version, but given that the custom editor option is relatively new (and does not seem to be used very often yet), I think we can take our time to hammer out the kinks and get it into v2.20.1 (or v2.20.0(2), or v2.21.0, whichever comes first after we're done with this PR). Does that sound okay, @aakriti-jain?

aakriti-jain:
Sorry @dscho I had exams in my college and had to pause the work on it. I will surely work on it and complete this task surely before the next release.

dscho:
@aakriti-jain no worries.  Do you want me to pick up some tasks, or do you want me to wait for you to do it all?

dscho:
@aakriti-jain excellent. I think we have enough time. Just let me know whenever I can be of assistance.

aakriti-jain:
@dscho I have written the code as you suggested such that function `PathIsValidExecutable` now searches for every executable (in `%PATHEXT%`) in locations in `%PATH%`. And there is no noticeable lag too 😄 
But when I test it with Sublime Text using the command `subl -w`, it would just get stuck, even after editing the temporary file and closing the editor. I tried this with the released version of Git for Windows (but instead giving complete path), had the same problem. VSCode works smoothly though.

So I guess this was not an error of the new code that I wrote, but with the Sublime Text instead. So should I try to fix this in the same PR or move it to a second?

dscho:
@aakriti-jain thank you so much! I worked a bit on top of your patches, and I think it should work with Sublime Text now. Could I ask you to test?

aakriti-jain:
Yes I will test it now! Sorry I was out of station these days and couldn't test it then. I will just test and complete this issue.

aakriti-jain:
@dscho all the cases have been tested and most of them worked perfectly fine! Thank you so much for the extra patches you made!

Just a thing more: This option is still disabled-
capture

I have Sublime Text 3 properly installed but its not showing so.

dscho:
> I have Sublime Text 3 properly installed but its not showing so.

Hmm. I don't have that installed...

Could you maybe debug by placing strategic "debug print statements" (I usually abuse `LogError()` for this purpose) and then run `./installer/release.sh -d Editor` in `/usr/src/build-extra/`?

aakriti-jain:
Can I work on it a bit later in a separate PR? Been a bit busy these days

dscho:
> Can I work on it a bit later in a separate PR? Been a bit busy these days

Of course!

PR #234, opened 2019-03-01
Provide an accessible `git askyesno` and use is in `git update-git-for-windows`

dscho:
It was reported in https://github.com/git-for-windows/git/issues/2099 that our updater is not accessible, but what is *actually* not accessible is that Tcl/Tk script we use to ask the question.

This is (sadly!) the first time I was made aware that this script, that we introduced a long, long time ago, makes it hard/impossible to use with screenreaders.

Let's fix that.


dscho:
@sukiletxe please do have a look.

Here is a build of the new executable:
[git-askyesno.zip](https://github.com/git-for-windows/build-extra/files/2919863/git-askyesno.zip)

Just like `git gui--askyesno`, it has the command-line usage `git askyesno [--title <title>] <question>`, and it returns exit code 0 upon Yes and 1 upon No.

Can you please test this and verify that it is accessible?

rimrul:
This is starting to be a thing that I'm working on preparing a PR and you're slightly faster, isn't it? Well, guess I'll review it.

dscho:
@rimrul sorry!!! ;-)

git-extra/git-askyesno.c
```diff
@@ -0,0 +1,23 @@
...
+	wchar_t *title = L"Question?";
+
+	if (argc > 2 && !wcscmp(L"--title", wargv[1])) {
+		title = wargv[2];
+		argc -=2;
+		wargv += 2;
+	}
+
+	if (argc != 2) {
+		MessageBoxW(NULL, L"Required: exactly one argument", L"Error!", MB_OK);
```
rimrul: `git gui--askyesno` wrote it's error message to `stderr` and had a different wording, but this should work, too. Could be confusing for end users, though, if they get a messagebox with an obscure error due to a script calling this in a bogus way.

git-extra/git-askyesno.c
```diff
@@ -0,0 +1,23 @@
...
+		return 1;
+	}
+
```
rimrul: You've dropped the part where the TCL script concatenated all arguments (after the title). Shouldn't matter too much.

rimrul: I'd have done something like this:

````C
char *prompt=NULL;
//...
for(int j = i; j < argc; j++)
    len += strlen(argv[j])+1;
prompt = (char*)malloc(len);
if(prompt != NULL)
{
      strcpy(prompt,argv[i]);
      for(int j = i+1; j < argc; j++)
      {
           strcat(prompt," ");
           strcat(prompt, argv[j]);
      }
//messagepox here
free(prompt);
}
else
    return 1;
````

though slightly modified for wchar_t instead of char.

**Comment by rimrul** [comment]
You had better equipment this time round anyways. I couln't have compiled the WinAPI parts atm.

File: git-extra/git-askyesno.c (hunk 1 above)
Comment (dscho) [reply git-extra/git-askyesno.c]: Yeah, I never liked that behavior. The cases where `git askyesno` is to be called are when we do *not* have a console to interact with the user, after all. So `git gui--askyesno` would have written the message to essentially `/dev/null` and returned `1` as if the user had said "No". But the user did not even see a flicker.

File: git-extra/git-askyesno.c (hunk 2 above)
Comment (dscho) [reply git-extra/git-askyesno.c]: Good point. Let me copy/butcher this.

**Comment by rimrul** [comment]
Maybe the proper approach would be a third return code for errors. Whoever calls `git gui--askyesno` should be checking the result in some way, but I guess that this might break callers that check if the result is !=1 or maybe do some conversion to a boolean value with it that doesn't expect a third value. Odd initial design chioce to use the same return value for errors and no.


-------- Originalnachricht --------
Betreff: Re: [git-for-windows/build-extra] [DO NOT MERGE YET] Provide and use an accessible `git askyesno` (#234)
Von: Johannes Schindelin
An: git-for-windows/build-extra
Cc: Matthias Asshauer ,Mention



@dscho commented on this pull request.

________________________________

In git-extra/git-askyesno.c<https://github.com/git-for-windows/build-extra/pull/234#discussion_r261654669>:

> +#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+
+int wmain(int argc, wchar_t **wargv)
+{
+       int res;
+       wchar_t *title = L"Question?";
+
+       if (argc > 2 && !wcscmp(L"--title", wargv[1])) {
+               title = wargv[2];
+               argc -=2;
+               wargv += 2;
+       }
+
+       if (argc != 2) {
+               MessageBoxW(NULL, L"Required: exactly one argument", L"Error!", MB_OK);


Yeah, I never liked that behavior. The cases where git askyesno is to be called are when we do not have a console to interact with the user, after all. So git gui--askyesno would have written the message to essentially /dev/null and returned 1 as if the user had said "No". But the user did not even see a flicker.

File: git-extra/git-askyesno.c (hunk 1 above)
Comment (rimrul) [reply git-extra/git-askyesno.c]: Maybe the proper approach would be a third return code for errors. Whoever calls `git gui--askyesno` should be checking the result in some way, but I guess that this might break callers that check if the result is !=1 or maybe do some conversion to a boolean value with it that doesn't expect a third value. Odd initial design chioce to use the same return value for errors and no.

**Comment by sukiletxe** [comment]
I confirm it works correctly. Thank you very much!

**Comment by sukiletxe** [comment]
Though I believe `-h` was previously supported, and now it isn't, but that's minor of course.

**Comment by dscho** [comment]
> Though I believe `-h` was previously supported, and now it isn't, but that's minor of course.

You mean in `git gui--askyesno`? No, that was not supported.

This PR won't change what `git update-git-for-windows -h` shows, that will still work.

**Comment by dscho** [comment]
> I confirm it works correctly.

@sukiletxe thank you!

**Comment by dscho** [comment]
> Maybe the proper approach would be a third return code for errors

Indeed. If it wasn't for existing callers. We ship `git gui--askyesno` for *years*, and I really feel bad to find out that it is not accessible, only now. I want it to be a no-brainer for every user to switch to `git askyesno`.

File: git-extra/git-askyesno.c (hunk 2 above)
Comment (dscho) [reply git-extra/git-askyesno.c]: Butchered and force-pushed. Could you have a look?

File: git-extra/git-askyesno.c [hunk 3]
```diff
@@ -4,20 +4,43 @@
...
+		if (!prompt) {
+			MessageBoxW(NULL, L"Out of memory asking a question", L"Error!", MB_OK);
+			return 1;
+		}
+
+		wcscpy(prompt, wargv[1]);
+		count = wcslen(wargv[1]);
+
+		for (i = 2; i < argc; i++) {
+			prompt[count++] = L' ';
```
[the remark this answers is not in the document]
rimrul: Ok, I've thought about it a little more, postfix is correct. Sorry for that noise.

git-extra/git-askyesno.c
```diff
@@ -4,20 +4,43 @@
...
 
-	res = MessageBoxW(NULL, wargv[1], title, MB_YESNO);
+	if (argc > 2) {
+		size_t count = wcslen(wargv[1]), i;
 
+		for (i = 2; i < argc; i++)
+			count += 1 + wcslen(wargv[i]);
+
+		prompt = malloc((count + 1) * sizeof(*prompt));
+		if (!prompt) {
```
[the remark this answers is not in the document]
rimrul: Ok, I think I'm also missreading the C standard, "null pointer constant" seems to mean "some kind of zero value". This should probabbly be fine with all modern-ish compilers.

git-extra/git-askyesno.c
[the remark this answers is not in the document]
dscho: No worries!

git-extra/git-askyesno.c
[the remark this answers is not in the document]
dscho: Git's source code uses `if (!ptr)` checks all the time. If `NULL` were not equivalent to `0`, I think we would have a lot of very serious problems, and `git askyesno` would be our smallest...

crackwitz:
since "Toast" is only available on Win10 (maybe earlier, certainly not Win7), maybe this problem will "go away" when people stop using Win7 and Win8/8.1.

I'm saying to not use MessageBoxW but create a custom window/form and control its creation flags. that doesn't require C#, only plain old winapi in C. it is more work though.

PR #235, opened 2019-03-01
installer: Fix version of installer and installed file

r1walz:
As reported by @JJClements, the File version and Product version of the installer and installed binary were not the same. They are set by the `installer/release.sh` and by using `sdk build (installer | git-and-installer)` commands, which created this discrepancy.

Closes: git-for-windows/git#1797

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -38,11 +38,10 @@ OutputDir={#GetEnv('USERPROFILE')}
 #endif
 #endif
 SolidCompression=yes
-#ifdef SOURCE_DIR
-SourceDir={#SOURCE_DIR}
-#else
-SourceDir={#SourcePath}\..\..\..\..
+#ifndef SOURCE_DIR
+#define SOURCE_DIR SourcePath+'\..\..\..\..'
 #endif
+SourceDir={#SOURCE_DIR}
 #if BITNESS=='64'
 ArchitecturesInstallIn64BitMode=x64
 #endif
@@ -50,13 +49,17 @@ ArchitecturesInstallIn64BitMode=x64
 SignTool=signtool
 #endif
 
+#define FILE_VERSION GetFileVersion(SOURCE_DIR+'\'+MINGW_BITNESS+'\bin\git.exe')
+#define PROD_VERSION GetStringFileInfo(SOURCE_DIR+'\'+MINGW_BITNESS \
+				+'\bin\git.exe', 'ProductVersion')
+
 ; Installer-related
 AllowNoIcons=yes
 AppName={#APP_NAME}
 AppPublisher=The Git Development Community
 AppPublisherURL={#APP_URL}
 AppSupportURL={#APP_CONTACT_URL}
-AppVersion={#APP_VERSION}
+AppVersion={#PROD_VERSION}
 ChangesAssociations=yes
 ChangesEnvironment=yes
 CloseApplications=no
@@ -73,11 +76,7 @@ PrivilegesRequired=none
 #endif
 UninstallDisplayIcon={app}\{#MINGW_BITNESS}\share\git\git-for-windows.ico
 #ifndef COMPILE_FROM_IDE
-#if Pos('-',APP_VERSION)>0
-VersionInfoVersion={#Copy(APP_VERSION,1,Pos('-',APP_VERSION)-1)}
-#else
-VersionInfoVersion={#APP_VERSION}
-#endif
+VersionInfoVersion={#FILE_VERSION}
 #endif
 
 ; Cosmetic


installer/install.iss
```diff
@@ -74,7 +74,7 @@ PrivilegesRequired=none
 UninstallDisplayIcon={app}\{#MINGW_BITNESS}\share\git\git-for-windows.ico
 #ifndef COMPILE_FROM_IDE
 #if Pos('-',APP_VERSION)>0
-VersionInfoVersion={#Copy(APP_VERSION,1,Pos('-',APP_VERSION)-1)}
+VersionInfoVersion=GetFileVersion('{#MINGW_BITNESS}\bin\git.exe')
```
dscho: Are you sure that this is executed? I could imagine that the `APP_VERSION` constant is defined somehow without a dash...

r1walz: Yes. But `APP_VERSION` is defined through `installer/release.sh`, right? This was just to make you remember what you said (which I think is not the correct approach). According to your word [that day](https://colabti.org/irclogger/irclogger_log/git-devel?date=2019-02-20#l61), you are actually changing the version of the installer, according to what (binary) is provided with git-sdk.

dscho: Well, the wish was for the file version in the properties of the installer executable to match the file version in the properties of the included `git.exe`, and I kind of agree...

And yes, `APP_VERSION` is defined in `config.iss` that is written by `installer/release.sh`. But is it not also used for the installer's window title? IOW `APP_VERSION` is used, but we may not want to use it for the file version of the `.exe`.

If you change this so that `VersionInfoVersion` is defined independently from `APP_VERSION`, but using `GetFileVersion()` instead, does it work?

r1walz: > But is it not also used for the installer's window title?

Yes, ofc, it is used for window's title and other stuff (regedit). But won't it be a bad **UX** if user see different versions, I mean, on title bar and installer/installed bin? What was the intended use of `APP_VERSION` again? Can't we change the version of file while packing?

> If you change this so that `VersionInfoVersion` is defined independently from `APP_VERSION`, but using `GetFileVersion()` instead, does it work?

I can make `GetFileVersion()` work, but again, it will face the formentioned problem. Also how can we generate `git-sdk` using `git-sdk`? (basically mingw64\bin\git.exe)

dscho: The file version is limited to four numbers separated by dots, IIUC, the version displayed in the window title is free form. So it can be a lot more expressive.

r1walz: @dscho my main concern here is **UX**, I mean if we are showing some different version on the title bar and installing some other version then this is bad(I know that we'll not be shipping such installer but yeah :P).

I'm making it to change the `VersionInfoVersion` to match the binary included and I guess we shall not be needing the conditional here because then we will get the same problematic installer when using `sdk build (installer|git-and-installer)` or `installer/release.sh <version-string>`?

https://github.com/git-for-windows/build-extra/blob/bbe200d9a0faace5249d912857a3968f90554b03/installer/install.iss#L76-L80

dscho: Sounds good.

r1walz:
@dscho can you please review this?

installer/install.iss
```diff
@@ -40,8 +40,15 @@ OutputDir={#GetEnv('USERPROFILE')}
...
 SourceDir={#SOURCE_DIR}
+#define FILE_VERSION GetFileVersion(SOURCE_DIR+"\"+MINGW_BITNESS+"\bin\git.exe")
+#define PROD_VERSION GetStringFileInfo(SOURCE_DIR+"\"+MINGW_BITNESS+"\bin\git.exe", \
+			"ProductVersion")
 #else
 SourceDir={#SourcePath}\..\..\..\..
+#define FILE_VERSION GetFileVersion(SourcePath+"\..\..\..\..\" \
+			+MINGW_BITNESS+"\bin\git.exe")
+#define PROD_VERSION GetStringFileInfo(SourcePath+"\..\..\..\..\" \
+			+MINGW_BITNESS+"\bin\git.exe", "ProductVersion")
```
dscho: Why not use `ExpandConstant('{#SourceDir}')` (or something similar to make use of `SourceDir`) for both arms, making them identical, i.e. you can move the definition of `FILE_VERSION` and `PROD_VERSION` out of the conditional?

r1walz:
> Could we simplify this further? Have you tested and verified that it works?

I'll see if we can simplify it further ... Yes, I've verified and it works according to my understandings.

dscho:
> I'll see if we can simplify it further

To be clear: I think there is no need to put those `#define`s of `FILE_VERSION` and `PROD_VERSION` into that `if ... else ...` because I think they should be defined identically, using the `SourceDir` value, outside that `if ... else ...`.

r1walz:
@dscho I tried changing the code, but it gave error `undeclared identifier ExpandConstant`, any views?

dscho:
>  I tried changing the code, but it gave error `undeclared identifier ExpandConstant`

As discussed [on IRC](https://colabti.org/irclogger/irclogger_log/git-devel?date=2019-03-06#l21), we'll try to use `#define SOURCE_DIR` instead, as it is apparently impossible to use an InnoSetup constant to define a pre-processor constant.

r1walz:
Yes, I've tested it, and it works!

r1walz:
Yes, I agree [this](https://github.com/git-for-windows/build-extra/pull/235#commitcomment-32614111) is prettier.

dscho:
Excellent, thank you so much!

PR #238, opened 2019-03-08
Make it a lot easier to start hacking on the installer

dscho:
One of the central missions of Git for Windows' SDK is to make it as easy as humanly possible for Git for Windows users to work on their own feature requests. Yes, it is a complex system, but that does not mean that everybody who wants to contribute as little as an enhanced description of an installer options needs to understand all that.

With this change, we can start to recommend to users to "just download the SDK and then call `sdk edit install.iss` followed by `sdk build installer`".

This PR also brings a couple of other touch-ups to the `sdk` function, more or less because I either needed them to develop this PR (such as `sdk reload --experimental` and `sdk edit git-sdk.sh`) or because I just saw those issues nearby and thought I'd just quickly fix them.

### DIFF
--- a/git-extra/git-sdk.sh
+++ b/git-extra/git-sdk.sh
@@ -89,7 +89,7 @@ sdk () {
 		;;
 	valid_projects)
 		printf "%s " build-extra git git-extra MINGW-packages \
-			MSYS2-packages msys2-runtime
+			MSYS2-packages msys2-runtime installer
 		;;
 	valid_build_targets)
 		printf "%s " git-and-installer $(sdk valid_projects | tr ' ' '\n' |
@@ -109,7 +109,7 @@ sdk () {
 				https://github.com/git-for-windows/"$2" ||
 			sdk die "Could not initialize $src_dir"
 			;;
-		git-extra)
+		git-extra|installer)
 			sdk init-lazy build-extra &&
 			src_dir="$src_dir/$2" ||
 			return 1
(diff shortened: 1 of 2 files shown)


git-extra/git-sdk.sh
```diff
@@ -87,6 +87,10 @@ sdk () {
 		printf "%s " build-extra git git-extra MINGW-packages \
 			MSYS2-packages msys2-runtime
 		;;
+	valid_build_targets)
+		printf "%s " git-and-installer $(sdk valid_projects | tr ' ' '\n' |
+			grep -v '^\(build-extra\|\(MINGW\|MSYS2\)-packages\)')
```
bbolli: Maybe use `grep -E` here?

git-extra/git-sdk.sh
```diff
@@ -79,6 +80,9 @@ sdk () {
 		echo "$*" >&2
 		return 1
 		;;
+	fmt_list)
+		fmt -w 64 | sed 's/^/\t/'
```
bbolli: 👍

git-extra/git-sdk.sh
dscho: I try to avoid non-standard options; it is a bad habit to get into when you have to contribute patches to Git's test suite (where we really need portability)...

dscho:
Thank you, @bbolli, for reviewing!

git-extra/git-sdk.sh
bbolli: But... would a `grep` that doesn't understand `-E` grok the extended syntax, even if the metacharacters are quoted? 🤔

PhilipOakley:
Throw away comment: Is there a place that we should also patch for Documentation regarding the SDK operation?
(have the various tidbits got spread too wide such as the wiki and other places..).
Mainly, where should the canonical location for the sdk documentation be?

dscho:
>  Mainly, where should the canonical location for the sdk documentation be?

How about `sdk help`?

git-extra/git-sdk.sh
dscho: I think so: https://www.regular-expressions.info/posix.html#bre

dscho: I think so: https://www.regular-expressions.info/posix.html#bre

PR #241, opened 2019-03-10
remove credential.helper manager for PortableGit

ousia:
might fix @git-for-windows/git#2116

Signed-off-by: Pablo Rodriguez <ousia@users.noreply.github.com>

### DIFF
--- a/portable/release.sh
+++ b/portable/release.sh
@@ -89,10 +89,7 @@ die "Could not copy post-install script"
 
 mkdir -p "$SCRIPT_PATH/root/mingw$BITNESS/etc" &&
 cp /mingw$BITNESS/etc/gitconfig \
-	"$SCRIPT_PATH/root/mingw$BITNESS/etc/gitconfig" &&
-git config -f "$SCRIPT_PATH/root/mingw$BITNESS/etc/gitconfig" \
-	credential.helper manager ||
-die "Could not configure Git-Credential-Manager as default"
+	"$SCRIPT_PATH/root/mingw$BITNESS/etc/gitconfig"
 test 64 != $BITNESS ||
 git config -f "$SCRIPT_PATH/root/mingw$BITNESS/etc/gitconfig" --unset pack.packSizeLimit
 


portable/release.sh
```diff
@@ -89,10 +89,7 @@ die "Could not copy post-install script"
 
 mkdir -p "$SCRIPT_PATH/root/mingw$BITNESS/etc" &&
 cp /mingw$BITNESS/etc/gitconfig \
-	"$SCRIPT_PATH/root/mingw$BITNESS/etc/gitconfig" &&
-git config -f "$SCRIPT_PATH/root/mingw$BITNESS/etc/gitconfig" \
-	credential.helper manager ||
-die "Could not configure Git-Credential-Manager as default"
+	"$SCRIPT_PATH/root/mingw$BITNESS/etc/gitconfig"
```
dscho: This breaks the `&&` chain. Please reinstate it.

dscho:
Thank you for starting this conversation with a patch.

I see that this tries to revert https://github.com/git-for-windows/build-extra/commit/7329c72a2204320c77053d58361fb295839092fb, although incompletely.

Unfortunately, the commit shares one really big problem with that original commit: neither commit message does an even adequate job at justifying the change.

Now, please note that I criticize *my own* commit, too. From the top of my head, I seem to remember that it was an explicit user request that triggered that change, and I regret not recording that rationale in the commit message, two years ago.

I *seem* to remember that the use case was a Git for Windows installation by a user who does not have admin privileges, i.e. a user-only installation. It is still my impression that that is by far the most common scenario where the portable Git is used.

Having said that, I have sympathy for the idea behind *this* patch, and I hope we can come up with a solution for both scenarios.

What I have in mind is not to change the current default, but offer a one-time warning the first time the portable Git is started, with a suggestion how to disable the Git Credential Manager, if so desired.

Alternatively, we could implement a meta credential helper and configure *that* by default: the first time Git asks for credentials, it would pop up and list all available credential helpers, lettingethe user choose which one to use (and offering to optionally configure that helper in the portable Git's system config to avoid calling the meta helper again).

That meta helper would most likely be the far superior solution, it just would require a little bit more work.

ousia:
@dscho, I’m afraid that this is beyond my knowledge. I cannot code (my background is in humanities, not even in science) and I don’t even have the time to learn.

I don’t  know why the `&&` chain breaks with my path. It makes sense to me. I know I must be wrong, but it is out of ignorance.

Sorry, but I reported the issue not to have the problem solved for me (which was solved hours before, in the computer where it took place), but to fix what could lead to a non–minor security issue.

I respect and I really appreciate all the hard work behind _Git_ and _PortableGit_ for _Windows_. But having to learn coding in _Bash_ to submit a patch for an issue I reported is something I cannot afford right now.

dscho:
> I don’t know why the `&&` chain breaks with my path.

In shell scripting, we often use the idiom `<this> && <that> && <also> || <error>` to execute `<that>` only if `<this>` succeeded, and `<also>` only if `<that>` succeeded, and `<error>` only if any of the three failed.

Your patch removes the `&&` from the end of the line, which breaks such a chain.

> I’m afraid that this is beyond my knowledge. I cannot code (my background is in humanities, not even in science) and I don’t even have the time to learn.

That's fine. Maybe some other volunteer can step in?

dscho:
Okay, I made up my mind, we should close this PR and continue the discussion over at git-for-windows/git#2116

dscho:
Okay, I made up my mind, we should close this PR and continue the discussion over at git-for-windows/git#2116

PR #242, opened 2019-03-13
avoid Github API rate limit

drizzd:
The Github API has a per-user rate limit. For anonymous users, which we
are, the limit counts per IP address. For large organizations this can
be an issue. Instead, make the latest version check against
gitforwindows.org.

We still make calls to the Github API when we decide that we should
download a new release. But this happens only once per user and Git
release.

Signed-off-by: Clemens Buchacher <drizzd@gmx.net>
See also https://github.com/git-for-windows/git/issues/1813.

### DIFF
--- a/git-extra/git-update-git-for-windows
+++ b/git-extra/git-update-git-for-windows
@@ -162,15 +162,15 @@ update_git_for_windows () {
 		;;
 	esac
 
-	releases_url=https://api.github.com/repos/git-for-windows/git/releases
-	releases=$(http_get $releases_url/latest) ||
+	latest_tag_url=https://gitforwindows.org/latest-tag.txt
+	latest_tag=$(http_get $latest_tag_url) ||
 	case $?,"$proxy" in
 	7,)
-		proxy="$(proxy-lookup.exe https://api.github.com)" &&
+		proxy="$(proxy-lookup.exe https://gitforwindows.org)" &&
 		test -n "$proxy" &&
 		export https_proxy="$proxy" &&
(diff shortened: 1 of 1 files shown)


git-extra/git-update-git-for-windows
```diff
@@ -162,25 +162,23 @@ update_git_for_windows () {
...
 		;;
 	*)
 		return
 		;;
 	esac
 
-	latest=$(echo "$releases" |
-		grep '"tag_name": "v' |
-		sed -E 's/.*"tag_name": "v([^"]*).*/\1/')
+	latest=$(echo "$latest_tag" | sed s/^v//)
```
dscho: How about `latest=${latest_tag#v}`? That should be a lot quicker, as it avoids spawning a new process.

git-extra/git-update-git-for-windows
```diff
@@ -205,6 +203,8 @@ update_git_for_windows () {
 	esac
 
 	echo "Update $latest is available" >&2
+	releases_url=https://api.github.com/repos/git-for-windows/git/releases
+	releases=$(http_get $releases_url/latest) || return
```
dscho: I guess we do not really need to take care of trying a different proxy for this URL, as `gitforwindows.org` is backed by GitHub Pages?

drizzd: Right. I could not think of a use case where gitforwindows.com should use a different proxy than api.github.com. And refactoring the code to lookup a new proxy each time seems premature.

dscho:
Thanks!

PR #244, opened 2019-03-22
git-extra: offer an easy way to update an SDK via Pacman

dscho:
This PR adds the Batch file `update-via-pacman.bat` to the top-level directory of any Git for Windows SDK. This script will make it easier, hopefully, to update a local copy of the SDK.

cc @SteveALee

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -48,7 +48,8 @@ source=('inputrc'
         'blocked-file-util.c'
         'zzz_rm_stackdumps.sh'
         'proxy-lookup.c'
-        'git-askyesno.c')
+        'git-askyesno.c'
+        'update-via-pacman.bat')
 sha256sums=('9efaf8dccc08c7cddc58cb589bab5aac5c0894661175a344ca02b2aa849382bd'
             'c26d22aaa1d3dc615d474e0d60c5c0f19598d61d9205e19ec87aac1b28bb07c1'
             '640d04d2a2da709419188a986d5e5550ad30df23c7ea9be1a389c37a6b917994'
@@ -130,5 +131,6 @@ package() {
   install -m755 astextplain $pkgdir/usr/bin
   install -m755 git-update-git-for-windows $pkgdir${MINGW_PREFIX}/libexec/git-core
   install -m755 git-update $pkgdir${MINGW_PREFIX}/libexec/git-core
+  install -m755 update-via-pacman.bat $pkgdir
   install -m755 zzz_rm_stackdumps.sh $pkgdir/usr/share/makepkg/lint_package
 }
--- a/git-extra/update-via-pacman.bat
+++ b/git-extra/update-via-pacman.bat
@@ -1,0 +1,94 @@
+@REM This script updates a Git for Windows SDK via Pacman.
+@REM
+@REM Note: an alternative would be to clone the SDK from the Git mirrors at
+@REM https://github.com/git-for-windows/git-sdk-64 or
+@REM https://github.com/git-for-windows/git-sdk-32 and then update it via
+@REM `git pull` (but that must be done outside of the Git SDK, e.g. via a
+@REM regular Git for Windows Bash that was installed using the official Git for
+@REM Windows installer).
+@REM
+@REM At this time, it is not recommended to mix both methods. Either use Git,
+@REM or use Pacman, but not both, to update your Git for Windows SDK.
+
+@REM Enable extensions
+@REM the `verify other 2>nul` is a trick from the setlocal help
+@VERIFY other 2>nul
+@SETLOCAL enableDelayedExpansion
+
+@IF ERRORLEVEL 1 (
+    @ECHO Unable to enable delayed expansion!
+    @EXIT /B 1
+)
+
+@REM Get the absolute path to the current directory (i.e. the SDK root)
+@FOR /F "delims=" %%I IN ("%~dp0..") do @set git_sdk_root=%%~fI
+@SET PATH=%~dp0\usr\bin;%PATH%
+
+@REM Set to MSYS mode
+@SET MSYSTEM=MSYS
+
+@ECHO Run Pacman
+@pacman -Syyu --noconfirm
+@IF ERRORLEVEL 1 GOTO DIE
+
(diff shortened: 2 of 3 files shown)


PhilipOakley:
I had a read through and didn't see anything untoward, but then again I haven't actually tried it.

The one other point is "How to bootstrap this for existing users?" - The commit message, or PR message, should have some small note to the 'less gifted' SDK users, who already have the older SDK installed, as to how to pick up the bat file into their own setup.  

Once the PR is accepted new users should be just fine...

SteveALee:
Looks good and works for me. It successfully updated the core and re ran to do rest.

A minor niggleif you run again and packman doesn't run the missing log seems to trip it up

```
 there is nothing to do
/c/Users/steve/.bashrc: line 29: cd: /c/projects/aim-web-app: No such file or directory
tail: cannot open 'var/log/pacman.log' for reading: No such file or directory
"Run Pacman again to upgrade the remaining (non-core) packages"
```

SteveALee:
PS I can update the wiki page if you like when this is merged?

dscho:
> /c/Users/steve/.bashrc: line 29: cd: /c/projects/aim-web-app: No such file or directory

What is in line 29 of that file?

dscho:
> The one other point is "How to bootstrap this for existing users?" - The commit message, or PR message, should have some small note to the 'less gifted' SDK users, who already have the older SDK installed, as to how to pick up the bat file into their own setup.

@PhilipOakley I added this paragraph to the commit message:

> Existing users can copy the `update-via-pacman.bat` file directly into their SDK's top-level directory, or get it via `pacman -Sy git-extra` once https://github.com/git-for-windows/build-extra/pull/244 is merged.

dscho:
> A minor niggleif you run again and pacman doesn't do anyhting the missing log seems to trip it up

@SteveALee you're right! I forgot two things:

- I forgot to use the absolute path for `/var/log/pacman.log` (in an earlier version, the script `cd`ed into the top-level directory of the SDK, but I figured that would no longer be necessary, so I removed it).
- You pointed out that there is a good chance that there is "nothing to do", in which case no needle is found. I tried to fix that, but have currently no good way to test this. Could I bother you with testing this again?

I force-pushed to address these two issues.

> PS I can update the wiki page if you like when this is merged?

That would be great, thank you!

SteveALee:
@dscho 

>>    /c/Users/steve/.bashrc: line 29: cd: /c/projects/aim-web-app: No such file or directory

> What is in line 29 of that file?

Ah, sorry - simply a `cd` to the directory that no longer exists - I was lazy :)

SteveALee:
> You pointed out that there is a good chance that there is "nothing to do", in which case no needle is found. I tried to fix that, but have currently no good way to test this. Could I bother you with testing this again?

Not quite there yet. Here's the output

```
warning: msys2-runtime: ignoring package upgrade (2.11.2-1 => 3.0.4-1)
warning: msys2-runtime-devel: ignoring package upgrade (2.11.2-1 => 3.0.4-1)
 there is nothing to do
:: Starting full system upgrade...
warning: grep: local (3.1-1) is newer than msys (3.0-2)
warning: libgc: local (8.0.0-1) is newer than msys (7.6.8-1)
warning: libxml2: ignoring package upgrade (2.9.8-1 => 2.9.9-2)
warning: mingw-w64-i686-binutils: local (2.31.1-2) is newer than mingw32 (2.30-5)
warning: mingw-w64-x86_64-binutils: local (2.31.1-2) is newer than mingw64 (2.30-5)
warning: mingw-w64-x86_64-gcc: ignoring package upgrade (7.3.0-2 => 8.3.0-2)
warning: mingw-w64-x86_64-gcc-libs: ignoring package upgrade (7.3.0-2 => 8.3.0-2)
warning: mingw-w64-x86_64-osslsigncode: ignoring package upgrade (1.7.1-3 => 1.7.1-4)
 there is nothing to do
```

So that looks better - I assume those messages are due to the temp G4W version issue you mentioned elsewhere?

Actually, why does it say "nothong to do" when you use `-Syy` which I understand always refreshes? 
 
```
tail: cannot open '/var/log/pacman.log' for reading: No such file or directory
```

Hmmm. :(

```
warning: git-extra-1.1.359.febab20-1 is up to date -- reinstalling
```

What is you expecteation for where we run this script? The root of the SDK seems reasonable - but perhasp add a usage to the comments or script output?.

I was running in a cmd shell as it's a .bat (.cmd).

I just found out that I CAN run it in the SDK bash shell. I assumed I would not as autocomplete fails for .bat (that is tab after typing a few chars like `./aut` ) But it's fine if will run if you type in the full name (can we fix that?).

dscho:
> So that looks better - I assume those messages are due to the temp G4W version issue you mentioned elsewhere?

Those messages are due to some packages having been available in MSYS2 at newer versions, but apparently no longer. No idea what's the story behind that.

> why does it say "nothong to do" when you use `-Syy` which I understand always refreshes?

`pacman -Syy` only makes sure that the databases (or package indices) are downloaded, even if `pacman` thinks that the local copies are up to date.

"Nothing to do" means that we have the newest versions of the installed packages.

So that is quite good!

> ```
> tail: cannot open '/var/log/pacman.log' for reading: No such file or directory
> ```
> 
> 
> Hmmm. :(

Can you verify that this file is indeed absent?

> What is you expectation for where we run this script? The root of the SDK seems reasonable - but perhasp add a usage to the comments or script output?.

It should not matter. `pacman` is supposed to take care of changing the working directory as needed.

> I was running in a cmd shell as it's a .bat (.cmd).

That's the right way to do it.

> I just found out that I CAN run it in the SDK bash shell.

But please don't! The entire idea of making this a `.bat` script is so that you do not have to run anything MSYS2-related (like the Git Bash), which could prevent a core update.

You see, `pacman` may overwrite `/usr/bin/msys-2.0.dll` with a newer version. That is specifically intended to work.

However, after that, the MSYS2 runtime has a problem with the tear-down of its POSIX emulation layer. Meaning: *any* MSYS2 program. Like `pacman`. And like `bash`!

Therefore: best not to call `pacman` from Bash.

dscho:
I guess this PR is as good as it will ever get. Merging time!

PR #245, opened 2019-03-23
Initial attempt at a portable MSYS2

SteveALee:
Lots of questions - inline comments in PR and `TODO` lines in files.

Main question which packages to include as currently it's bigget than the G4W portable

Also need to be clear why this is better than just using [MSYS2](https://www.msys2.org/) - I think MinGW packages are used where possible so better performance. Is that right? What else?


make-file-list-msys2.sh
```diff
@@ -0,0 +1,128 @@
...
+	if test -n "$PACKAGE_VERSIONS_FILE"
+	then
+		pacman -Q $package_list >"$PACKAGE_VERSIONS_FILE"
+	fi &&
+	pacman -Ql $package_list |
+	grep -v '/$' |
+	sed 's/^[^ ]* //'
+}
+
+install_required () {
```
SteveALee: I didn't include the stuff about include despite upgrade as guessed it wouldn't be needed

make-file-list-msys2.sh
```diff
@@ -0,0 +1,128 @@
...
+}
+
+this_script_dir="$(cd "$(dirname "$0")" && pwd -W)" ||
+die "Could not determine this script's dir"
+
+install_required
+
+SH_FOR_REBASE=dash
+
+BASE_PACKAGES="$(pacman -Qg base | awk '{print $2}' | tr '\n' ' ')"
```
SteveALee: Peobably don;t need all these or and removed via PACKAGE_EXCLUDES

make-file-list-msys2.sh
```diff
@@ -0,0 +1,128 @@
...
+GIT_PACKAGES="mingw-w64-$ARCH-git mingw-w64-$ARCH-git-credential-manager git-extra openssh"
+UTIL_PACKAGES=
+packages="$BASE_PACKAGES $GIT_PACKAGES $UTIL_PACKAGES"
+
+# TODO need this? should be a parameter? Use BITNESS? 
+PACKAGE_EXCLUDES="db info heimdal git util-linux curl git-for-windows-keyring
+	mingw-w64-p11-kit filesystem msys2-launcher-git rebase"
+# TODO 05-home-dir.post required by filesystem package but is missing from SDK
```
SteveALee: For some reason this file wanted but I was missing it in my SDK

make-file-list-msys2.sh
```diff
@@ -0,0 +1,128 @@
...
+# TODO should we exlude all the tz stuff?
```
SteveALee: Lots of stuff like perl (needed or git) Local and TZ files really bloat the zip but I don;t kno when might be needed

make-file-list-msys2.sh
```diff
@@ -0,0 +1,128 @@
...
+EXTRA_FILE_EXCLUDES="/etc/post-install/05-home-dir.post
+	/mingw$BITNESS/libexec/git-core/git-update-git-for-windows"
+
+pacman_list $packages "$@" |
+grep -v -e '\.[acho]$' -e '\.l[ao]$' -e '/aclocal/' \
```
SteveALee: I'm not clear if these filters are actually useful given my almost completely different package list

make-file-list-msys2.sh
```diff
@@ -0,0 +1,128 @@
...
+	-e "^\\($(echo $EXTRA_FILE_EXCLUDES |
+		 sed 's/ /\\|/g')\\)\$" |
+sort |
+grep --perl-regexp -v -e '^/usr/(lib|share)/terminfo/(?!.*/(cygwin|dumb|screen.*|xterm.*)$)' |
+sed 's/^\///'
+
+# TODO - need this?
+test -z "$PACKAGE_VERSIONS_FILE" ||
+pacman -Q filesystem $SH_FOR_REBASE rebase \
+	util-linux unzip mingw-w64-$ARCH-xpdf-tools \
```
SteveALee: should w64 use BITNESS?

portable-msys2/release.sh
```diff
@@ -0,0 +1,152 @@
...
+	grep -v "^mingw$BITNESS/etc/gitconfig$")" ||
+die "Could not generate file list"
+
+rm -rf "$SCRIPT_PATH/root/mingw$BITNESS/libexec/git-core" &&
+mkdir -p "$SCRIPT_PATH/root/mingw$BITNESS/libexec/git-core" &&
+ln $(echo "$LIST" | sed -n "s|^mingw$BITNESS/bin/[^/]*\.dll$|/&|p") \
+	"$SCRIPT_PATH/root/mingw$BITNESS/libexec/git-core/" ||
+die "Could not copy .dll files into libexec/git-core/"
+
+# TODO there are few extra untacked files under /root. Can't figure how they got there
```
SteveALee: this is probably solved now after drilling down into package filelists

portable-msys2/release.sh
```diff
@@ -0,0 +1,152 @@
...
+die "Could not unpack .pdb files"
+
+# 7-Zip will strip absolute paths completely... therefore, we can add another
+# root directory like this:
+
+LIST="$LIST $SCRIPT_PATH/root/*"
+
+# Make the self-extracting package
+
+type 7za > /dev/null ||
```
SteveALee: redirect to the old bit bucket

portable-msys2/root/README.portable
```diff
@@ -0,0 +1,106 @@
+README.portable
```
SteveALee: Oops will need to rewrite this

make-file-list-msys2.sh
```diff
@@ -0,0 +1,128 @@
...
+	-e '^/mingw../share/locale/' \
+	-e '^/usr/share/locale/' |
+grep -v \
+	-e "^\\($(echo $EXTRA_FILE_EXCLUDES |
+		 sed 's/ /\\|/g')\\)\$" |
+sort |
+grep --perl-regexp -v -e '^/usr/(lib|share)/terminfo/(?!.*/(cygwin|dumb|screen.*|xterm.*)$)' |
+sed 's/^\///'
+
+# TODO - need this?
```
SteveALee: I'm not at all sure why this is wanted

And shouldn't rebase be being run on 32 bit version?

make-file-list-msys2.sh
```diff
@@ -0,0 +1,128 @@
...
+	-e '^/usr/share/perl5/core_perl/pods/' \
+	-e '^/usr/share/perl5/core_perl/Locale/' \
+	-e '^/usr/share/perl5/core_perl/Pod/' \
+	-e '^/usr/share/perl5/core_perl/ExtUtils/' \
+	-e '^/usr/share/perl5/core_perl/CPAN/' \
+	-e '^/usr/share/perl5/core_perl/TAP/' \
+	-e '^/usr/share/vim/vim74/lang/' \
+	-e '^/etc/profile.d/git-sdk.sh$' |
+grep -v \
+	-e '^/mingw../share/locale/' \
```
SteveALee: can prolly ditch TZ stuff too

portable-msys2/root/etc/post-install/release-bash.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/sh
```
SteveALee: hmm - this shouldn't be here I guess. Not sure how it happened

portable-msys2/release.sh
```diff
@@ -0,0 +1,152 @@
...
+ echo 'BeginPrompt="This archive extracts a minimal MSYS2 for Windows '$BITNESS'-bit"' &&
+ echo 'CancelPrompt="Do you want to cancel the portable MSYS2 installation?"' &&
+ echo 'ExtractDialogText="Please, wait..."' &&
+ echo 'ExtractPathText="Where do you want to install portable MSYS2?"' &&
+ echo 'ExtractTitle="Extracting..."' &&
+ echo 'GUIFlags="8+32+64+256+4096"' &&
+ echo 'GUIMode="1"' &&
+ echo 'InstallPath="%%S\\PortableMSYS2"' &&
+ echo 'OverwriteMode="0"' &&
+ echo "RunProgram=\"git-bash.exe --needs-console --hide --no-cd --command=post-install.bat\"" &&
```
SteveALee: The existing post-install.bat looks good to me.

WSLUser:
If you like feedback on tools. Getting `make`, `gcc`, and useful terminal commands such as `more` available would be optimal.

tomalex4real1 (approved):

martin-braun:
Can we also make it portable between terminals? See https://stackoverflow.com/a/77877407/1540350

PR #246, opened 2019-04-16
Add a meta credential helper

dscho:
The idea is to have a credential helper that discovers all other credential helpers on the `PATH` and lets the user choose one (or none).

This meta helper will come in handy to avoid surprises where users run a portable Git on someone else's computer and expect no credentials to be stored there.

This addresses https://github.com/git-for-windows/git/issues/2116.

### DIFF
--- a/git-extra/Makefile
+++ b/git-extra/Makefile
@@ -1,5 +1,6 @@
 BUILDDIR ?= .
 SRCDIR ?= .
+WINDRES ?= windres
 CFLAGS ?= -Wall
 CXXFLAGS ?= -Wall
 
@@ -13,6 +14,9 @@ $(BUILDDIR)/create-shortcut.exe: $(BUILDDIR)/create-shortcut.o
 
 $(BUILDDIR)/%.o: $(SRCDIR)/%.c
 	$(CC) -c $(CFLAGS) $< -o $@
+
+$(BUILDDIR)/%.res: $(SRCDIR)/%.rc
+	$(WINDRES) --input $< --output $@ --output-format coff
 
 $(BUILDDIR)/WhoUses.exe: $(BUILDDIR)/WhoUses.o $(BUILDDIR)/SystemInfo.o
 	$(CXX) $(CXXFLAGS) -o $@ $^
@@ -37,7 +41,8 @@ $(BUILDDIR)/git-askyesno.exe: $(BUILDDIR)/git-askyesno.o
 $(BUILDDIR)/git-credential-helper-selector.o: CFLAGS += -DUNICODE
 
 $(BUILDDIR)/git-credential-helper-selector.exe: \
-	$(BUILDDIR)/git-credential-helper-selector.o
+	$(BUILDDIR)/git-credential-helper-selector.o \
+	$(BUILDDIR)/git-credential-helper-selector.res
 	$(CC) -municode $(CFLAGS) -o $@ $^ -lgdi32 -lcomctl32
 clean:
 	$(RM) $(BUILDDIR)/create-shortcut.exe $(BUILDDIR)/create-shortcut.o
--- a/git-extra/Makefile
+++ b/git-extra/Makefile
@@ -5,7 +5,8 @@ CXXFLAGS ?= -Wall
 
 all: $(BUILDDIR)/create-shortcut.exe $(BUILDDIR)/WhoUses.exe \
 	$(BUILDDIR)/blocked-file-util.exe $(BUILDDIR)/proxy-lookup.exe \
-	$(BUILDDIR)/git-askyesno.exe
+	$(BUILDDIR)/git-askyesno.exe \
+	$(BUILDDIR)/git-credential-helper-selector.exe
 
 $(BUILDDIR)/create-shortcut.exe: $(BUILDDIR)/create-shortcut.o
 	$(CC) $(CFLAGS) -o $@ $^ -luuid -lole32
@@ -33,9 +34,16 @@ $(BUILDDIR)/git-askyesno.o: CFLAGS += -DUNICODE
 $(BUILDDIR)/git-askyesno.exe: $(BUILDDIR)/git-askyesno.o
 	$(CC) -municode $(CFLAGS) -o $@ $^
 
+$(BUILDDIR)/git-credential-helper-selector.o: CFLAGS += -DUNICODE
+
+$(BUILDDIR)/git-credential-helper-selector.exe: \
+	$(BUILDDIR)/git-credential-helper-selector.o
+	$(CC) -municode $(CFLAGS) -o $@ $^ -lgdi32 -lcomctl32
 clean:
 	$(RM) $(BUILDDIR)/create-shortcut.exe $(BUILDDIR)/create-shortcut.o
 	$(RM) $(BUILDDIR)/WhoUses.exe $(BUILDDIR)/WhoUses.o \
 		$(BUILDDIR)/SystemInfo.o
 	$(RM) $(BUILDDIR)/blocked-file-util.exe $(BUILDDIR)/blocked-file-util.o
 	$(RM) $(BUILDDIR)/proxy-lookup.exe $(BUILDDIR)/proxy-lookup.o
+	$(RM) $(BUILDDIR)/git-credential-helper-selector.exe \
+		$(BUILDDIR)/git-credential-helper-selector.o
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -49,7 +49,8 @@ source=('inputrc'
         'zzz_rm_stackdumps.sh'
         'proxy-lookup.c'
         'git-askyesno.c'
-        'update-via-pacman.bat')
+        'update-via-pacman.bat'
+        'git-credential-helper-selector.c')
 sha256sums=('9efaf8dccc08c7cddc58cb589bab5aac5c0894661175a344ca02b2aa849382bd'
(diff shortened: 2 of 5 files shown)


bbolli:
Sorry for my late reply, @dscho . I don't want to belittle your efforts in any way; without them I'd never have started using git over 10 years ago.

But now it's 2019 and we're still doing manual memory management in C code. I propose to rewrite this helper in C++ where there's std::wstring and std::vector classes that do all this low-level stuff for you.

dscho:
> I propose to rewrite this helper in C++ where there's std::wstring and std::vector classes that do all this low-level stuff for you.

Hmm. That would be the first one C++ application we include in Git for Windows, I think, and I am not sure I want to deal with the dependencies...

git-extra/git-credential-chooser.c
```diff
@@ -10,20 +11,99 @@ static size_t helper_nr, selected_helper;
...
-	helper_nr = 2;
-	helper_name = malloc(helper_nr * sizeof(*helper_name));
-	helper_path = malloc(helper_nr * sizeof(*helper_path));
+	WCHAR *pattern_suffix = L"\\git-credential-*";
+	size_t alloc = 16, env_size, pattern_len, pattern_suffix_len = wcslen(pattern_suffix);
+	WCHAR env[32768], pattern[32768 + pattern_suffix_len], *p;
+	HANDLE h;
+	WIN32_FIND_DATAW find_data;
+
+	env_size = GetEnvironmentVariableW(L"PATH", env, 32768);
```
bbolli: DRY: use `sizeof env` instead of the number.

git-extra/git-credential-chooser.c
```diff
@@ -10,20 +11,99 @@ static size_t helper_nr, selected_helper;
...
+	if (!env_size || env_size > 32767) {
```
bbolli: Same.

git-extra/git-credential-chooser.c
```diff
@@ -10,20 +11,99 @@ static size_t helper_nr, selected_helper;
...
+					alloc = helper_nr + 16;
+				helper_name = realloc(helper_name, alloc * sizeof(*helper_name));
+				helper_path = realloc(helper_path, alloc * sizeof(*helper_path));
+				if (!helper_name || !helper_path)
+					goto out_of_memory;
+			}
+
+			helper_path[helper_nr] = malloc((pattern_len + 1 + len + 1) * sizeof(WCHAR));
+			if (!helper_path[helper_nr])
+				goto out_of_memory;
```
bbolli: This leaks the memory of the `helper_path`s that have already been allocated.

git-extra/git-credential-chooser.c
```diff
@@ -10,20 +11,99 @@ static size_t helper_nr, selected_helper;
...
+			memcpy(helper_path[helper_nr], pattern, pattern_len * sizeof(WCHAR));
+			helper_path[helper_nr][pattern_len] = L'\\';
+			memcpy(helper_path[helper_nr] + pattern_len + 1, find_data.cFileName, (len + 1) * sizeof(WCHAR));
+
+			helper_name[helper_nr] = malloc((len + 1 - (pattern_suffix_len - 2)) * sizeof(WCHAR));
+			if (!helper_name[helper_nr])
+				goto out_of_memory;
```
bbolli: Same.

git-extra/git-credential-chooser.c
```diff
@@ -10,20 +11,99 @@ static size_t helper_nr, selected_helper;
...
+
+next_file:
+			if (!FindNextFileW(h, &find_data))
+				break;
+		}
+
+		if (!q)
+			break;
+		p = q + 1;
+	}
```
bbolli: We need to `FindClose(h)` somewhere around here.

git-extra/git-credential-chooser.c
```diff
@@ -584,7 +585,7 @@ static LRESULT CALLBACK window_proc(HWND hwnd, UINT message, WPARAM wParam,
 			      hwnd, (HMENU) ID_PERSIST, NULL, NULL);
 
 		CreateWindowW(L"Button", L"Choose",
-			      WS_VISIBLE | WS_CHILD,
+			      WS_VISIBLE | WS_CHILD | BS_DEFPUSHBUTTON,
```
bbolli: :smiley:

git-extra/git-credential-chooser.c
```diff
@@ -9,6 +9,148 @@ static size_t helper_nr, selected_helper;
...
+	/* Skip detection for .exe, .bat, and .cmd */
+	count = wcslen(path);
+	if (count > 4 && path[count - 4] == L'.' &&
+	    (!wcscmp(path + count - 3, L"exe") ||
+	     !wcscmp(path + count - 3, L"bat") ||
+	     !wcscmp(path + count - 3, L"cmd")))
+		return NULL;
+
+	h = CreateFileW(path, GENERIC_READ, FILE_SHARE_READ, NULL,
+			       OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
```
bbolli: Wrong indent? Or Github using 8-spaced TABs and your editor 4-spaced?

git-extra/git-credential-chooser.c
```diff
@@ -9,6 +9,148 @@ static size_t helper_nr, selected_helper;
...
+}
+
+static LPWSTR find_exe(LPWSTR exe_name)
+{
+	static LPWSTR path;
+	static size_t alloc;
+	WCHAR env[32768], *p;
+	size_t exe_name_len = wcslen(exe_name), env_size;
+
+	env_size = GetEnvironmentVariableW(L"PATH", env, 32768);
```
bbolli: Use `sizeof env`.

git-extra/git-credential-chooser.c
dscho: `sizeof(env)` is sadly double the number of characters (`env` is a Unicode string). Besides, in this case I really wanted to use the documented maximal length of an environment variable. Do you know any constant for that?

dscho: Aha! It's `_MAX_ENV`: https://docs.microsoft.com/en-us/cpp/c-runtime-library/max-env?view=vs-2019

bbolli:
>> I propose to rewrite this helper in C++ where there's std::wstring and std::vector classes that do all this low-level stuff for you.

> Hmm. That would be the first one C++ application we include in Git for Windows, I think, and I am not sure I want to deal with the dependencies...

Checking my Git for Windows installation, all the needed DLLs are present under `/mingw64/bin`. According to the `ldd` output, these are `libstdc++-6.dll`, `libgcc_s_seh-1.dll` and `libwinpthread-1.dll`.

Would you be interested in a follow-up series that does the conversion to C++?

git-extra/git-credential-chooser.c
dscho: I think that's okay. The credential helper process is short-lived, anyway.

git-extra/git-credential-chooser.c
dscho: Good point!

git-extra/git-credential-chooser.c
dscho: Sharp eyes! I fixed it.

bbolli:
All commit messages start with "credential-helper-selector: ", but the tool is called "git-credential-chooser".

dscho:
> All commit messages start with "credential-helper-selector: ", but the tool is called "git-credential-chooser".

I just force-pushed... It should all be called `helper-selector` now.

dscho:
> Checking my Git for Windows installation, all the needed DLLs are present under `/mingw64/bin`. According to the `ldd` output, these are `libstdc++-6.dll`, `libgcc_s_seh-1.dll` and `libwinpthread-1.dll`.
> 
> Would you be interested in a follow-up series that does the conversion to C++?

If you really want to... But I'd rather not. In my experience, converting from one language to another always leaves a slightly odd taste: it still looks quite a bit like the old language and that does not feel right.

In this instance, I think it would be a bit hard to break the functions apart and turn them into proper methods. My design clearly had C in mind...

dscho:
BTW I just checked, and I was wrong: we already build a C++ executable in `git-extra`: `WhoUses.exe` (I am not really proud of that one, it does not work reliably, but that's beside the point).

bbolli:
All right, no problem!

dscho:
BTW thank you *so much* for your review! I think it does indeed look much better using `_MAX_ENV`, and the `FindClose()` also looks much better!

bbolli:
There's also SystemInfo.cpp, but I'm not sure if this is only used during the build.

dscho:
> There's also SystemInfo.cpp, but I'm not sure if this is only used duri

Yep, that's used in `WhoUses.exe`. As you can see, it had a proper object-oriented design ;-)

bbolli:
> Yep, that's used in WhoUses.exe.

Yes, I've just noticed this.

BTW, I'm not of the opinion that every C++ program must have an OO design. Just using the STL brings lots of advantages Re: resource management.

dscho:
> I'm not of the opinion that every C++ program must have an OO design. Just using the STL brings lots of advantages Re: resource management.

That's a valid point.

In this case, I did have an ulterior motive: maintaining Git for Windows felt like a burden, so I wanted to do something completely different, and challenging myself to implementing a Windows GUI app in pure C without any support library whatsoever definitely improved my mood... ;-)

It also leaves the opportunity to copy-paste code into other projects, and since it is in C it can be used in both C and C++ projects. I like that flexibility a lot.

bbolli:
Glad you could let go of some of the stress!

PR #247, opened 2019-05-08
git-prompt: source generic bash completion scripts from the user's home

sschuberth:
This allows e.g.

https://github.com/gradle/gradle-completion#install-manually-1

to work.

### DIFF
--- a/git-extra/git-prompt.sh
+++ b/git-extra/git-prompt.sh
@@ -1,3 +1,8 @@
+for c in $(find "$HOME/bash_completion.d" -name "*-completion.bash" -type f 2> /dev/null)
+do
+	. $c
+done
+
 if test -f /etc/profile.d/git-sdk.sh
 then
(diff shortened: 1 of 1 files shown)


git-extra/git-prompt.sh
```diff
@@ -1,3 +1,5 @@
+. "$HOME"/bash_completion.d/*-completion.bash
```
dscho: Does this even work when there are more than one scripts in that directory, or when that directory does not even exist?

I think we need this instead:

```sh
for c in "$HOME"/bash_completion.d/*.bash
do
    # Handle absence of any scripts (or the folder) gracefully
    test ! -f "$c" ||
    . "$c"
done
```

sschuberth: > NIH syndrome?

No, just usually disagreeing with your coding taste.

dscho: We're saying the same thing.

PR #251, opened 2019-07-19
Set ssh.variant for plink or tortoise plink

webstech:
git-for-windows/git issue #2018 suggests ssh.variant should be set if the user selects plink for ssh.  This change expands on that requirement.  
- adds some text to indicate ssh.variant will be set
- adds a checkbox to select if tortoise plink is being used.  This affects the command parameters when running plink
- sets ssh.variant to putty or tortoiseshell if needed
- unset ssh.variant if it was set in a previous install
- track previous selection of the checkbox

Installer git configuration settings are not always removed during uninstall and this update does not change current behavior.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1694,7 +1694,7 @@ begin
             'variables will be adjusted to point to the following executable:',TabOrder,Top,Left);
         EdtPlink:=TEdit.Create(PuTTYPage);
         EdtPlink.Left:=ScaleX(Left+24);
-        EdtPlink.Top:=ScaleY(Top+9);
+        EdtPlink.Top:=ScaleY(Top);
         with EdtPlink do begin
             Parent:=PuTTYPage.Surface;
 
@@ -1717,7 +1717,7 @@ begin
         end;
         BtnPlink:=TButton.Create(PuTTYPage);
         BtnPlink.Left:=ScaleX(Left+344);
-        BtnPlink.Top:=ScaleY(Top+9);
+        BtnPlink.Top:=ScaleY(Top);
         with BtnPlink do begin
             Parent:=PuTTYPage.Surface;
             Caption:='...';
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -401,6 +401,7 @@ var
     PuTTYPage:TWizardPage;
     RdbSSH:array[GS_OpenSSH..GS_Plink] of TRadioButton;
     EdtPlink:TEdit;
+    TortoisePlink:TCheckBox;
 
     // Wizard page and variables for the HTTPS implementation (cURL) settings.
     CurlVariantPage:TWizardPage;
@@ -1690,8 +1691,9 @@ begin
         RdbSSH[GS_Plink]:=CreateRadioButton(PuTTYPage,'Use (Tortoise)Plink',
             'PuTTY sessions were found in your Registry. You may specify the path'+#13+
             'to an existing copy of (Tortoise)Plink.exe from the TortoiseGit/SVN/CVS'+#13+
-            'or PuTTY applications. The GIT_SSH and SVN_SSH environment'+#13+
-            'variables will be adjusted to point to the following executable:',TabOrder,Top,Left);
+            'or PuTTY applications. "ssh.variant" will be set in the GIT configuration. '+#13+
+            'The GIT_SSH and SVN_SSH environment variables will be adjusted to point '+#13+
+            'to the following executable:',TabOrder,Top,Left);
         EdtPlink:=TEdit.Create(PuTTYPage);
         EdtPlink.Left:=ScaleX(Left+24);
         EdtPlink.Top:=ScaleY(Top);
@@ -1714,7 +1716,9 @@ begin
 
             Width:=ScaleX(316);
             Height:=ScaleY(13);
+            TabOrder:=TabOrder;
         end;
+        TabOrder:=TabOrder+1;
         BtnPlink:=TButton.Create(PuTTYPage);
         BtnPlink.Left:=ScaleX(Left+344);
         BtnPlink.Top:=ScaleY(Top);
@@ -1724,8 +1728,24 @@ begin
             OnClick:=@BrowseForPuTTYFolder;
             Width:=ScaleX(21);
             Height:=ScaleY(21);
+            TabOrder:=TabOrder;
         end;
-        Top:=Top+30;
+        TabOrder:=TabOrder+1;
+        Top:=Top+29;
+
+        // Add checkbox for tortoise plink
+        TortoisePlink:=TCheckBox.Create(PuTTYPage);
+        TortoisePlink.Left:=ScaleX(Left+24);
+        TortoisePlink.Top:=ScaleY(Top);
+        with TortoisePlink do begin
+            Caption:='Set ssh.variant for Tortoise Plink';
+            Parent:=PuTTYPage.Surface;
+            Width:=ScaleX(405);
+            Height:=ScaleY(17);
+            TabOrder:=TabOrder;
+        end;
+        TabOrder:=TabOrder+1;
+        Top:=Top+17;
 
         // Restore the setting chosen during a previous install.
         case ReplayChoice('SSH Option','OpenSSH') of
@@ -1734,6 +1754,10 @@ begin
         else
             RdbSSH[GS_OpenSSH].Checked:=True;
         end;
+
+        data:=ReplayChoice('Tortoise Option','');
+        if (data='true') then
+            TortoisePlink.Checked:=True;
     end else begin
         PuTTYPage:=NIL;
     end;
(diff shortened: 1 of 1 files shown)


installer/install.iss
```diff
@@ -1730,6 +1754,10 @@ begin
         else
             RdbSSH[GS_OpenSSH].Checked:=True;
         end;
+
+        data:=ReplayChoice('Tortoise Option','');
+        if (data='true') then
+            TortoisePlink.Checked:=True;
```
dscho: In the surrounding code, we don't seem to call `ReplacyChoice()`. Where do the other pages call that function?

installer/install.iss
```diff
@@ -2525,9 +2553,21 @@ begin
...
+            if TortoisePlink.Checked then begin
+                Log('Line {#__LINE__}: Setting config option ssh.variant=tortoiseplink');
+                GitSystemConfigSet('ssh.variant','tortoiseplink')
+            end else begin
+                Log('Line {#__LINE__}: Setting config option ssh.variant=putty');
+                GitSystemConfigSet('ssh.variant','putty');
+            end;
+        end else begin
+            Log('Line {#__LINE__}: Clearing config option ssh.variant');
+            GitSystemConfigSet('ssh.variant',#0);
```
dscho: Since this is the much shorter conditional arm of the `if ... then ... else` construct, maybe negate the condition and switch the arms around? That way, it would be easier to read.

installer/install.iss
```diff
@@ -2758,6 +2798,8 @@ begin
     end else if RdbSSH[GS_Plink].Checked then begin
         Data:='Plink';
         RecordChoice(PreviousDataKey,'Plink Path',EdtPlink.Text);
+        if TortoisePlink.Checked then
+            RecordChoice(PreviousDataKey,'Tortoise Option','true');
```
dscho: There is no corresponding `false` statement... Maybe this instead?

```
RecordChoice(PreviousDataKey,'Using Tortoise Plink',TortoisePlink.Checked);`
```

installer/install.iss
```diff
@@ -1687,10 +1688,15 @@ begin
         RdbSSH[GS_OpenSSH]:=CreateRadioButton(PuTTYPage,'Use OpenSSH','This uses ssh.exe that comes with Git. The GIT_SSH and SVN_SSH'+#13+'environment variables will not be modified.',TabOrder,Top,Left);
 
         // 2nd choice
-        RdbSSH[GS_Plink]:=CreateRadioButton(PuTTYPage,'Use (Tortoise)Plink','PuTTY sessions were found in your Registry. You may specify the path'+#13+'to an existing copy of (Tortoise)Plink.exe from the TortoiseGit/SVN/CVS'+#13+'or PuTTY applications. The GIT_SSH and SVN_SSH environment'+#13+'variables will be adjusted to point to the following executable:',TabOrder,Top,Left);
+        RdbSSH[GS_Plink]:=CreateRadioButton(PuTTYPage,'Use (Tortoise)Plink',
+            'PuTTY sessions were found in your Registry. You may specify the path'+#13+
+            'to an existing copy of (Tortoise)Plink.exe from the TortoiseGit/SVN/CVS'+#13+
+            'or PuTTY applications. "ssh.variant" will be set in the GIT configuration. '+#13+
+            'The GIT_SSH and SVN_SSH environment variables will be adjusted to point '+#13+
+            'to the following executable:',TabOrder,Top,Left);
```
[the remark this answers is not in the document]
webstech: Do you want a commit with just the reformatting before the new text was added to the description?  Just trying to understand.  I reformatted to make it easier to see the relative line lengths when editing.

installer/install.iss
```diff
@@ -1687,10 +1688,15 @@ begin
...
         EdtPlink:=TEdit.Create(PuTTYPage);
         EdtPlink.Left:=ScaleX(Left+24);
-        EdtPlink.Top:=ScaleY(Top+9);
+        EdtPlink.Top:=ScaleY(Top);
```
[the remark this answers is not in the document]
webstech: This change reduces the space between the description text and the edit box.  The current install has a bit more space but adding an extra line to the description and the Tortoise plink checkbox got a bit crowded.  Initially, the checkbox was partially visible but there was no scroll bar.  I think the last paste in the issue shows the new spacing.

installer/install.iss
webstech: This is immediately after the replay for the radio buttons starting at line 1750.

installer/install.iss
webstech: Ok.

installer/install.iss
webstech: I thought it was just noise to track this if the user was not requesting it.  It will not be present in any current installs.  I can always set it if you would prefer.

webstech:
I apologize for royally messing this up.  Testing requires a reinstall of git.  Did I lose auto sign-off?

So, if there are to be multiple commits (3?), would it be better if I closed this and opened a new PR with the commits?

dscho:
Wait a second... 🤔 It seems that GitHub was unable to calculate whether this pull request is mergeable!

I'll check again if someone adds some more commits to this pull request!

webstech:
Sigh.  Rebasing master cleared out my changes and automatically closed this.  I have my change on plink branch.  I will merge to master but would like your replies for direction first re: extra commits.

webstech:
@dscho Re-opened with a cleaned up commits and requested changes except one.  I left the RecordChoice of the 'Tortoise Option' to only be done if checked.  I can change if you prefer.

installer/install.iss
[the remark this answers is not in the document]
dscho: > Do you want a commit with just the reformatting before the new text was added to the description?

Yes, that would be most preferred.

installer/install.iss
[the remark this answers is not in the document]
dscho: > This change reduces the space between the description text and the edit box. The current install has a bit more space but adding an extra line to the description and the Tortoise plink checkbox got a bit crowded. Initially, the checkbox was partially visible but there was no scroll bar. I think the last paste in the issue shows the new spacing.

This already sounds like a pretty rad commit message to me.

installer/install.iss
dscho: Ah, right, there were no checkboxes to record previously. Thanks for clarifying!

installer/install.iss
dscho: The problem is: if a user will check this box, but in a subsequent install _uncheck_ it, it won't be recorded with the new preference. The old preference will prevail, forever.

installer/install.iss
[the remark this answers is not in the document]
dscho: (Maybe amend bd5bb1e7112c4091146cfc5dcee5718143893191?)

installer/install.iss
```diff
@@ -2529,9 +2553,21 @@ begin
     DeleteMarkedEnvString('GIT_SSH');
     DeleteMarkedEnvString('SVN_SSH');
 
-    if (PuTTYPage<>NIL) and RdbSSH[GS_Plink].Checked then begin
-        SetAndMarkEnvString('GIT_SSH',EdtPlink.Text,True);
-        SetAndMarkEnvString('SVN_SSH',EdtPlink.Text,True);
+    if (PuTTYPage<>NIL) then begin
+        if RdbSSH[GS_OpenSSH].Checked then begin
```
dscho: Shouldn't this be `if not RdbSSH[GS_Plink].Checked` instead? Just in case that future patches want to add a third option there?

dscho:
I think we're _really_ close now!

installer/install.iss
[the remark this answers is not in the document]
webstech: Amended the commit message.

installer/install.iss
webstech: If another variant is added, isn't it likely to need to set ssh.variant (and env vars) to specific values?

How about:
    if (PuTTYPage<>NIL) then begin
        GitSystemConfigSet('ssh.variant',#0);
        if RdbSSH[GS_Plink].Checked then begin
If the putty page is present, clear and reset (if needed) the ssh.variant config value.  It may mean double calls to git config but ensures ssh.variant is cleared if there are any possible putty states and any future variants start with a clean slate.

installer/install.iss
webstech: For the average install, isn't the install-options.txt re-written for each install?  If the user is installing with /saveinf == /loadinf then yes, the old preference will prevail.  Perhaps I misunderstood this.  Having said that, to avoid issues, I will change to always set the turtle option.  I was trying to avoid this because I see the use of plink options going away over time (just an opinion).

webstech: Just noting that 'Plink Path' is not cleared if changing to Open SSH.  Similar to the tortoise check box, it will show the previous value when saveinf == loadinf.  No big deal since it does not affect execution.

installer/install.iss
dscho: I like that idea!

installer/install.iss
dscho: > For the average install, isn't the install-options.txt re-written for each install?

Yes, but the choices are not replayed from the `install-options.txt` file, but via InnoSetup's mechanism, from the registry.

> Just noting that 'Plink Path' is not cleared if changing to Open SSH.

Right, but that is a bit different. It means that that option is remembered when the user chooses OpenSSH, and should they want to choose Plink again, conveniently the path will be remembered.

In the issue at hand, the issue is if the user checks that check box, but then makes up their mind and re-installs, _unchecking_ the checkbox. We'd want to record this _new_ preference (unchecked) rather than leave the previous preference (checked) in place.

PR #253, opened 2019-08-20
Use modern default font for credential helper selector dialog

Chocobo1:
Screenshots:
Old:
old
New: 
new


PhilipOakley:
The commit message will need, in addition to your sign-off (check what you are signing for!), a rationale for why the old font is/was/could be a problem, and why the new font will be better (including any comments about localisation). Include any rationale for any extra, or special, code.

These things may seem obvious to you now, but if there are changes or problems later it can easily look like some random change because no one remembers what happened. If it's resolving an issue say which one e.g. "resolves #xxxx"

Chocobo1:
I've updated the commit msg.

dscho:
I like the new look!

dscho:
Excellent, thank you so much!

PR #255, opened 2019-08-30
Add dev directory during setup

stepro:
A minimal fix for git-for-windows/git#2291.

Note no change in mingit because that is just git.exe, and the missing /dev directory is only a problem for bash.exe.

Built and validated the regular installer, NuGet package and portable installers.


dscho:

I augmented the commit messages, spend a couple of hours to get the `CreateCygwinSymlink()` function right and to verify that everything works as intended.

stepro:
Thanks @dscho. Are you thinking a patch release for this fix or wait until 2.24?

dscho:
@stepro before I even think about that, I want to hear a couple of confirmations that the latest [snapshot](https://wingit.blob.core.windows.net/files/index.html) works as intended.

stepro:
I tried the installer and it worked fine for me.

PR #256, opened 2019-09-05
Revert "git-lfs: adjust config settings to the latest ones"

rimrul:
I've messed up the default LFS config in git-for-windows/git#2140, but this should fix it.


dscho:
Thanks!

PR #257, opened 2019-09-20
fix installer's Product Version inconsistency with older versions

AudricGuerin:
Why: Third-party applications or IT tools must be able to know
if the latest Git version is installed on the system
by means of version number comparisons.

How: Remove platform identifier from the Product Version.

Product Version now displays x.x.x.x instead of x.x.x.windows.x

See https://github.com/git-for-windows/git/issues/2223

Signed-off-by: Audric Guerin <49029709+AudricGuerin@users.noreply.github.com>

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -50,8 +50,6 @@ SignTool=signtool
 #endif
 
 #define FILE_VERSION GetFileVersion(SOURCE_DIR+'\'+MINGW_BITNESS+'\bin\git.exe')
-#define PROD_VERSION GetStringFileInfo(SOURCE_DIR+'\'+MINGW_BITNESS \
-				+'\bin\git.exe', 'ProductVersion')
 
 ; Installer-related
 AllowNoIcons=yes
@@ -59,7 +57,7 @@ AppName={#APP_NAME}
 AppPublisher=The Git Development Community
 AppPublisherURL={#APP_URL}
 AppSupportURL={#APP_CONTACT_URL}
-AppVersion={#PROD_VERSION}
+AppVersion={#APP_VERSION}
 ChangesAssociations=yes
 ChangesEnvironment=yes
 CloseApplications=no


PhilipOakley:
I think this may also need a method for updating the `git version --build-options` so that when some user makes a report on the main Git mailing list folks can tell if they are talking about Linux version or Windows version , especially if a Linux maint version (.1) becomes confused with the Windows initial version (.1)

(I accept that some users don't know how to report bugs so are missing such info, but let's help those who make an effort..)

PhilipOakley:
> I think this may also need a method for updating the `git version --build-options`

Maybe add a 
`OS platform: xxx` line

AudricGuerin:
Only the Product Version of the installer was changed so `git version` still returns:
git version 2.23.0.windows.1

There is no problem with git-for-windows adding a platform identifier for its internal versioning purposes.
I totally understand why you would need this as I had to do such a thing this year for my company... However, I would not add an app-specific version identifier into the 'Product Version' or 'File Version' of any Windows installers or binaries of the project.

Instead I would write it in a .txt, .xml or .json file as suggested by @dscho here:
https://github.com/git-for-windows/git/issues/2223#issuecomment-502240425
> For example, I would be totally willing to accept a PR that writes the current version in plain text (and in the expected, non-SemVer format) into the top-level directory.

And if you really want a Windows user to access Git internal version identifier through _Properties > Details_ of the installer or _Settings > Apps & features_, then add it to 'Product Name' as @sendqueery said:
https://github.com/git-for-windows/git/issues/2223#issuecomment-502730291
> If you really want it that badly, change the product name field (which also displays the version for some reason) to something like "Git version x.x.x for Windows". (Sidenote: don't do this, as it's still an unnecessary tag. We know it's Windows software by virtue of the fact that WE'RE INSTALLING A WIN32 BINARY ON WINDOWS.)

And by the way, thank you for redirecting me to the right place, Philip. ;)

PhilipOakley:
Do you fancy copying some of that clarifying explanation into the commit message so I don't fall into the same traps again... Especially if we have to explain it (our version**s** vs their single version) to the Linux upstream mail list etc etc.

AudricGuerin:
I added some remarks into the commit message to clarify that the Product Version of the installer file is the only version identifier affected by this commit.

PR #259, opened 2019-10-09
Make `sdk build` less cumbersome to use

dscho:
The recommended way in the Git for Windows SDK to start developing patches for any package is to run something like this:

```sh
sdk cd <package-name>
sdk build
```

However, this is not as easy to use as it should: after editing any source file, it would complain about checksum mismatches (which is correct, but not helpful). For other packages, it will complain that it does not find the GPG key with which the upstream sources were signed.

Let's make all of this much more fun by skipping those checks. The checksums will be updated anyway, as part of the semi-automated process to build and upload the package to Git for Windows' Pacman repository. Not anything a contributor should need to worry about.

### DIFF
--- a/git-extra/git-sdk.sh
+++ b/git-extra/git-sdk.sh
@@ -147,7 +147,7 @@ sdk () {
 			sdk init MSYS2-packages &&
 			(cd "$src_dir/$2" &&
 			 test -d src/msys2-runtime ||
-			 makepkg --nobuild --syncdeps --noconfirm) &&
+			 makepkg --nobuild --syncdeps --noconfirm --skipchecksums --skippgpcheck) &&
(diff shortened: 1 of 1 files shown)


dscho:
Thanks @PhilipOakley !

bbolli:
Sorry for the late review, but would it make sense to move all those common options into a variable to reuse where needed? This could also be a follow-up commit, as this is already merged now.

Otherwise ACKed by me.

dscho:
> would it make sense to move all those common options into a variable to reuse where needed?

Good idea! Could you give me a hand in this?

bbolli:
Sure, in a few hours when I'm back home...

bbolli:
PR sent just now: #260

PR #260, opened 2019-10-11
git-sdk: don't repeat the common makepkg options

bbolli:
In order to simplify changes similar to 5bbd68e ("sdk build: do not
verify checksums nor GPG keys", 2019-10-09), move the common makepkg
options into a variable.

### DIFF
--- a/git-extra/git-sdk.sh
+++ b/git-extra/git-sdk.sh
@@ -129,6 +129,11 @@ sdk () {
 		printf "%s " git-sdk.sh sdk.completion ReleaseNotes.md \
 			install.iss
 		;;
+	# for building
+	makepkg|makepkg-mingw)
+		cmd=$1; shift
+		$cmd --syncdeps --noconfirm --skipchecksums --skippgpcheck "$@"
+		;;
 	# here start the commands
 	init-lazy)
 		case "$2" in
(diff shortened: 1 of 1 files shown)


git-extra/git-sdk.sh
```diff
@@ -20,6 +20,9 @@ case "$rootdir" in
 esac
 
 sdk () {
+	# common options for makepkg calls
+	MAKEPKG_COMMON_OPTS="--syncdeps --noconfirm --skipchecksums --skippgpcheck"
```
dscho: I am slightly worried that this variable might get re-set between this line and the lines where it is used.

Maybe a better idea would be to introduce a function? I.e. something like

```sh
sdk () {
    case "$1" in
    [...]
    makepkg|makepkg-mingw)
        argv0=/usr/bin/$1
        shift
        exec $argv0 --syncdeps --noconfirm --skipchecksums --skippgpcheck "$@"
        ;;
    [...]
```

dscho: (And it would be used as e.g. `sdk makepkg --nobuild`.)

dscho:
Thank you for your help!

git-extra/git-sdk.sh
dscho: > Who or what would change the variable's value?

Anyone? The `sdk` function is defined in the scope of the interactive shell. Any `MAKEPKG_COMMON_OPTS=hahaha` in the Git Bash will make this break.

bbolli: I thought the intention of your original change was that users would only use `sdk build` to build stuff. In this case, the variable would be set to the correct value before use. My assumption seems to have been wrong.

dscho: > I thought the intention of your original change was that users would only use `sdk build` to build stuff.

Yes. And in my mind, the code of `build)` would call the code of `makepkg|makepkg-mingw)`, that way we can avoid relying on a global (and user-visible, and user-modifiable) environment variable.

bbolli: All right. New version incoming.

dscho:
Thank you!!!

PR #262, opened 2019-10-17
Update installer/portable/mingit for the new system config location

dscho:
This is a follow-up PR for https://github.com/git-for-windows/git/pull/2358, and it will need to be merged _before_ that one, so that the next snapshot won't be broken.

The idea is to prepare the scripts that generate the Git for Windows installer, the Portable Git and the MinGit versions to accommodate for Git versions that have only one system config: Instead of reading `C:\ProgramData\Git\config` and then `C:\Program Files\Git\mingw64\etc\gitconfig` (or for 32-bit Git on 64-bit Windows, `C:\Program Files (x86)\Git\mingw32\etc\gitconfig`), we now only read from the quite intuitive location `C:\Program Files\Git\etc\gitconfig`, i.e. without the `mingw64` part.

This also addresses the issue raised many times that the Portable Git relied on settings configured _outside_ of the Portable Git (`C:\ProgramData\Git\config`), in addition to squelching complaints that there is no equivalent to the `--system` option for writing to the ProgramData config.

Finally, the ProgramData config was meant to be a truly Windows-wide config, but it lacked buy-in. For example, JGit (which is implicitly used by pretty much every Android Studio/Eclipse user) does not support it. At the same time, libgit2 seems to never have stopped looking for `C:\Program Files\Git\etc\gitconfig`. In short: the ProgramData config failed to serve its purpose, anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ edit-git-bash.exe
 /git-for-windows-keyring/git-for-windows-keyring-*.src.tar.gz
 /installer/ReleaseNotes.html
 /installer/config.iss
-/installer/gitconfig.system
 /installer/install.log
 /installer/is-unicode.exe
 /installer/file-list.iss
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ edit-git-bash.exe
 /installer/is-unicode.exe
 /installer/file-list.iss
 /installer/package-versions.txt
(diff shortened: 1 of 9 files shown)


dscho:
I merged it because Git v2.24.0 was already released, and now I have to scramble to get Git for Windows v2.24.0 out...

PR #263, opened 2019-10-18
Upgrade mingw-w64-git-lfs to 2.9.0

slonopotamus:
Signed-off-by: Marat Radchenko <marat@slonopotamus.org>

### DIFF
--- a/mingw-w64-git-lfs/PKGBUILD
+++ b/mingw-w64-git-lfs/PKGBUILD
@@ -3,7 +3,7 @@
 _realname="git-lfs"
 pkgbase="mingw-w64-${_realname}"
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=2.8.0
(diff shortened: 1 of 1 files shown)


dscho:

You could not know this, but I have automation to perform this task. It uses [`please.sh upgrade git-lfs`](https://github.com/git-for-windows/build-extra/blob/e64636f001404eb5c53d3ec2efdcee947393fb8d/please.sh#L2720-L3677) which takes all the burden from me to figure out whether there is a new version, and if so, what SHA-256 checksums the file has, and also what the tree structure is inside (because sometimes, the `git-lfs.exe` is at the top-level directory inside that `.zip`, sometimes it is not).

As it were, I had to merge #262 first to make all this work, that's why I did not run the automation as soon as I got the notification about a new Git LFS version...

But now I did, so I'll close this PR because it is no longer necessary.

slonopotamus:
Sure, no problem

PR #264, opened 2019-10-30
Installer: stop complaining about "downgrading" when upgrading from an rc version

dscho:
There is a pretty old bug where the Git for Windows installer warns about a "downgrade" when upgrading from, say, v2.23.0-rc2 to v2.23.0, because the versions that are actually compared are of the format `N.N.N.N` (by stripping out non-numerical letters, in the example, the compared versions would be `2.23.0.2` vs `2.23.0.0`).

Let's fix this by trying to be a bit more careful when we're about to warn about a downgrade, and compare the output of the installed `git version` vs the one we're about to install.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -869,6 +869,16 @@ begin
     ExitProcess(0);
 end;
 
+function CountDots(S:String):Integer;
+var
+    i:Integer;
+begin
+    Result:=0;
+    for i:=1 to Length(S) do
+        if (S[i]=#46) then
+            Result:=Result+1;
+end;
+
 function IsDowngrade(CurrentVersion,PreviousVersion:String):Boolean;
 var
     Path:String;
@@ -876,8 +886,8 @@ var
 begin
     Result:=(VersionCompare(CurrentVersion,PreviousVersion)<0);
 #ifdef GIT_VERSION
-    if Result then begin
-        // maybe the previous version was a prerelease?
+    if Result or (CountDots(CurrentVersion)>3) or (CountDots(PreviousVersion)>3) then begin
+        // maybe the previous version was a prerelease (prereleases have five numbers: v2.24.0-rc1.windows.1 reduces to '2.24.0.1.1')?
         if (RegQueryStringValue(HKEY_LOCAL_MACHINE,'Software\GitForWindows','InstallPath',Path))
                 and (Exec(ExpandConstant('{cmd}'),'/c ""'+Path+'\cmd\git.exe" version >"'+ExpandConstant('{tmp}')+'\previous.version""','',SW_HIDE,ewWaitUntilTerminated,i))
                 and (i=0) then begin
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -816,6 +816,13 @@ begin
         if Previous<0 then begin
             if Current>=0 then
                 Result:=+1;
+            Result:=Ord(CurrentVersion[i])-Ord(PreviousVersion[j]);
+            if (Result=0) then begin
+                // skip identical non-numerical characters
+                i:=i+1;
+                j:=j+1;
+                Continue;
+            end;
             Exit;
         end;
         if Current<0 then begin
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -869,6 +869,26 @@ begin
     ExitProcess(0);
 end;
 
+function IsDowngrade(CurrentVersion,PreviousVersion:String):Boolean;
+var
+    Path:String;
+    i,j,CurrentLength,PreviousLength:Integer;
+begin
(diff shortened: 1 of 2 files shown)


dscho:
@webstech could you have a look, too? You've been the second-most active developer in this area in the past year...

installer/install.iss
```diff
@@ -649,6 +649,10 @@ begin
         Result:='(no output)'
     else
         Result:=Contents;
+    if (Length(Result)>0) and (Result[Length(Result)]=#10) then
+        SetLength(Result,Length(Result)-1);
+    if (Length(Result)>0) and (Result[Length(Result)]=#13) then
+        SetLength(Result,Length(Result)-1);
```
webstech: Does TrimRight(Result) do this?

installer/release.sh
```diff
@@ -119,6 +119,11 @@ else
 	die "Could not generate file list"
 fi
 
+cmd_git="$(echo "$LIST" | grep '^cmd/git\.exe$')"
```
webstech: Will this get an error if the --skip-files option is specified?  I understand that is not part of a normal build of the install.

bbolli:
I find this way of doing version comparisons a bit brittle and hardly extensible (imagine if we had alpha and beta version numbers). The "right" way would be to build a tuple (prio, number) out of each segment of the version string. Stand-alone numbers get a high prio, rcN gets one less, beta and alpha one less each as well. Then you compare the list of tuples from left to right.

Other than that, the patches seem to do what you intend :-)

PhilipOakley:
> The `#ifdef GIT_VERSION` is about..

> .. it is essentially the dynamically-generated part of the installer.

Hadn't appreciated that part (I think you are saying this is dynamic at install time..?)

Hmm, re-reading, I see that that part (the `GIT_VERSION` is defined in the _`config.iss`_ file, rather than being an `if` within the installer interrogating the installation (two different places, needing two different types of `if`) - mainly writing this out long hand for other readers (assuming I got it right :wink:)

installer/install.iss
dscho: Almost. It would also trim trailing spaces, which I don't want to do. Also, it would trim more than one trailing newline, which I _also_ don't want to do...

Technically, I only want to trim CR when there was an LF to begin with, but this way the code looks nicer and I don't _really_ care about requiring that LF before the CR...

dscho:
> I find this way of doing version comparisons a bit brittle and hardly extensible (imagine if we had alpha and beta version numbers). The "right" way would be to build a tuple (prio, number) out of each segment of the version string. Stand-alone numbers get a high prio, rcN gets one less, beta and alpha one less each as well. Then you compare the list of tuples from left to right.

I'd rather cross that bridge when (if?) we need to... Think [YAGNI](https://www.martinfowler.com/bliki/Yagni.html). It would require a _ton_ of work: we would need to build a new class for that, non-trivial parsing, etc. And: no regression test suite we can use to build/maintain confidence...

> Other than that, the patches seem to do what you intend :-)

:+1:

> the `GIT_VERSION` is defined in the _`config.iss`_ file, rather than being an `if` within the installer interrogating the installation

Yes, `GIT_VERSION` reflects the version reported by `git version` as of the `git.exe` _about to be installed_, as opposed to the version reported by the _installed_ `git version`.

And yes, the two different types of `if` reflect compile time vs run time.

dscho:
> > @@ -119,6 +119,11 @@ else
>  	die "Could not generate file list"
>  fi
>
> +cmd_git="$(echo "$LIST" | grep '^cmd/git\.exe$')"
>
> Will this get an error if the --skip-files option is specified?  I understand that is not part of a normal build of the install.

Well, the `grep` might fail, but the exit code is ignored.

installer/install.iss
```diff
@@ -858,6 +869,36 @@ begin
...
+function IsDowngrade(CurrentVersion,PreviousVersion:String):Boolean;
+var
+    Path:String;
+    i,j,CurrentLength,PreviousLength:Integer;
+begin
+    Result:=(VersionCompare(CurrentVersion,PreviousVersion)<0);
+#ifdef GIT_VERSION
+    if Result or (CountDots(CurrentVersion)>3) or (CountDots(PreviousVersion)>3) then begin
+        // maybe the previous version was a prerelease (prereleases have five numbers: v2.24.0-rc1.windows.1 reduces to '2.24.0.1.1')?
+        if (RegQueryStringValue(HKEY_LOCAL_MACHINE,'Software\GitForWindows','InstallPath',Path))
```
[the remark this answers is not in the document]
dscho: True.

I think that we're currently safe, given that the installer always requests to be run elevated.

dscho:
Thank you all, reviewers!

PR #265, opened 2019-11-05
Teach the installer to create Cygwin symlinks even in non-US locales

dscho:
This fixes https://github.com/git-for-windows/git/issues/2388.


dscho:
Sadly, I have not heard back from the reporters, but no less than the MSYS2 maintainer themselves helped me verify this.

PR #266, opened 2019-11-08
Handle patch number and some other improvements and fixes

sinloss:
**Handle patch number** 

While there's a patch number greater than 1 appended, the name of the file about to be downloaded should also have that number appended after the version part.   
For example: `v2.24.0.windows.2` ==> `PortableGit-2.24.0.2-64-bit.7z.exe`

**Remove the inadvertently added "-n" option**

In an updating scenario, this "-n" option in the `update` function would prevent files from being overwritten, yet our intention indeed is to overwrite them.

**Move the backed up files to /tmp instead of "rm -rf" them**

This can help to keep the original files at least in the temporary folder in case something unexpected happen. Users can restore the original files as they wish.

**Let `Git Bash Here` perform an identical behavior on Cygwin comparing to Git-for-windows**

To let `Git Bash Here` on Cygwin perform the identical behavior, the final executed bash in the command string of the shell item should have "--login -i" options. So that the environment is well prepared the same way `Git-for-windows` does. And this also makes the bootstrap bash of the command string redundant thus should be removed.

Signed-off-by: liuxy <391861737@qq.com>

### DIFF
--- a/git-extra/getgit
+++ b/git-extra/getgit
@@ -135,9 +135,10 @@ else
 fi
 latest=$(curl $gh_tags)
 version=$(echo $latest | grep -Po "\d+\.\d+.\d+")
-echo "Gathered { system: \"$system\", bit: \"$bit\", latest: \"$latest\", version: \"$version\" }"
+patch=$(echo $latest | grep -Po "windows\.\K\d")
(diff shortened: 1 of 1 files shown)


dscho:
Thank you!

PR #267, opened 2019-11-08
Various system gitconfig-related fixes, in particular for MinGit

dscho:
I made a mistake when I added those `include.path` settings to MinGit's system config which are intended to inherit the installed Git for Windows' defaults: these includes are added _to the beginning_, which means that MinGit's hard-coded `core.autocrlf` and friends always override Git for Windows' defaults. Which defeats the purpose.

Let's fix that, and while at it also remove the no longer needed `http.sslCAInfo` from the system gitconfig of Git for Windows' SDK (the installer will _still_ set it, to help Git LFS use Git for Windows' certificate bundle).

### DIFF
--- a/git-extra/git-extra.install.in
+++ b/git-extra/git-extra.install.in
@@ -216,10 +216,19 @@ test -d "$TMPDIR" || test ! -d "$TMP" || {\
 		sed -i '$i#undef _POSIX_THREAD_SAFE_FUNCTIONS\n#define _POSIX_THREAD_SAFE_FUNCTIONS 200112L\n' $header
 	done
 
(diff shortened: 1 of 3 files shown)


dscho:
@shiftkey could you have a look?

shiftkey:
I think this is fine for GitHub Desktop's packaging, as we unset `http.sslCAInfo` and set `http.schannelUseSSLCAInfo` to `false` on our end

https://github.com/desktop/dugite-native/blob/9ce314bb2dcaa233a3d576f4b797ca2a57c11808/script/build-win32.sh#L74-L86

cc @outofambit @niik for visibility

PR #268, opened 2019-11-08
Show the update message in the Action Center again

dscho:
The `git update-git-for-windows` script tries to be smart and only show the Toast in Windows 8 and later. But the test was broken by an update to Cygwin v3.x, which changed the format of the output of `uname -s`.

This fixes that.

### DIFF
--- a/git-extra/git-update-git-for-windows
+++ b/git-extra/git-update-git-for-windows
@@ -154,7 +154,7 @@ update_git_for_windows () {
 	try_toast=
 	test -z "$use_gui" ||
 	case "$(uname -s)" in
-	*-6.[23]|*-10.0)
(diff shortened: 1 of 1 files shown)


rimrul:
Yeah, that explains it. There's a typo in the PR description though. It should be Windows 8 and later.

dscho:
> There's a typo in the PR description though. It should be Windows 8 and later.

Whoops, that's not even a typo but a misunderstanding. Fixed.

HimbeersaftLP:
A window that displays the download progress would be useful, too. Especially for people with slower internet connections.

dscho:
> A window that displays the download progress would be useful, too. Especially for people with slower internet connections.

Feel free to work on this feature!

PR #269, opened 2019-11-13
Assorted fixes for the `sdk` function

dscho:
Inspired by a couple recent user reports that `sdk cd installer` did not work (because it failed to initialize the `/usr/src/build-extra/` worktree), I went ahead and fixed that, and while at it, also fixed another pet peeve of mine: when running `sdk build` in, say, `mingw-w64-git`, the build was neither parallel nor restricted to the 64-bit build.

This PR addresses those issues.

### DIFF
--- a/git-extra/git-sdk.sh
+++ b/git-extra/git-sdk.sh
@@ -220,14 +220,20 @@ sdk () {
 		;;
 	init)
 		sdk init-lazy "$2" &&
-		if test refs/heads/master = \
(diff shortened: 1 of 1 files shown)


dscho:
> I'm not sure if every package can handle the parallel build from the second commit, but that would be a bug in that package, wouldn't it?

Indeed. There are some `PKGBUILD` files that specifically override it, via `make -j1` and/or `MAKEFLAGS=-j1`.

PR #270, opened 2019-11-20
skip options pages when upgrading

denised:
When upgrading Git-for-Windows, the current default behavior is to go through the entire installation dialog, which shows a number of pages for setting options.  This change suppresses those pages during upgrade, using the previous values (or values specified on the command installer command line, if they are present).  It does so by checking, for each options page, whether all the values on it have previously-set values.  

Behavior on two edge cases:
If a new option page is added, it's values won't be previously set, so it will be shown, even on upgrade.
If the user uninstalls and then reinstalls Git for Windows, all the pages/options are shown.

One known flaw: when all pages are shown, the last page before installation has a button labeled 'Install' instead of 'Next'.  With this change, the user may not see an 'Install' button.  I think that is a reasonable trade-off for the simpler experience for users.

```
        Signed-off-by: Denise Draper <draperd@acm.org>
```

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -372,6 +372,10 @@ var
     // The options chosen at install time, to be written to /etc/install-options.txt
     ChosenOptions:String;
 
+    // Accumulated set of custom pages that have options, and those that have 'new' parameters on them
+    CurrentCustomPageID:Integer;
+    AllCustomPages,CustomPagesWithUnseenOptions:String;
+
     // Previous Git for Windows version (if upgrading)
     PreviousGitForWindowsVersion:String;
 
@@ -899,6 +903,21 @@ begin
 #endif
 end;
 
+{ Represent a set as a string of comma-separated values }
+function IsInSet(var ASet:String;Value:Integer):Boolean;
+begin
+    Result:=(Pos(','+IntToStr(Value)+',',ASet)>0);
+end;
+
+procedure AddToSet(var ASet:String;Value:Integer);
+begin
+    if ASet='' then
+        ASet:=',';
+    if not IsInSet(ASet,Value) then
+        ASet:=ASet+IntToStr(Value)+',';
+end;
+
+
 function InitializeSetup:Boolean;
 var
     CurrentVersion,Msg:String;
@@ -962,6 +981,10 @@ begin
     NoSpaces:=Key;
     StringChangeEx(NoSpaces,' ','',True);
 
+    // A side effect of ReplayChoice is to collect a set of pages that have options on them, and the subset
+    // of those pages have _new_ options (options whose values have no previously-set value)
+    AddToSet(AllCustomPages,CurrentCustomPageID);
+
     // Interpret /o:PathOption=Cmd and friends
     Result:=ExpandConstant('{param:o:'+NoSpaces+'| }');
     if Result<>' ' then
@@ -977,6 +1000,10 @@ begin
         else
             // Restore the settings chosen during a previous install.
             Result:=GetPreviousData(Key,Default);
+            // Check to see if this result was the default, or was previously set.
+            // If it was the default, the user has not seen this option yet.
+            if GetPreviousData(Key,'z'+Result)<>Result then
+                AddToSet(CustomPagesWithUnseenOptions,CurrentCustomPageID)
     end;
 end;
 
@@ -1072,6 +1099,7 @@ end;
 function CreatePage(var PrevPageID:Integer;const Caption,Description:String;var TabOrder,Top,Left:Integer):TWizardPage;
 begin
     Result:=CreateCustomPage(PrevPageID,Caption,Description);
+    CurrentCustomPageID:=Result.ID;
     PrevPageID:=Result.ID;
     TabOrder:=0;
     Top:=8;
@@ -1081,6 +1109,7 @@ end;
 function CreateFilePage(var PrevPageID:Integer;const Caption,Description,SubCaption:String;var TabOrder,Top,Left:Integer):TInputFileWizardPage;
 begin
     Result:=CreateInputFilePage(PrevPageID,Caption,Description,SubCaption);
+    CurrentCustomPageID:=Result.ID;
     PrevPageID:=Result.ID;
     TabOrder:=0;
     Top:=8;
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1577,6 +1577,8 @@ begin
         Height:=WizardForm.CancelButton.Height;
         Top:=WizardForm.CancelButton.Top;
     end;
+    if (PreviousGitForWindowsVersion='') then
+        OnlyShowNewOptions.Hide;
 
     (*
      * Create a custom page for configuring the default Git editor.
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -2496,7 +2496,7 @@ var
     RootKey:Integer;
 begin
     if CurStep=ssInstall then begin
-#ifdef DEBUG_WIZARD_PAGE
+#ifdef DO_NOT_INSTALL
         ExitEarlyWithSuccess();
 #endif
         // Shutdown locking processes just before the actual installation starts.
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -937,9 +937,9 @@ begin
     if (OnlyShowNewOptions.Checked) then begin
         // The "Select Program Group" page is suppressed when
         // re-installing/upgrading/downgrading, but not the Components page.
-        if (PageID=wpSelectProgramGroup) or ((PreviousGitForWindowsVersion<>'') and (PageID=wpSelectComponents)) then
+        if (PageID<=wpSelectProgramGroup) then
             PageID:=FirstCustomPageID;
-        while (PageID<PageIDBeforeInstall) and IsInSet(AllCustomPages,PageID) and not IsInSet(CustomPagesWithUnseenOptions,PageID) do
+        while (PageID<PageIDBeforeInstall) and not IsInSet(CustomPagesWithUnseenOptions,PageID) do
             PageID:=PageID+1;
     end;
     Result:=(PageID=PageIDBeforeInstall);
@@ -2111,9 +2111,12 @@ begin
         end;
         RefreshProcessList(NIL);
         Result:=(GetArrayLength(Processes)=0);
-    end else if OnlyShowNewOptions.Checked then
-        Result:=IsInSet(AllCustomPages,PageID) and not IsInSet(CustomPagesWithUnseenOptions,PageID)
-    else
+    end else if OnlyShowNewOptions.Checked then begin
+        if (IsInSet(AllCustomPages,PageID)) then
+            Result:=not IsInSet(CustomPagesWithUnseenOptions,PageID)
+        else
+            Result:=(PageID<>wpInfoBefore);
+    end else
         Result:=False;
 #ifdef DEBUG_WIZARD_PAGE
     Result:=PageID<>DebugWizardPage
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -373,7 +373,7 @@ var
     ChosenOptions:String;
 
     // Accumulated set of custom pages that have options, and those that have 'new' parameters on them
-    CurrentCustomPageID:Integer;
+    CurrentCustomPageID,FirstCustomPageID:Integer;
     AllCustomPages,CustomPagesWithUnseenOptions:String;
 
     // Previous Git for Windows version (if upgrading)
@@ -927,6 +927,32 @@ begin
         ASet:=ASet+IntToStr(Value)+',';
 end;
 
+function IsLastPageBeforeInstall(PageID:Integer):Boolean;
+begin
+    if (OnlyShowNewOptions.Checked) then begin
+        // The "Select Program Group" page is suppressed when
+        // re-installing/upgrading/downgrading, but not the Components page.
+        if (PageID=wpSelectProgramGroup) or ((PreviousGitForWindowsVersion<>'') and (PageID=wpSelectComponents)) then
+            PageID:=FirstCustomPageID;
+        while (PageID<PageIDBeforeInstall) and IsInSet(AllCustomPages,PageID) and not IsInSet(CustomPagesWithUnseenOptions,PageID) do
+            PageID:=PageID+1;
+    end;
+    Result:=(PageID=PageIDBeforeInstall);
+end;
+
+procedure AdjustNextButtonLabel(Sender:TObject);
+begin
+    if (CurrentCustomPageID=ProcessesPage.ID) then
+        WizardForm.NextButton.Caption:=SetupMessage(msgButtonInstall)
+    else if IsLastPageBeforeInstall(CurrentCustomPageID) then begin
+        RefreshProcessList(NIL);
+        if GetArrayLength(Processes)=0 then
+            WizardForm.NextButton.Caption:=SetupMessage(msgButtonInstall)
+        else
+            WizardForm.NextButton.Caption:=SetupMessage(msgButtonNext);
+    end else
+        WizardForm.NextButton.Caption:=SetupMessage(msgButtonNext);
+end;
 
 function InitializeSetup:Boolean;
 var
@@ -1110,6 +1136,8 @@ function CreatePage(var PrevPageID:Integer;const Caption,Description:String;var 
 begin
     Result:=CreateCustomPage(PrevPageID,Caption,Description);
     CurrentCustomPageID:=Result.ID;
+    if (FirstCustomPageID=0) then
+        FirstCustomPageID:=Result.ID;
     PrevPageID:=Result.ID;
     TabOrder:=0;
     Top:=8;
@@ -1120,6 +1148,8 @@ function CreateFilePage(var PrevPageID:Integer;const Caption,Description,SubCapt
 begin
     Result:=CreateInputFilePage(PrevPageID,Caption,Description,SubCaption);
     CurrentCustomPageID:=Result.ID;
+    if (FirstCustomPageID=0) then
+        FirstCustomPageID:=Result.ID;
     PrevPageID:=Result.ID;
     TabOrder:=0;
     Top:=8;
@@ -1574,6 +1604,7 @@ begin
         Width:=GetTextWidth(Caption,Font)+20; // 20 is the estimated width of the checkbox itself
         Left:=WizardForm.BackButton.Left-Width-(WizardForm.CancelButton.Left-WizardForm.NextButton.Left-WizardForm.NextButton.Width);
         Checked:=IsUpgrade(ExpandConstant('{#APP_VERSION}'),PreviousGitForWindowsVersion);
+        OnClick:=@AdjustNextButtonLabel;
         Height:=WizardForm.CancelButton.Height;
         Top:=WizardForm.CancelButton.Top;
     end;
@@ -2091,6 +2122,7 @@ end;
 
 procedure CurPageChanged(CurPageID:Integer);
 begin
+    CurrentCustomPageID:=CurPageID;
     if CurPageID=wpInfoBefore then begin
         if WizardForm.NextButton.Enabled then begin
             // By default, do not show a blinking cursor for InfoBeforeFile.
@@ -2102,10 +2134,6 @@ begin
             // This will be checked later again when the user clicks "Next".
             WizardForm.DirEdit.Text:=ExpandConstant('{userpf}\{#APP_NAME}');
         end;
-    end else if CurPageID=PageIDBeforeInstall then begin
-        RefreshProcessList(NIL);
-        if GetArrayLength(Processes)=0 then
-            WizardForm.NextButton.Caption:=SetupMessage(msgButtonInstall);
     end else if (ProcessesPage<>NIL) and (CurPageID=ProcessesPage.ID) then begin
         // Show the "Refresh" button only on the processes page.
         ProcessesRefresh.Show;
@@ -2113,6 +2141,7 @@ begin
     end else begin
         ProcessesRefresh.Hide;
     end;
+    AdjustNextButtonLabel(Nil);
 end;
 
 function NextButtonClick(CurPageID:Integer):Boolean;
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -304,10 +304,6 @@ begin
 end;
 
 const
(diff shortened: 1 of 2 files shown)


dscho:
Great, thank you for working on this!

>   Signed-off-by: Denise Draper <draperd@acm.org>

@denised could you add that to the commit message, please? It should be as easy as `git commit --amend -s`, then force-pushing.

installer/install.iss
```diff
@@ -375,6 +375,11 @@ var
     // Previous Git for Windows version (if upgrading)
     PreviousGitForWindowsVersion:String;
 
+    // List of pages that _only_ have options whose value has been already set (by command line or previous install)
+    PagesPreSet:String;
+    // Tracks whether all options on a page have been pre-set
+    OptionsArePreSet:Boolean;
```
dscho: If you revert the logic, i.e. if you track the pages that have new options that have not previously been seen by the user (the variable name `PagesWithPreviouslyUnseenOptions` comes to mind), then you need only one variable, and you also need less code changes.

installer/install.iss
```diff
@@ -974,9 +979,13 @@ begin
         i:=InferredDefaultKeys.IndexOf(Key); // cannot use .Find because the list is not sorted
         if (i>=0) then
             Result:=InferredDefaultValues[i]
-        else
+        else begin
             // Restore the settings chosen during a previous install.
             Result:=GetPreviousData(Key,Default);
+            // Test to see if we really set this to the Default value
+            if ('z'+Default=GetPreviousData(Key,'z'+Default)) then
```
dscho: If -- for whatever abstruse reason -- the previously recorded choice happens to be identical to `'z'+Default`, this condition does not do what we want.

I really think that we need a different condition here. Let's see. We need to call `GetPreviousData()` again, that's for sure. We cannot change the `Key` parameter, because we want to see whether there has been any pre-recorded value. So we have to play with the second parameter, i.e. with the default value to use if no value has been recorded in previous installs. Therefore we need to use a default value there that must be different from whatever the `GetPreviousData()` call in line 984 could possibly have returned. So if we modify the value that was actually returned, e.g. `'z'+Result`, we have _exactly_ what we need.

Therefore, I think the condition you want is actually:

```pascal
if ('z'+Result=GetPreviousData(Key,'z'+Result)) then
```

Now, I still think that it would be much easier to reason about if we looked only for the pages that we want to show, rather than for the options that we do not need to show because the user saw them already.

In other words, I would suggest something like this instead:

```pascal
if (Pos(','+PrevPageID+',',PagesWithPreviouslyUnseenOptions)=0) then begin
    if (PagesWithPreviouslyUnseenOptions='') then
        PagesWithPreviouslyUnseenOptions:=',';
    PagesWithPreviouslyUnseenOptions:=PagesWithPreviouslyUnseenOptions+IntToStr(PrevPageID)+',';
end
```

installer/install.iss
```diff
@@ -1946,6 +1978,8 @@ begin
     // Restore the settings chosen during a previous install
     RdbExperimentalOptions[GP_BuiltinAddI].Checked:=ReplayChoice('Enable Builtin Interactive Add','Auto')='Enabled';
 #endif
+    if OptionsArePreSet then
+        PagesPreSet:=PagesPreSet+IntToStr(PrevPageID)+',';
```
dscho: While this certainly works, it is an _awful_ lot of repeated code, and even worse: it is all too easy to forget when adding new pages (or to forget to remove the code when removing pages!).

Besides, I could imagine that it does not even do what we need it to do if there are no experimental built-ins, i.e. if that page is not even shown.

And even if it does what we need it to do, it is not obvious from looking at this code change.

So really, I would like to go the other way round: accumulate the IDs of the pages that we want to show. (And now that I think about it, maybe an even better name than `PagesWithPreviouslyUnseenOptions` would be `PagesToShow`.)

installer/install.iss
```diff
@@ -2024,7 +2058,8 @@ begin
         RefreshProcessList(NIL);
         Result:=(GetArrayLength(Processes)=0);
     end else begin
-        Result:=False;
+        // skip pages that are entirely preset
+        Result:=(Pos(','+IntToStr(PageID)+',',PagesPreSet)>0);
```
dscho: Whoa. This change would _always_ skip the options pages when upgrading, not only in some sort of "easy mode"...

I think we _have_ to guard this behind an initial page that lets the user choose between the easy and the advanced installation.

Further, I am not entirely sure that all of our pages are handled correctly: at least the component page, and probably also the page that lets the user choose whether to run Git Bash and whether to show the release notes, would probably be skipped by this patch? (I don't know, I did not build an installer with this patch, did you?)

denised:
Let me do a meta-comment that addresses several of your comments and the rationale for the way I coded it.    I'm open to another solution, but I want to be sure this rationale is understood:

There are two kinds of pages put out by the installer: built in pages like the components page, and custom pages that are all created within the `InitializeWizard` procedure.  The code changes I have made affect _only_ the custom pages.   

Currently, the standard pages that can be displayed are the Licensing page, the Installation Location page, the Components Page and the Start Menu location page.  By experimentation, I see that it is already the case that the installer does not show the Installation Location or Start Menu Location pages on upgrade.  It only shows the License and components pages.  I did not investigate how or why what is the case, but point it out that there is precedence for upgrade being a simpler process.

The existence of standard pages, not created in code, is also why I chose to make the list an affirmative list of pages to be skipped.  If there were instead a list of pages to be shown, then I would have to be sure that list contained the ID of every page, including the standard pages that are created only by configuration.  No matter how that list was initialized, it would create a point in the code that anyone modifying the standard pages would have to remember to check and possibly update.

Conversely, the fact that all the custom pages are created within one procedure in install.iss makes it fairly easy for a future developer to see the pattern and follow it if they want to add another page to the installer.

Finally, if someone does by chance omit the code to skip a page, the result is to show the user a page they didn't have to see.  This is less problematic than the other way around, which is to *not* show the user a page that they should see.   In summary, I think the direction of the logic I chose for determining if a page should be shown is the safer of the two alternatives.

The usability question is whether to show a 'Fast Install' option or simply to do a fast install in the case of upgrade.  That is a matter of taste, I think.  Personally, I cannot imagine a situation in which a user would want to revisit the installation options when doing a routine upgrade, so I find this solution preferable.  But if you feel strongly that it should be the other way, I don't object, and it should be possible to do.

BTW, I tested various combinations of this patched version and the existing installer on a VM to validate the behavior, particularly distinguishing between upgrade and uninstall/reinstall, to make sure it behaved as expected.

installer/install.iss
denised: > Further, I am not entirely sure that all of our pages are handled correctly: at least the component page, and probably also the page that lets the user choose whether to run Git Bash and whether to show the release notes, would probably be skipped by this patch? (I don't know, I did not build an installer with this patch, did you?)

I left the built-in pages with no changes, which causes the license and components pages to be shown on upgrade.  We could figure out how to skip these pages as well if you would prefer that.  [Aside: One thing I notice is that many of the items on the components page seem more like options than components, and it might make sense to have them as custom options page(s) instead?]

installer/install.iss
denised: >Besides, I could imagine that it does not even do what we need it to do if there are no experimental built-ins, i.e. if that page is not even shown.

It does behave correctly in this case, because all it will do is add a (in this case useless) ID to the pages-to-skip list.   Note there is also an outer #ifdef that controls whether the experimental page is to be shown at all, and the code changes are inside that outer #ifdef, meaning that if no experimental page, then this code won't run either.

installer/install.iss
denised: The `'z'+Result` fix is a good idea; I will change that.

dscho:
> Let me do a meta-comment that addresses several of your comments and the rationale for the way I coded it.

Thank you for this thoughtful and detailed write-up.

> The existence of standard pages, not created in code, is also why I chose to make the list an affirmative list of pages to be skipped.

Let me add my rationale for doing the reverse: to me, "add to skip" is a contradiction in itself. To me, it would make a lot more sense to add pages to the list to be shown.

As to the conundrum about internal pages: you're right, I overlooked that. But that is also easily solved: In the very same function where I'd like to build the list of pages to be shown (`GetPreviousData()`), I would now also like to build the list of custom pages. That way, we can test very clearly: is this a custom page (i.e. a candidate for skipping)? If yes, does it have previously unseen choices?

The reason why I harp on this is that following the Don't Repeat Yourself principle has saved us time and time again from maintenance hassle.

> inally, if someone does by chance omit the code to skip a page, the result is to show the user a page they didn't have to see. This is less problematic than the other way around, which is to _not_ show the user a page that they should see.

I agree that it is less problematic, but not by much. Users would complain, rightfully, and very very loudly, that this piece of *** is so broken it shows some pages and some not and it is inconsistent. I'd rather not even hear those voices, I've had my share.

The better idea is to make it easier to get future changes right. And that means that we should avoid forcing future changes to repeat the same statements (or very, very similar ones) that we already added in plenty of places.

Not repeating yourself has a further benefit: it makes it easier for reviewers not to miss something obvious, such as a forgotten place to add those repeated statements.

And as I write this, I realize that I repeat myself quite often here. I'd rather not, but it is an important point: Don't Repeat Yourself in your code. It won't end well, ever.

> Personally, I cannot imagine a situation in which a user would want to revisit the installation options when doing a routine upgrade, so I find this solution preferable.

Sure, during an upgrade you might want to skip those options, that's a nice thing to have.

But we do suggest, frequently, to reinstall Git for Windows, to change options. For example, if a user wants to use some command-line program that would require `winpty` in the Git Bash, and that does not actually play well with `winpty`, then we would suggest for that user to re-install Git for Windows and choosing the default Windows Console instead of MinTTY.

We must not break that use case.

Of course, it is a rarer use case, so I agree that we should probably skip the already-seen options by default.

BTW there is one particular scenario that is still not addressed, not even by ideas. Suppose that Git for Windows adds support for Yet Another Editor, and users are eager to choose that option. In the current state of this PR, they would not get that opportunity, *ever*. Any ideas how to address that?

installer/install.iss
dscho: > I left the built-in pages with no changes, which causes the license and components pages to be shown on upgrade. We could figure out how to skip these pages as well if you would prefer that.

I think that would actually make a lot of sense. The license has not changed, the user accepted it already (by virtue of installing the software), and the same thing for options.

Except for that hard problem of new options...

> [Aside: One thing I notice is that many of the items on the components page seem more like options than components, and it might make sense to have them as custom options page(s) instead?]

I think you're right. The Git LFS option looks somewhat like it is a component, although I wouldn't mind turning that into a proper option.

installer/install.iss
dscho: > It does behave correctly in this case, because all it will do is add a (in this case useless) ID to the pages-to-skip list.

Ah, that's right. Still: harder to reason about that I'd bargain for.

> Note there is also an outer #ifdef that controls whether the experimental page is to be shown at all, and the code changes are inside that outer #ifdef, meaning that if no experimental page, then this code won't run either.

Valid point, too. Playing even more into my argument that it is tricky to confirm the correctness of this patch. I'd prefer a patch that is super easy to confirm, if possible.

denised:
Thank you for your thoughtful reply :-)
I am willing to concede the point and switch the nature of the logic. I didn’t like ‘add to skip’ either, (_except that it matches the logic of the ShouldSkip procedure_), and agree that ‘should show’ is easier to understand.

Also, I was unaware of the recommendations to re-install GfW as a means to access options; that does change priorities. You could achieve the same effect by uninstalling and re-installing (which would take care of your editor case, BTW), but that would also forget all your prior settings, which isn’t what you want if your goal is to change a single setting. So yes, switching to a fast-path / complete path installation option is the better way to go.

-- 
 Denise Draper
 draperd@acm.org

denised:
And BTW, the nature of the changes proposed is extensive enough that it is easier to do a new pull request than to amend this one. If you agree, close this one, and I will create a new one.
-- 
 Denise Draper
 draperd@acm.org

dscho:
> So yes, switching to a fast-path / complete path installation option is the better way to go.

Right, and if we detect that the same version (or a downgrade), then we don't skip a thing.

> the nature of the changes proposed is extensive enough that it is easier to do a new pull request than to amend this one

I agree that the changes are extensive, but the goal remains the same, and it builds on top of the conversation here. So I'd much prefer amending over opening a separate (seemingly unrelated) PR.

denised:
I have completely rewritten this feature following the discussion above.  There is now an additional page, which shows only during upgrade, allowing the user to select whether to skip options or not.
As before, if there is a _new_ option that the user has not previously selected a value for, that option page will be shown even if the user selects 'skip options'.

I have tested upgrade and re-install scenarios, and also temporarily added a new option just to make sure that logic worked.

Some comments/observations:

- I have still left the original logic that suppresses two built-in installer pages: setting the installation directory and start menu directory.  As it stands, these pages are only shown for an initial installation (no git-for-windows installed).  I think this is reasonable, but if you really wanted consistency, you might also want these pages to be shown if the user requests 'review all options.'
- The built-in components page is always shown.  This is by design from the inno install perspective, but given that most of the things on the components page are actually installation options, it may be confusing from a user's perspective.  The fix would be to move the components 'options' to custom options pages similar to the other options.
- It bothers me a little bit that when you skip options, the last 'Next' button jumps straight into installation, whereas when you review options, you get an 'Install' button.  Inno Setup supports a 'pre-install' page that gives the user a stopping point before jumping into the actual install.  That page can be used to show all the selected options.  I think this might be a good idea to add (especially since it will allow a user to quickly review all the options they are implicitly agreeing to.)
- @dscho previously brought up a corner case: what if a new editor is added to the list?  With the current code this will not be caught as a 'new option', however re-installing or selecting 'review all options' will easily permit a user to get select the new editor.

denised:
Hmm... I left a big comment just a few minutes ago, and it seems to have disappeared.  If this is a duplicate, you can skip reading it now...

Commit 8c84615 is a complete rewrite of the feature following the discussion above.  The installer now shows a page (only during upgrade), allowing a user to chose to skip options or review all options.  The options pages are shown or hidden accordingly.  As before, if there is a _new_ option (one that the user has not previously selected a value for), that options page will be shown even if the user selected skip options.

I have tested upgrade and re-install scenarios, and also (temporarily) added a new option just to make sure that part of the logic worked as well.

PhilipOakley:
>  I left a big comment just a few minutes ago, and it seems to have disappeared.

It did come through on email, and is currently showing on this issue page #270 with your comments marked as "1 hour ago", so looks OK.

installer/install.iss
```diff
@@ -899,6 +911,27 @@ begin
...
+    Result:=(not (CurrentVersion=PreviousVersion) and not IsDowngrade(CurrentVersion,PreviousVersion))
+end;
+
+{ Represent a set as a string of comma-separated values }
+function IsInSet(ASet:String;Value:Integer):Boolean;
+begin
+    Result:=(Pos(','+IntToStr(Value)+',',ASet)>0);
+end;
+
+function AddToSet(ASet:String;Value:Integer):String;
```
dscho: How about making this a `procedure` and turning `ASet` into a `var` (i.e. passing it by reference rather than by value)?

installer/install.iss
```diff
@@ -1524,6 +1564,19 @@ begin
 
     PrevPageID:=wpSelectProgramGroup;
 
+    (*
+     * See if this is an upgrade, and if so, create a custom page to allow the user to select whether or not to do a fast install.
+     *)
+    CurrentVersion:=ExpandConstant('{#APP_VERSION}');
+    if ISUpgrade(CurrentVersion,PreviousGitForWindowsVersion) then begin
```
dscho: Please use the same case as in the function declaration: `IsUpgrade` (the `s` should be lower-case).

dscho:
>     * It bothers me a little bit that when you skip options, the last 'Next' button jumps straight into installation, whereas when you review options, you get an 'Install' button

You could imitate [the existing code](https://github.com/git-for-windows/build-extra/blob/a9f7367db16d1f2867f4cd2fb0e191350f385ae0/installer/install.iss#L2051) that sets that label to <kbd>Install</kbd> on the last page when no processes have to be shown.

Or maybe we can adjust `PageIDBeforeInstall` to reflect the last page that is not skipped? One way to do that would be to have a loop that tests whether `PageIDBeforeInstall` would be skipped, and if so, automatically goes back by one. Probably similar to this completely untested code:

```pascal
while (PageIDBeforeInstall>PathPage.ID) and
         IsInSet(CustomOptionsPages, PageIDBeforeInstall) and
        not IsInSet(CustomOptionsPagesWithNewValues,PageIDBeforeInstall) do
    PageIDBeforeInstall:=PageIDBeforeInstall-1;
```

denised:
> > * It bothers me a little bit that when you skip options, the last 'Next' button jumps straight into installation, whereas when you review options, you get an 'Install' button
> 
> You could imitate [the existing code](https://github.com/git-for-windows/build-extra/blob/a9f7367db16d1f2867f4cd2fb0e191350f385ae0/installer/install.iss#L2051) that sets that label to Install on the last page when no processes have to be shown.
> 
> Or maybe we can adjust `PageIDBeforeInstall` to reflect the last page that is not skipped? One way to do that would be to have a loop that tests whether `PageIDBeforeInstall` would be skipped, and if so, automatically goes back by one. Probably similar to this completely untested code:
> 
> ```pascal
> while (PageIDBeforeInstall>PathPage.ID) and
>          IsInSet(CustomOptionsPages, PageIDBeforeInstall) and
>         not IsInSet(CustomOptionsPagesWithNewValues,PageIDBeforeInstall) do
>     PageIDBeforeInstall:=PageIDBeforeInstall-1;
> ```

I don't think these would work because the last page may very well be the page asking the user if they skip or full options, and I don't think it would be a good idea to change the button interactively :-)   If you are not comfortable with a separate 'pre-install' page, then I think the current design should be left alone.

denised:
> Would you mind squashing the commits into a single one? I do not think that it would make sense to keep the earlier attempt in the commit history.
> 
> If you want, though, I would not mind having the change separated out into its own commit where `ReplayChoice()` is modified to take the page ID as a function parameter.
> 
> Or could we use `PrevPageID` instead of having to change every `ReplayChoice()` call?

Whoops!  I did the squash before reading your full comment.
PrevPageID is not global currently, which is why I did not use it that way.  We could make it global, but then it really should have a different name (since it actually refers to the _current_ page, not the previous one), ... but that would then contradict the naming convention used in the Inno Setup samples that uses the variable PrevPageID in exactly the way it is used in this code.  In my view, passing the ID explicitly into ReplayChoice is a reasonable thing to do anyway, considering we are asking it to perform this page-accumulation as a side effect.

Separating the code into two commits is possible if you think it important.

dscho:
> PrevPageID is not global currently

That's a good point which I had missed. (And yes, that means I would suggest to split out the refactoring as a preparatory step, as it makes the commit topology clearer as well as easier to review.)

Yet another idea occurred to me after sleeping over it: other installers have a <kbd>Next</kbd> _and_ an <kbd>Install</kbd> button, where the latter just skips the rest of the wizard pages.

And a variation of that idea would be to add a checkbox to the bottom, adjacent to the <kbd>Next</kbd> button. This checkbox would toggle whether or not to skip wizard pages with only already-seen options. The only question would be how to label it (it should have a succinct, yet intuitive label).

What do you think?

denised:
> PrevPageID is not global currently
> 
> That's a good point which I had missed. (And yes, that means I would suggest to split out the refactoring as a preparatory step, as it makes the commit topology clearer as well as easier to review.)
> Yet another idea occurred to me after sleeping over it: other installers have a Next and an Install button, where the latter just skips the rest of the wizard pages.
> And a variation of that idea would be to add a checkbox to the bottom, adjacent to the Next button. This checkbox would toggle whether or not to skip wizard pages with only already-seen options. The only question would be how to label it (it should have a succinct, yet intuitive label).
> What do you think?

I thought of this approach, but also didn't know how to label the button properly :-)
I think it is okay to accept this design as is, and if someone later has a better inspiration, they can make that a separate change.

I have separated the commits into two.  Unless there is anything more, I think this pull request is complete.

denised:
Ack!  I hit the close button by accident!

dscho:
I haven't forgotten about this PR, promise! Life just got in the way, and I still want to sleep a couple nights over the question whether we can introduce a checkbox with a nice label to skip ahead, as well as whether there is possibly a good way to point out new options (such as a new editor) while upgrading.

denised:
And have you come to any conclusions? :-)

Personally I cannot think of a way to do either of these things where the code impact does not far outweigh the potential benefit.

-- 
 Denise Draper
 draperd@acm.org

dscho:
> And have you come to any conclusions? :-)

Yes, I have.

In light of [the download statistics](https://www.somsubhra.com/github-release-stats/?username=git-for-windows&repository=git) (three quarters of a million, slightly less than a week after it has been released), I do think that it makes sense to put in an extra effort to make this thing easy and pleasant to use.

> Personally I cannot think of a way to do either of these things where the code impact does not far outweigh the potential benefit.

Please have a look at the changes I added (either look at the [interdiff](https://github.com/git-for-windows/build-extra/compare/b05c9d9bd45a0c77f84e5a0e84ddb966f02c7d3a..5e5651029dbcb399c2b885568ac9b41f5decce62), or at [the individual commits](https://github.com/git-for-windows/build-extra/pull/270/commits) -- including your initial commits which I fixed up a little).

The short version of it is that an upgrade will now ideally look similar to this:

image

Note that:
1. there is a checkbox at the bottom,
2. it is checked by default for upgrades, and
3. the label of the <kbd>Next ></kbd> button was changed to <kbd>Install</kbd>.

All the user has to do is to hit <kbd>Enter</kbd>, and not a dozen times, but only once. If they so desire, they can uncheck that box (even with the convenient keyboard shortcut <kbd>Ctrl</kbd>+<kbd>O</kbd>) and walk through all the pages at their own leisure.

denised:
Very nice!

-- 
 Denise Draper
 draperd@acm.org

linquize:
When "Only show new options" is checked, it will automatically show **release notes** at the end of installation.

dscho:
> When "Only show new options" is checked, it will automatically show **release notes** at the end of installation, skipping "Completing the Git Setup Wizard" page.

Yes. This is an option you already saw (but your choice has not been recorded) ;-)

dscho:
@linquize if you think that the last page should be shown after a successful installation even if that checkbox is checked, please open a PR to make it so.

PR #271, opened 2019-12-06
Enable developer mode checking for standard symbolic-link checkbox

Kristof-Mattei:
- [x] Figure out a better way to auto-check the box
- [x] Figure out how to compile this
- Test!
  - Old install
    - [x] Retain settings. Notice that this is untouched, and we may want to opt-in users, or try to opt-in. See https://github.com/git-for-windows/build-extra/blob/1defc6d25ea750a806d180c7f0222162edcbb003/installer/install.iss#L1915 I'm suggesting we bump the 2.14.1 to the current version. 
  - New install
     - Admin
       - [x] Developer mode enabled, checkbox checked 
       - [x] Developer mode disabled, checkbox unchecked
     - Non-admin
       - [x] Developer mode enabled, checkbox checked 
         - Developer mode disabled
         - [x] User CAN create symlink (i.e. has `SecreateSymbolicLinkPrivilege`), checkbox checked 
         - [x] User CANNOT create symlink, checkbox unchecked

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1048,12 +1048,29 @@ end;
 
 function EnableSymlinksByDefault():Boolean;
 var
+    SymlinksForRegularUsers:Cardinal;
+    Root:Integer;
     ResultCode:Integer;
+    Version:TWindowsVersion;
 begin
+
+    RegQueryDwordValue(HKEY_LOCAL_MACHINE, 'SOFTWARE\Microsoft\Windows\CurrentVersion\AppModelUnlock', 'AllowDevelopmentWithoutDevLicense', SymlinksForRegularUsers)
+    GetWindowsVersionEx(Version);
+
+    if ((SymlinksForRegularUsers=1) and (Version.Build>=14972)) then begin
+        // awesome, Developer mode enabled, and we're in Creators update, we can create symlinks w/o being admin
+        Log('Symbolic links enabled, machine has Developer Mode enabled');
+        Result:=True
+        Exit;
+    end;
+
     if IsOriginalUserAdmin then begin
+        // detection only works when we're not running as admin
         Log('Symbolic link permission detection failed: running as admin');
         Result:=False;
     end else begin
+        // maybe rights assigned through group policy without enabling developer mode?
+        // let's test by creating a symbolic link
         ExecAsOriginalUser(ExpandConstant('{cmd}'),ExpandConstant('/c mklink /d "{tmp}\symbolic link" "{tmp}" >"{tmp}\symlink test.txt"'),'',SW_HIDE,ewWaitUntilTerminated,ResultCode);
         Result:=DirExists(ExpandConstant('{tmp}\symbolic link'));
     end;


installer/install.iss
```diff
@@ -1048,14 +1048,27 @@ end;
...
     ResultCode:Integer;
 begin
-    if IsOriginalUserAdmin then begin
-        Log('Symbolic link permission detection failed: running as admin');
-        Result:=False;
+    RegQueryDwordValue(HKEY_LOCAL_MACHINE, 'SOFTWARE\Microsoft\Windows\CurrentVersion\AppModelUnlock', 'AllowDevelopmentWithoutDevLicense', SymlinksForRegularUsers)
+
+    if (SymlinksForRegularUsers = 1) then begin
+        // awesome, Developer mode enabled, we can create symlinks w/o being admin
+        Log('Symbolic links enabled, machine has Developer Mode enabled');
```
dscho: I think we also need to verify that we're running on a "new enough" Windows 10: [Developer Mode only started allowing non-administrators to create symbolic links with Creators Update](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/).

We should probably use [`GetWindowsVersionEx()`](http://www.jrsoftware.org/ishelp/index.php?topic=isxfunc_GetWindowsVersionEx) and then verify the Build:

```pascal
    Version:TWindowsVersion;
[...]
GetWindowsVersionEx(Version);
if ((SymlinksForRegularUsers=1) and (Version.Build>=14972)) then begin
[...]
```

installer/install.iss
```diff
@@ -1048,14 +1048,27 @@ end;
 
 function EnableSymlinksByDefault():Boolean;
 var
+    SymlinksForRegularUsers:Cardinal;
+    Root:Integer;
     ResultCode:Integer;
 begin
-    if IsOriginalUserAdmin then begin
-        Log('Symbolic link permission detection failed: running as admin');
-        Result:=False;
```
dscho: How about keeping the indentation intact by prefixing this code with the new code? You can use `Exit;` to return from the function early. Something like this:

```pascal
if [... detect developer mode ...] then begin
    [... log it ...]
    Result:=True;
    Exit;
end;
```

That will avoid the extra indentation, and will make reviewing the code easier (and as a consequence, it will make it harder for bugs to hide).

installer/install.iss
```diff
@@ -1048,14 +1048,27 @@ end;
```
dscho: The commit message is a bit cryptic because it is so short. Also, `build-extra` is about more than just the installer, therefore I would ask to prefix the commit message with the "area code" `installer:`. Something like this:

```
installer: make symlink support detection more robust

When running the installer from an administrator account, currently the
detection fails whether regular users can create symbolic links.

Fix this by detecting whether running in Windows 10 Creators Update or
later in Developer Mode, falling back to the previous method.
```

dscho:
> @dscho thanks for the feedback. I’ll implement them ASAP.

Thank you @Kristof-Mattei!

installer/install.iss
Kristof-Mattei: What's the policy on putting in fixes for unsupported versions of Windows? 

The first public release after `14972` is Windows 10 (1703) with build number 10.0.15063, and support for that one expired on October 8, 2019 for Enterprise (October 8, 2018 for regular editions), i.e. there's no supported Windows 10 version out there that doesn't have symlinks with Developer Mode. 

Happy to put in the fix if that's policy.

installer/install.iss
Kristof-Mattei: Fixed.

Kristof-Mattei:
I'm going to do a full test, and then squash the commit in to 1 with a meaningful message.

Kristof-Mattei:
@dscho I believe I have addressed your feedback & concerns. 

I have 2 questions: 

1) dos Git For Windows support unsupported versions of Windows: https://github.com/git-for-windows/build-extra/pull/271#discussion_r355085286
2) do we want to bump the number at: https://github.com/git-for-windows/build-extra/blob/1defc6d25ea750a806d180c7f0222162edcbb003/installer/install.iss#L1915, the 2.14.1, because now we can support more?

I have completed retesting the modified code in a VM testing all cases. 

Let me know!

installer/install.iss
dscho: At this point in the code, I believe it is uninteresting what Windows versions Git for Windows supports, and much more interesting which Windows build introduced the behavior on which your patch relies. Therefore using 14972 is the correct value to use here.

dscho:
>     1. dos Git For Windows support unsupported versions of Windows: [#271 (comment)](https://github.com/git-for-windows/build-extra/pull/271#discussion_r355085286)

Git for Windows tries to support the widest range of Windows versions as can be done with a reasonable amount of effort. For example, Git for Windows _still_ supports Windows Vista.

>     2\. do we want to bump the number at: https://github.com/git-for-windows/build-extra/blob/1defc6d25ea750a806d180c7f0222162edcbb003/installer/install.iss#L1915

It looks like https://github.com/git-for-windows/build-extra/commit/360e19d3d90041aff70bb95756a806fd34828cb4 introduced this line, and I think the intention was to not only change the default but essentially to ignore the previous choice if it matched the previous default (because the check was probably chosen incorrectly).

If you want to bump that, under the assumption that symlinks were disabled for many users by mistake, we can do that.

But I am not sure that many Git for Windows installations have symlinks turned off by mistake. What is your scenario that needed fixing? I have no dependable statistics to rely on, therefore we really need some sort of really convincing argument to override previous choices when upgrading Git for Windows.

Kristof-Mattei:
Gotcha on the support.

In the case of a previously unselected checkbox, and then updating to 'this' version: 
Someone who is an Admin and has Developer Mode enabled (after Creators Update) now gets the box correctly ticked when they're upgrading their Git installation. For free. 
It will not hurt the user's configuration.

dscho:
Thank you for your contribution, and sorry for the delay in merging it (I was a bit busy with the security fix release).

PR #272, opened 2019-12-12
Let `sdk create-desktop-icon` use the correct API to determine the Desktop's location

Kristof-Mattei:
Given OneDrive with the Desktop Redirection feature, and other supported methods to NOT have your Desktop in %USERPROFILE%\Desktop, we should use a more sophisticated method to get the path to the Desktop in order to put a shortcut there. Since we are using Bash we need to step out and invoke a vbs script to get the SpecialFolder called Desktop, and use that as a path.

### DIFF
--- a/git-extra/git-sdk.sh
+++ b/git-extra/git-sdk.sh
@@ -65,14 +65,14 @@ sdk () {
 		'') break;;
 		-*) sdk die "Unknown option: %s\n" "$1"; return 1;;
 		esac; do shift; done &&
-
 		case "$(uname -m)" in
 		i686) bitness=" 32-bit";;
 		x86_64) bitness=" 64-bit";;
 		*) bitness=;;
 		esac &&
-		desktop_icon_path="$(reg query 'HKCU\Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders' //v Desktop |
-			sed -ne 'y/\\/\//' -e 's/.*REG_SZ *//p')/Git SDK$bitness.lnk" &&
+		desktop_icon_path="Git SDK$bitness.lnk" &&
+		desktop_icon_path="$(create-shortcut.exe -n --desktop-shortcut /git-bash.exe "$desktop_icon_path" |
+			sed -n 's/^destination: //p')" &&
 		if test -n "$force" || test ! -f "$desktop_icon_path"
 		then
 			create-shortcut.exe --icon-file /msys2.ico --work-dir \
--- a/git-extra/create-shortcut.c
+++ b/git-extra/create-shortcut.c
@@ -25,7 +25,7 @@ int main(int argc, char **argv)
 	const char *progname = argv[0];
 	const char *work_dir = NULL, *arguments = NULL, *icon_file = NULL;
 	const char *description = NULL;
-	int show_cmd = 1;
+	int show_cmd = 1, desktop_shortcut = 0;
 
 	static WCHAR wsz[1024];
 	HRESULT hres;
@@ -45,7 +45,12 @@ int main(int argc, char **argv)
 			icon_file = argv[2];
 		else if (!strcmp(argv[1], "--description"))
 			description = argv[2];
-		else {
+		else if (!strcmp(argv[1], "--desktop-shortcut")) {
+			desktop_shortcut = 1;
+			argc--;
+			argv++;
+			continue;
+		} else {
 			fprintf(stderr, "Unknown option: %s\n", argv[1]);
 			return 1;
 		}
@@ -59,7 +64,7 @@ int main(int argc, char **argv)
 		argv++;
 	}
 
-	if (argc < 3) {
+	if (argc != 3) {
 		fprintf(stderr, "Usage: %s [options] <source> <destination>\n",
 			progname);
 		return 1;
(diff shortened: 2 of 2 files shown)


git-extra/git-sdk.sh
```diff
@@ -65,13 +65,22 @@ sdk () {
...
 		case "$(uname -m)" in
 		i686) bitness=" 32-bit";;
 		x86_64) bitness=" 64-bit";;
 		*) bitness=;;
 		esac &&
-		desktop_icon_path="$USERPROFILE/Desktop/Git SDK$bitness.lnk" &&
+		desktop_icon_path="$path_to_desktop/Git SDK$bitness.lnk" &&
 		if test -n "$force" || test ! -f "$desktop_icon_path"
 		then
 			create-shortcut.exe --icon-file /msys2.ico --work-dir \
```
dscho: We already use the `create-shortcut.exe` that is compiled in the very same package... Would it not make more sense to add an option to that executable that tells it to interpret the path as relative to the Desktop, and then use [`SHGetKnownFolderPath()`](https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath) directly in C?

That would avoid requiring Visual Basic Script, and make `create-shortcut.exe` more useful.

Kristof-Mattei: That would make the `create-shortcut` more specialized. 

At the moment it just creates a shortcut to a location, with a workdir and an icon.

There's no  way in the Windows API do something like this:

Expand `shell:desktop\git-bash.lnk` do `C:\Users\KM\OneDrive\Desktop\git-bash.lnk`

We can fix it by offering a full-path option (which is what's there now) and amend so that it would take a `KNOWNFOLDERID` and a relative location.

dscho: What I meant was `create-shortcut.exe --desktop-shortcut [...] "Git SDK$bitness.lnk"`.

dscho: ... or more precisely, what I meant was this:

```diff
diff --git a/git-extra/create-shortcut.c b/git-extra/create-shortcut.c
index 81ebce9fd..c0472e2f1 100644
--- a/git-extra/create-shortcut.c
+++ b/git-extra/create-shortcut.c
@@ -25,7 +25,7 @@ int main(int argc, char **argv)
 	const char *progname = argv[0];
 	const char *work_dir = NULL, *arguments = NULL, *icon_file = NULL;
 	const char *description = NULL;
-	int show_cmd = 1;
+	int show_cmd = 1, desktop_shortcut = 0;
 
 	static WCHAR wsz[1024];
 	HRESULT hres;
@@ -45,7 +45,12 @@ int main(int argc, char **argv)
 			icon_file = argv[2];
 		else if (!strcmp(argv[1], "--description"))
 			description = argv[2];
-		else {
+		else if (!strcmp(argv[1], "--desktop-shortcut")) {
+			desktop_shortcut = 1;
+			argc--;
+			argv++;
+			continue;
+		} else {
 			fprintf(stderr, "Unknown option: %s\n", argv[1]);
 			return 1;
 		}
@@ -92,9 +97,30 @@ int main(int argc, char **argv)
 	if (description)
 		psl->lpVtbl->SetDescription(psl, description);
 
-	wsz[0] = 0;
-	MultiByteToWideChar(CP_ACP,
-			0, argv[2], -1, wsz, 1024);
+	if (!desktop_shortcut)
+		wsz[0] = 0;
+	else {
+		PWSTR p;
+
+		hres = SHGetKnownFolderPath(&FOLDERID_Desktop,
+					    KF_FLAG_DONT_UNEXPAND, NULL, &p);
+		check_hres(hres, "Could not get desktop path");
+
+		desktop_shortcut = wcslen(p);
+		if (desktop_shortcut + 2 + strlen(argv[2]) * 3 >
+		    sizeof(wsz) / sizeof(WCHAR)) {
+			fwprintf(stderr,
+				 L"Error: Too long Desktop path: %s\n", p);
+			exit(1);
+		}
+		memcpy(wsz, p, sizeof(WCHAR) * desktop_shortcut);
+		wsz[desktop_shortcut++] = L'\\';
+		wsz[desktop_shortcut] = L'\0';
+		CoTaskMemFree(p);
+	}
+	MultiByteToWideChar(CP_ACP, 0, argv[2], -1,
+			    wsz + desktop_shortcut,
+			    sizeof(wsz) / sizeof(WCHAR) - desktop_shortcut);
 	hres = ppf->lpVtbl->Save(ppf,
 			(const WCHAR*)wsz, TRUE);
 
```

dscho: I now pushed this to the [`improve-desktop-location-detection`](https://github.com/dscho/build-extra/commits/improve-desktop-location-detection) branch in my fork.

Kristof-Mattei: Welp, my C/C++ is not at the level that I'd feel comfortable making a contribution like that.

dscho: You don't have to, you can just cherry-pick the two top commits of the branch I pushed. Then add your commit on top, using `--desktop-shortcut` instead of the Visual Basic scriptlet.

dscho (approved):
I manually fixed a ton of issues because I will not wait for answers or fixes any longer.

PR #273, opened 2020-01-14
Enable FSCache in the Portable Git by default

DanPristupov:
Enable `core.fscache` in the portable git by default as discussed at
[#2467](https://github.com/git-for-windows/git/issues/2467#issuecomment-574109742).


dscho:
Thanks!

PR #274, opened 2020-01-26
Cache recentlySeenVersion in ~/.git-for-windows-updater instead of global config

ciarancourtney:
Implementation as suggested by @dscho in https://github.com/git-for-windows/git/issues/2485#issuecomment-576653219

Depends on `set_recently_seen` being called to make transision, up to user to rm exising global config which is ignored going forward.

Personally I'd rather store it somewhere outside of $HOME, like `/tmp` (problematic for per-user installs I'd guess), `~/.cache/gitforwindows` or even the registry. Any suggestions?

### DIFF
--- a/git-extra/git-update-git-for-windows
+++ b/git-extra/git-update-git-for-windows
@@ -115,13 +115,15 @@ http_get () {
 }
 
 get_recently_seen() {
-    git config --global winUpdater.recentlySeenVersion
+    if [ -f ~/.git-for-windows-updater ]; then
+        git config -f ~/.git-for-windows-updater winUpdater.recentlySeenVersion
+    else
+        git config --global winUpdater.recentlySeenVersion
+    fi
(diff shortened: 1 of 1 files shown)


dscho:
Thank you!

> Depends on `set_recently_seen` being called to make transision, up to user to rm exising global config which is ignored going forward.

Sounds good.

> Personally I'd rather store it somewhere outside of $HOME, like `/tmp` (problematic for per-user installs I'd guess), `~/.cache/gitforwindows` or even the registry. Any suggestions?

`/tmp` is not problematic for per-user installs because `/tmp` is _already_ mapped to the user-specific `%TEMP%` directory.

As to the registry: I do not see any benefit there, I do however see the complication of accessing the registry from a shell script...

So I think what you implemented here is the best we can do, and I thank you very much for contributing this PR!

PR #275, opened 2020-01-29
Un-deprecate Git CMD

dscho:
The reason to deprecate Git CMD went away as of https://github.com/git-for-windows/MINGW-packages/commit/82172388b. So let's revert the deprecation.


dscho:
> I see no problems with this patch, provided that `doskey` is available on all supported Windows versions.

I am pretty certain that it is: https://en.wikipedia.org/wiki/DOSKEY#History

PR #276, opened 2020-01-31
installer: handle Vim 8.2

bbolli:
The installer bash script comments out a part of Vim's defaults.vim
file. This part resets the cursor line to the last one when reopening a
file. This is not desired in the case of commit messages, which are
always the same file name but composed from scratch each time.

In the meantime, MSYS2 Vim was updated to 8.2.

This causes a warning on installing git-extra because the installer
script unconditionally greps a Vim 8.1 file that is no longer present.
Fix this by checking the file's presence.

A similar installer script section is not necessary for Vim 8.2, because
this version includes in this line-resetting automatism an exception for
all file types matching "*commit*".

```
```

Thank you very much.

### DIFF
--- a/git-extra/git-extra.install
+++ b/git-extra/git-extra.install
@@ -186,6 +186,7 @@ test -d "$TMPDIR" || test ! -d "$TMP" || {\
 	sed -i -e '/^# Define default printer/,/^$/d' \
 		-e 's/^\(export .* \)PRINTER /\1/' /etc/profile
 
+	test ! -f /usr/share/vim/vim81/defaults.vim ||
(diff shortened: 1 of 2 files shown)


dscho:
Thank you!

bbolli:
Somewhat related: should `git-extras.installer` be added to `.gitignore`? It's been changed after building git-extras. Or is it committed by the build pipeline?

dscho:
It is committed on purpose: otherwise `makepkg-mingw` will not work.

bbolli:
Ok. Thanks.

PR #280, opened 2020-03-21
Adding missing quotes to setup-git-sdk.bat

evandrocoan:
https://github.com/evandroforks/anki/runs/524380395?check_suite_focus=true#step:2:114

### DIFF
--- a/sdk-installer/setup-git-sdk.bat
+++ b/sdk-installer/setup-git-sdk.bat
@@ -5,13 +5,13 @@
 @REM https://technet.microsoft.com/en-us/library/bb490909.aspx says:
 @REM <percent>~dpI Expands <percent>I to a drive letter and path only.
 @REM <percent>~fI Expands <percent>I to a fully qualified path name.
-@FOR /F "delims=" %%D in ("%~dp0") do @set cwd=%%~fD
+@FOR /F "delims=" %%D in ("%~dp0") do @set "cwd=%%~fD"
 
 @CD "%cwd%"
 @IF ERRORLEVEL 1 GOTO DIE
 
 @REM set PATH
-@set PATH=%cwd%\mini\mingw@@BITNESS@@\bin;%PATH%
+@set "PATH=%cwd%\mini\mingw@@BITNESS@@\bin;%PATH%"
 
 @ECHO Cloning the Git for Windows SDK...
 @git init


sdk-installer/setup-git-sdk.bat
```diff
@@ -5,13 +5,13 @@
 @REM https://technet.microsoft.com/en-us/library/bb490909.aspx says:
 @REM <percent>~dpI Expands <percent>I to a drive letter and path only.
 @REM <percent>~fI Expands <percent>I to a fully qualified path name.
-@FOR /F "delims=" %%D in ("%~dp0") do @set cwd=%%~fD
+@FOR /F "delims=" %%D in ("%~dp0") do @set "cwd=%%~fD"
```
dscho, line 8: Sorry for missing this PR!

So I guess the underlying issue is that the script does not run correctly e.g. when installing into a path whose name has spaces in it?

I am not sure if both this change and the quoting of the `PATH` assignment are needed... When I run

```cmd
set PATH=%PATH%;C:\with spaces
```

it works over here...

sdk-installer/setup-git-sdk.bat
```diff
@@ -31,8 +31,8 @@
 
 @REM Avoid overlapping address ranges
 @IF 32 == @@BITNESS@@ @(
-	ECHO Auto-rebasing .dll files
-	CALL autorebase.bat
+    ECHO Auto-rebasing .dll files
+    CALL autorebase.bat
```
dscho, line 35: Please drop this hunk, it is only changing white-space and that is orthogonal to the purpose of this PR.

sdk-installer/setup-git-sdk.bat
```diff
@@ -47,4 +47,3 @@
 @ECHO Installation of Git for Windows' SDK failed!
 @PAUSE
 @EXIT /B 1
-
```
dscho, line 50: Please drop this white-space only hunk, too.

sdk-installer/setup-git-sdk.bat
evandrocoan:
```suggestion
	CALL autorebase.bat
```

evandrocoan, line 35:
```suggestion
	CALL autorebase.bat
```

sdk-installer/setup-git-sdk.bat
```diff
@@ -31,8 +31,8 @@
 
 @REM Avoid overlapping address ranges
 @IF 32 == @@BITNESS@@ @(
-	ECHO Auto-rebasing .dll files
-	CALL autorebase.bat
+    ECHO Auto-rebasing .dll files
```
evandrocoan, line 34:
```suggestion
	ECHO Auto-rebasing .dll files
```

sdk-installer/setup-git-sdk.bat
```diff
@@ -47,4 +47,3 @@
 @ECHO Installation of Git for Windows' SDK failed!
 @PAUSE
 @EXIT /B 1
```
evandrocoan, line 49:
```suggestion
@EXIT /B 1

```

sdk-installer/setup-git-sdk.bat
evandrocoan: Running `et PATH=%PATH%;C:\with spaces` will set the path to `et PATH=%PATH%;C:\with`

evandrocoan:
What is `<copy-from-sign-off>`?

PhilipOakley:
`<copy-from-sign-off>` is an instruction that you should use, in that place, your formal name and email address, as used in the DCO 'sign-off', so as to make you the formal author as a proper record of your contribution. Without that, other readers will see at as a random drive-by change. Check the DCO for what it means.

dscho:
> `<copy-from-sign-off>` is an instruction that you should use, in that place, your formal name and email address, as used in the DCO 'sign-off', so as to make you the formal author as a proper record of your contribution. Without that, other readers will see at as a random drive-by change. Check the DCO for what it means.

Right. But I overlooked something really important: the Signed-off-by: line is expected to contain a valid email address. So the right thing here would be to call

```sh
git commit --amend --author="Evandro Coan <evandrocoan@hotmail.com>" -s
```

and delete the "Signed-off-by:" line with the URL in it.

While at it, the second commit should be squashed into the first. There is no need to keep the chronological history of the patch in the commit history. In other words:

```sh
git reset --soft HEAD^
git commit --amend --author="Evandro Coan <evandrocoan@hotmail.com>" -s
```

evandrocoan:
I won't put my email in plain text anywhere in a commit history. I added:
```
Signed-off-by: Evandro Coan <https://github.com/evandrocoan>
```

If you want to know my email just access my profile `https://github.com/evandrocoan` and you will find it.

evandrocoan:
@dscho, Please remove my email from your message. I already receive enough spam for several lifetimes.

dscho:
> Please remove my email from your message.

Done.

> I already receive enough spam for several lifetimes.

Yeah, it is really annoying. I totally understand.

PR #282, opened 2020-05-01
Add git-bash process detection script

sladyn98:

```
```

Thank you very much.

### DIFF
--- a/git-extra/git-process.sh
+++ b/git-extra/git-process.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# Check if git is running
+# -x flag only match processes whose name (or command line if -f is specified) exactly match the pattern. 
+if pgrep -x "git-bash.exe" > /dev/null
+then
+    echo "An update has been detected, unfortunately we cannot update while git is running."
+else
+    exec git update-git-for-windows
+fi


git-extra/git-process.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Check if git is running
+# -x flag only match processes whose name (or command line if -f is specified) exactly match the pattern. 
+if pgrep -x "git-bash.exe" > /dev/null
+then
+    echo "An update has been detected, unfortunately we cannot update while git is running."
+else
+    exec git update-git-for-windows
```
sladyn98, line 9: @dscho I am not too sure if this should be called from here ?

git-extra/git-process.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
```
sladyn98, line 2: Check for an update, I am assuming this would be done from the installer.iss but I did not see any bash script being called from there.

dscho:
> A friendly reminder for new contributors:
> # What is DCO sign-off? Why is it required?
> 
> DCO stands for [Developer's Certificate of Origin](https://github.com/git/git/blob/v2.18.0/Documentation/SubmittingPatches#L304-L349).
> 
> We require all commits to be DCO signed-off in case we need to track down and handle legal and/or technical issues.
> 
> To DCO sign-off your commits, you could run the `git-commit` command with [the `-s` option](https://git-scm.com/docs/git-commit#git-commit--s) or just add a line in your commit message in this format:
> 
> ```
>         Signed-off-by: yourFirstName yourLastName <yourEmail@example.org>
> ```
> 
> For more information, check out how PRs are checked for DCO sign-off: https://github.com/probot/dco#how-it-works .
> 
> Thank you very much.

The intention of this template was for you to read it, realize what it says, and that it is not a good description for your Pull Request, so that you can replace it by a good description.

git-extra/git-process.sh
```diff
@@ -0,0 +1,10 @@
...
+# Check if git is running
+# -x flag only match processes whose name (or command line if -f is specified) exactly match the pattern. 
+if pgrep -x "git-bash.exe" > /dev/null
```
dscho, line 5: This would catch _all_ `git-bash.exe` processes, even the ones from portable Gits that are not at all affected by the installer.

git-extra/git-process.sh
dscho: No, `installer.iss` defines what is run during the installer. Why would you check for an update when running the installer?

git-extra/git-process.sh
dscho: I fear that I did not express myself correctly.

The installer has this option where it registers a scheduled task that runs the `git update-git-for-windows --quiet --gui` command once a day. This will check for an update, and if one is available _ask the user_. If I understood correctly, what you wanted to do was to not ask the user, at least not if Git isn't running.

So why not do that?

That would _not_ require a new script, but _build on top of existing work_, i.e. it would be a patch to `git-extra/git-update-git-for-windows` and maybe another option in the installer to make this behavior opt-in.

dscho: To make this explicit, here is the existing code that checks for still-running Git Bash instances: https://github.com/git-for-windows/build-extra/blob/8a462c2105c9103ac53b7d75f20934f75f2a12c6/git-extra/git-update-git-for-windows#L229-L238

The idea would be to introduce a new mode that won't even download, let alone start, the update if even one Git Bash is running. It _might_ be as easy as introducing a new value for the `yn` variable, say, `if-no-git-bash-is-running`, and then inserting something before https://github.com/git-for-windows/build-extra/blob/8a462c2105c9103ac53b7d75f20934f75f2a12c6/git-extra/git-update-git-for-windows#L227-L228, something along these lines:

```sh
if test if-no-git-bash-is-running = "$yn"
then
        # Exit unless no Git Bash is running
        test 0 = $(count_other_bashes) || exit 0
elif test -z "$yn"
[...]
```

sladyn98: Thanks for the explanation, but I am a bit confused, as you mentioned earlier if the user has opted in to the behaviour of automatic updates why are we checking the value of `yn` ? Could we not just check whether the user has opted in to the automatic updates behaviour and then call update for windows without asking for the user.
```
if test opted-in-behaviour= "y"
(Check for running bashes and exit if not found ) 
 test 0 = $(count_other_bashes) || exit 0
then
       update-git-for-windows (without asking)
```

dscho: No, we cannot. The behavior of the feature as it is implemented right now is what most users desire. And we cannot just change it behind their backs.

If you desire a new behavior, you have to implement a new mode.

sladyn98: Got it, :+1: 
Also how would assigning this variable help with what we are trying to achieve . If I understand correctly the yn variable `--yes Automatic yes to download and install prompt` should we not instead check the opt-in behaviour. and then directly begin the update.
```
if test opt-in-behaviour = "y"
then
```

dscho: I just thought that we could extend the `yn` variable to not be a simple "yes" or "no" but make it a tristate, where it now can also indicate that the desire is _not_ to bother the user, but also to just skip the update if the user is currently using Git.

sladyn98: Looking at the code this section gets printed if `$# -gt 0` so we ought to skip this printed section on the console right  ? 
```
printf >&2 '%s\n%s\n\t%s\n\t%s\n' \
			"Usage: git update-git-for-windows [options]" \
			"Options:" \
			"-g, --gui Use GUI instead of terminal to prompt" \
			"-y, --yes Automatic yes to download and install prompt"
		return 1
```

sladyn98: So I am thinking we could pick the yes or no from the opt-in behaviour during initial installation, and then use that answer to carry out our update

dscho: `test $# -gt 0` means that there were some left-over arguments that could not be parsed. In that case, we show an error message. I think we really, really want to retain this behavior.

sladyn98: Yes, but what if the variable `$yn` is `n` then what is the point of having this variable `test if-no-git-bash-is-running = "$yn"` why can't we do something like
if behaviour opted in is y then simply tell the user that the git bash will be updated the next time you start

dscho: There is no way I know of to trigger a Git for Windows install "next time you restart".

sladyn98: Maybe we can take this PR one step at a time :)
Step 1: How do I get the user to choose the opt-in behaviour

dscho: If I were you, I would make this a sub-option of the "Check automatically for updates" checkbox, and if clicked, would add a config setting to the system config that I would then query in the `git-update-git-for-windows` script.

That's just an idea, though, and it is up to you to either take it and drive it forward or think of a different way and drive _that_ forward.

sladyn98: Perfect :+1:

dscho: Any news on that?

PR #283, opened 2020-05-02
Add Git FAQ icon to start menu linking to the wiki/FAQ

PhilipOakley:
---
Note that the icons are listed in alphabetic order on W10 start menu (on my machine), so we have:
 Bash, CMD, FAQs, GUI, Release Notes.


dscho:
Does that really work if you use a URL for a filename?

PhilipOakley:
> Does that really work if you use a URL for a filename?

Yes it does. I was concerned too. Note that the `{app}` is dropped. But mainly see [Innosetup [Icons] section](https://jrsoftware.org/ishelp/index.php?topic=iconssection) where it says:

> In addition to file and folder names, URLs (web site addresses) may also be specified. When a URL is specified, Setup will create an "Internet Shortcut" (.url) file, and ignore the Parameters, WorkingDir, HotKey, and Comment parameters.

The examples then shows the dropping of the `{app}`.

I now see that I could have dropped the `Parameters, WorkingDir, HotKey, and Comment parameters`. :wink:

PhilipOakley:
Updated the commit message, and tweaked the changes to drop the no longer required parameters for the Icon setup.

dscho:
Thanks!

PR #284, opened 2020-05-03
Release notes: show version number for "Changes since" section

PhilipOakley:
Add the current version number as a prefix to the 'Changes since' line.

The "Changes since" sections of the release notes are 'below the fold'
so it is not immediately obvious to readers what is the version the
changes refer to. The version is already available as the header id
but this isn't displayed to the user in any obvious manner.

The html release notes processing is multi-phase. There are maintainer
actions to add new notes to the original markdown version.
The html preparation is part of the `sdk build installer  process. This
provides conversion of the markdown to html, and finally some format
adjustments to the html in render-release-notes.sh.

This code change is just to the final html adjustments to avoid any
changes to the maintainer workflow.

The updated perl script isn't quite as neat as previously,
but is obvious.

---
With the extra prefix the heading line felt rather big and intimidating, maybe larger than some smaller screens, so I have chosen to make the prefix use `<small>` font. Comments welcomed.

### DIFF
--- a/render-release-notes.sh
+++ b/render-release-notes.sh
@@ -115,7 +115,7 @@ render_release_notes () {
 
 				$nr = 0 if (!$nr);
 				$nr++;
-				s/^<h2/$& id="$v" nr="$nr" class="collapsible"/;
+				s/^<h2>/<h2 id="$v" nr="$nr" class="collapsible"><small>$v<\/small>: /;
 				$v = $previous_version;
 				s/.*/<\/div>$&<div>/;
 			}'


dscho:
The current patch has an ugly artifact when release notes are rendered for "in-between" versions:

image

Notice how the unreleased version is mistaken for `v2.26.2`?

dscho:
Maybe something like this, squashed in?

```diff
diff --git a/render-release-notes.sh b/render-release-notes.sh
index d82d75032..173ddde40 100755
--- a/render-release-notes.sh
+++ b/render-release-notes.sh
@@ -115,7 +115,15 @@ render_release_notes () {
 
 				$nr = 0 if (!$nr);
 				$nr++;
-				s/^<h2>/<h2 id="$v" nr="$nr" class="collapsible"><small>$v<\/small>: /;
+				if ($v eq $previous_version) {
+					# in-between version
+					$id = "";
+					$label = "";
+				} else {
+					$id = " id=\"$v\"";
+					$label = "<small>$v</small>: ";
+				}
+				s/^<h2>/<h2$id nr="$nr" class="collapsible">$label/;
 				$v = $previous_version;
 				s/.*/<\/div>$&<div>/;
 			}'
```

dscho:
It's a bit unfortunate that the version number takes such a lot of screen estate. I'd rather have it as a floating label on top of the blue bar, but my HTML fu is not so strong right now.

PhilipOakley:
Hi,  I hadn't noticed the problem with unreleased versions (not sure how versions become unreleased, but that's not important). Yes it does look odd now you point it out.

Thanks for the suggested code snippet. I'll have a look at how to best use it..

I'd agree that the version number (as with the rest of the heading) is taking up a lot of real estate. I'd felt the repetitive "Changes since" lines were a bit bold anyway. I'll look at generally reducing the sizes or some top label, but like you (or worse) my fu is also rather weak (or weaker!) (I could ask the internet, it has opinions, some maybe useful ;-).

Thanks for the feedback.

dscho:
> I hadn't noticed the problem with unreleased versions (not sure how versions become unreleased, but that's not important).

They don't "become" unreleased. They _are_ unreleased, e.g. almost [all the snapshots](https://wingit.blob.core.windows.net/files/index.html).

> I'd agree that the version number (as with the rest of the heading) is taking up a lot of real estate. I'd felt the repetitive "Changes since" lines were a bit bold anyway. I'll look at generally reducing the sizes or some top label, but like you (or worse) my fu is also rather weak (or weaker!) (I could ask the internet, it has opinions, some maybe useful ;-).

How about moving the "Changes since" line below the banner, and putting something like `Git for Windows v<version>` into it (and for the unreleased ones, use something like `Git for Windows snapshot`).

PhilipOakley:
This is superseded by https://github.com/git-for-windows/build-extra/pull/288

Snapshots are detected. Version strip hangs neatly in normal font at the top left.

PR #285, opened 2020-05-03
Release notes: emphasis the upstream release notes link

PhilipOakley:
This two patch series first updates the please.sh used by the maintainer for mentioning of new upstream Git versions in the .md file, and secondly updates the older entries in the current .md file.

The second patch has been tested, while guidance is needed to test the second patch (to get a suitable 'new' git version to actually mention...)

Philip

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -54,7 +54,7 @@ Yet another security fix release: With a crafted URL that contains a newline or 
 
 ### New Features
 
-* Comes with [Git v2.26.2](https://github.com/git/git/blob/v2.26.2/Documentation/RelNotes/2.26.2.txt).
+* Comes with [Git v2.26.2 (Upstream release notes)](https://github.com/git/git/blob/v2.26.2/Documentation/RelNotes/2.26.2.txt).
 * Comes with [tig v2.5.1](https://github.com/jonas/tig/releases/tag/tig-2.5.1).
 * Worktree updates (e.g. `git checkout`, `git reset --hard`) [got a performance boost in sparse checkouts](https://github.com/git-for-windows/git/pull/2589).
 
@@ -68,7 +68,7 @@ This includes a fix for CVE-2020-5260.
 
 ### New Features
 
-* Comes with [Git v2.26.1](https://github.com/git/git/blob/v2.26.1/Documentation/RelNotes/2.26.1.txt).
+* Comes with [Git v2.26.1 (Upstream release notes)](https://github.com/git/git/blob/v2.26.1/Documentation/RelNotes/2.26.1.txt).
 * Comes with [OpenSSL v1.1.1f](https://www.openssl.org/news/openssl-1.1.1-notes.html).
 
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


dscho:
Hmm. Maybe it would make more sense to increase the font size and/or italicize the entry? We could do that via a `sed` expression in `render-release-notes.sh`.

I just don't see regular users having an easy time understanding what "upstream notes" are. It's not exactly English.

PhilipOakley:
Hi @dscho Thanks for the feedback.

Trying to decide how to say that "the release notes that the user is reading isn't the release notes they were looking for" gets to be a bit Catch 22.  Maybe if we called the 'upstream Git release notes' the "Product Release Notes" (in brackets) it would avoid the less familiar 'Upstream' phrasing. What do you think.

The idea of updating the `render-release-notes.sh` now sounds appealing. I hadn't seen that(!) as I'd come via the `please` script's mention() method, and only late last night managed the second patch to create that full `sed` translation expression - I'd never used sed in anger before. I see that I could simply copy the sed expr across to the release sh.  

Also possible is to simply append a " - Product Release Notes" _after_ the link. I was hoping to avoid font size and other fancy tweaks.

dscho:
I'm not crazy about "product" in this context.

My preference would really be: just change the style for the item to stick out better.

PhilipOakley:
I have gone ahead and tried the use of `sed` to pre-adjust the .md file just before it's converted by `markdown` into HTML on my branch [productrelnotes](https://github.com/PhilipOakley/build-extra/tree/productrelnotes)

I felt it was still necessary to explicitly tell the reader/user what's in the git links so that they are more likely to read those release notes. So I have retained the Product Release notes suffix.

 I did have a look around for something that could also present the upstream release notes.txt as if they were markdown (i.e looking nicer) but didn't find anything useful.

I'll also create a PR for that alternate trial in case it passes muster.

dscho:
> My preference would really be: just change the style for the item to stick out better.

I am playing with this patch, but neither italic nor bold look right to me:

```diff
diff --git a/render-release-notes.sh b/render-release-notes.sh
index ff1296d65..9ba592043 100755
--- a/render-release-notes.sh
+++ b/render-release-notes.sh
@@ -99,6 +99,7 @@ render_release_notes () {
 		body="$(markdown "$SCRIPT_PATH"/ReleaseNotes.md)" ||
 		die "Could not generate ReleaseNotes.html"
 		echo "$body" | perl -pe '
+			s/^(<li>)(Comes with <a.*>Git v[1-9].*)(<\/li>)$/\1<big><b>\2<\/b><\/big>\3/g;
 			s/^(<h1)(>Known issues)/\1 id="known-issues" class="collapsible"\2/;
 			s/^(<h2)(>Licenses.*)/\1 id="licenses" class="collapsible"\2<div>/;
 			$v = $1 if (/<h1>Git for Windows (\S+)/);
```

So maybe it should be a subtly different background and/or a frame/shadow.

> I felt it was still necessary to explicitly tell the reader/user what's in the git links so that they are more likely to read those release notes. So I have retained the Product Release notes suffix.

I still disagree, and not just mildly.

dscho:
> > I felt it was still necessary to explicitly tell the reader/user what's in the git links so that they are more likely to read those release notes. So I have retained the Product Release notes suffix.
> 
> I still disagree

And I still do ;-) The upstream release notes are a comprehensive list, way more detailed than a regular Git for Windows user ever wants to read.

PR #286, opened 2020-05-04
ReleaseNotes: highlight the upstream Product release notes links

PhilipOakley:
The "* Comes with Git v2.x.y" in the link does not do justice to the
reality that this is _the_ link to the upstream's release notes for
the key bug fixes, improvements and backward compatibility issues in
this release.

Explicitly state that this link is for the `Product release notes`.

Editing is performed here to avoid affecting the maintainer scripts.

### DIFF
--- a/render-release-notes.sh
+++ b/render-release-notes.sh
@@ -96,7 +96,7 @@ render_release_notes () {
 				</div>
 				<div class="content">
 		EOF
-		body="$(markdown "$SCRIPT_PATH"/ReleaseNotes.md)" ||
+		body="$(sed -r 's|(^\* Comes with \[Git v[0-9].*\))|\1 \(Product release notes\)|' "$SCRIPT_PATH"/ReleaseNotes.md | markdown)" ||
 		die "Could not generate ReleaseNotes.html"
 		echo "$body" | perl -pe '
 			s/^(<h1)(>Known issues)/\1 id="known-issues" class="collapsible"\2/;


dscho:
I really dislike the term "product" here: we are not selling anything, and nobody is paying money. Calling it a "product" would reinforce the already too common misconception that users need only mention bugs and somebody who is not them will magically fix them for them.

But then, I like things they are right now, so...

PhilipOakley:
I'd agree that the term for the results of production ("product") has been misappropriated by economists/capitalists/snake-oil-sales-people, but I couldn't find an alternative (the thesaurus/synonyms did come up with anything either). (Nothing did any of the various 'open source' resources either - a bit of a Newspeak problem).

I did come up with just referring to 'it' as "Git-core" (also came up in a discussion thread this evening).

This would at least distinguish between the "Git" (as in Git-for-Windows) and "Git" (as in upstream, git-core) for those arriving without any real knowledge of the distinctions. Leaving the 'comes with' as simply 'Git' feels a bit of an oxymoron, given that they are already reading the Git release notes....

Would "Git-core release notes" be a reasonable compromise?

dscho:
I don't really think that any of these suggestions improve on the current wording.

PhilipOakley:
> I don't really think that any of these suggestions improve on the
> current wording.
>
>

Ok. Let's wait while the effects of making the release notes more
visible works through (#281 build-extra).

I still feel that as user facing documentation it shouldn't be so terse.

Philip

dscho:
@PhilipOakley if you don't mind, I would like to simply close this. The reason is two-fold:

- We seem not to reach any consensus how this should be worded, and
- the upstream release notes are rather tedious to read because they focus on a comprehensive list of changes instead of highlighting the most visible changes. If there was a way to link automatically to the incredibly well-written blog posts published by @ttaylorr of GitHub about every new Git version, I would indeed consider doing it. As it stands, I would like to keep the current (non-emphasized) rendering of the item linking to Git's own release notes.

ttaylorr:
> * the upstream release notes are rather tedious to read because they focus on a comprehensive list of changes instead of highlighting the most visible changes. If there was a way to link automatically to the incredibly well-written blog posts published by @ttaylorr of GitHub about every new Git version

I'm flattered that you would include GitHub's blog post, but I actually think that upstream's full release notes are more appropriate here.

PhilipOakley:
> The "* Comes with Git v2.x.y" in the link does not do justice to the
> reality that this is _the_ link to the upstream's release notes for
> the key bug fixes, improvements and backward compatibility issues in
> this release.

I feel that my initial concern is still true. I can see that you may not want to call the core Git software a 'Product' with a potential commercial association.

I can also see that there is a difference in tone between Taylor's nice overview, and the techy/dev details of Junio's release notes, but either way, I'm suggesting we point to the Git notes with a little bit more emphasis. 

Maybe it's just a case of **emboldening** the link? (Internet suggests to add style="font-weight:bold" in between <a and href. ... which is a small improvement). I haven't looked yet at how that would work into the somewhat convoluted scripts that prepare the notes;-)

PhilipOakley:
> > * If there was a way to link automatically to the incredibly well-written blog posts published by @ttaylorr of GitHub about every new Git version
> 
> I'm flattered that you would include GitHub's blog post, but I actually think that upstream's full release notes are more appropriate here.

@ttaylorr Hi, Is your blog post a personal contribution (independent of GitHub), or part of the GitHub support? I'm just thinking of any workload issues, should we added a generic link in the Git For Windows Release notes to the GitHub blog, as it's a place that will provide an _overview_ of the changes. However I didn't see a specific link that would best reflect just the main Git release blogs (e.g. somewhere in, maybe, https://github.blog/category/engineering/ ?) - Is there one?

It looks like your blog for releases comes out more or less at the same time as the release, so any timing delays shouldn't be too much of an issue (early adopters generally having been primed by announcements on the Git/GfW lists) as the blog should be there soon enough for most.

dscho:
> Maybe it's just a case of **emboldening** the link?

That could work.

Simply putting two asterisks around the `Git v<version>` would probably break too much (e.g. [this logic](https://github.com/git-for-windows/build-extra/blob/8a8198a9391975d508dbb14378517979770a6ef7/please.sh#L3582-L3593)), so it would maybe make most sense to add something like `else { s/(<a [^>]*(>Git v)/\1 class="upstream-git-version"\2/; }` [here](https://github.com/git-for-windows/build-extra/blob/8a8198a9391975d508dbb14378517979770a6ef7/render-release-notes.sh#L127) and then adding something like this to [`ReleaseNotes.css`](https://github.com/git-for-windows/build-extra/blob/main/ReleaseNotes.css):

```css

.upstream-git-version {
	font-weight: bold;
}
```

PhilipOakley:
> > Maybe it's just a case of **emboldening** the link?
> 
> That could work.
> 
> Simply putting two asterisks around the `Git v<version>` would probably break too much (e.g. [this logic](https://github.com/git-for-windows/build-extra/blob/8a8198a9391975d508dbb14378517979770a6ef7/please.sh#L3582-L3593)), so it would maybe make most sense to add something like `else { s/(<a [^>]*(>Git v)/\1 class="upstream-git-version"\2/; }` [here](https://github.com/git-for-windows/build-extra/blob/8a8198a9391975d508dbb14378517979770a6ef7/render-release-notes.sh#L127) and then adding something like this to [`ReleaseNotes.css`](https://github.com/git-for-windows/build-extra/blob/main/ReleaseNotes.css):
> 
> ```css
> .upstream-git-version {
> 	font-weight: bold;
> }
> ```

@dscho Many thanks for those hints. 

I'll see if I can get something sorted over the next few days. I'm doing too much juggling without getting anything completed at the moment, (weary sigh)  😩

dscho:
>     * the upstream release notes are rather tedious to read because they focus on a comprehensive list of changes instead of highlighting the most visible changes.

I think this still applies, and will therefore close this PR.

PR #288, opened 2020-05-20
Show 'current' version for Changes since collapsible headings

PhilipOakley:
These commits are a replacement for PR #284. It provides a small reminder of the release version that has the changes that are being discussed within the collapsed heading.  Snapshots are properly indicated. Other h2 collapsible headings have a similar tag at the top left.

The Known Issues heading is also set to h2, and the Other Problems paragraph given it's own small h4 heading.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -7,7 +7,7 @@ These release notes describe issues specific to the Git for Windows release. The
 
 See [http://git-scm.com/](http://git-scm.com/) for further details about Git including ports to other operating systems. Git for Windows is hosted at [https://gitforwindows.org/](https://gitforwindows.org/).
 
-# Known issues
+## Known issues
 * On Windows 10 before 1703, or when Developer Mode is turned off, special permissions are required when cloning repositories with symbolic links, therefore support for symbolic links is disabled by default. Use `git clone -c core.symlinks=true <URL>` to enable it, see details [here](https://github.com/git-for-windows/git/wiki/Symbolic-Links).
 * If configured to use Plink, you will have to connect with [putty](http://www.chiark.greenend.org.uk/~sgtatham/putty/) first and accept the host key.
 * Some console programs, most notably non-MSYS2 Python, PHP, Node and OpenSSL, interact correctly with MinTTY only when called through `winpty` (e.g. the Python console needs to be started as `winpty python` instead of just `python`).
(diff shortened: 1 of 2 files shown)


dscho:
Unfortunately, this is too late for v2.27.0-rc1, as [the build](https://dev.azure.com/git-for-windows/git/_build/results?buildId=56523&view=logs&j=31a7ecae-43f3-5669-2a92-3fc5c046b795) is already running.

I'll have a look at the patches now.

dscho:
I see a couple of problems with the current approach:

- when "Known Issues" and the Git for Windows versions are the same type, it is much harder to see quickly what is what. But the known issues are a fundamentally different thing than the release notes for a given version!

- Those labels inside the headers look ugly: 
image
  - for single-line headers, the labels overlap with the top line, but not for double-line headers
  - the `snapshot` label collides with the header, depending on the window size.

Quite honestly, I'd rather have the labels _only_ for the individual versions' release notes, and I'd like the labels to be visually pleasing.

PhilipOakley:
No problem. One old thing I just noticed (isn't it always jsut after submission..), the collapsible regions stop at 
```
Changes since Git for Windows v2.6.4
(December 14th 2015)
```
because the subsequent changes sections are labelled as 
`## Changes since Git-2.6.0 (September 29th 2015)`   
 (i.e. `Git-xxx`, which isn't caught by the if/substitution https://github.com/git-for-windows/build-extra/blob/master/render-release-notes.sh#L105-L121

dscho:
Good point!

PhilipOakley:
The version number and title weren't overlapping for me, so I hadn't noticed that part.

dscho:
> The version number and title weren't overlapping for me, so I hadn't noticed that part.

Yep. Web design ain't easy.

dscho:
I took the bug fix early, so that this here PR can really concentrate on showing the version numbers in a pretty fashion.

PhilipOakley:
Thanks. 

One thought was to adjust the top border of the h2 block, and then adjust the pointer position, so that the version stays clear of the main h2 title line.

Also should look at the next-previous detection logic for the older Git-x.y.z versions, and the final `<div>` which gobbles everything to the end. (as well at the Licence/ KnownIssues blocks, which are similar collapsibles)

PhilipOakley:
Ahh. see the WinGit is fixed. and some other fixes I haven't fully groked yet.

dscho:
I dunno. I could imagine that a small div with rounded corners and a different background color could sit in (or above) the top-left corner of those section headers. That would look neat.

PhilipOakley:
> a small div with rounded corners and a different background color could sit in (or above) the top-left corner of those section header

Sounds like a HTML fu research project :notebook:

dscho:
> > a small div with rounded corners and a different background color could sit in (or above) the top-left corner of those section header
> 
> Sounds like a HTML fu research project 📓

More like a CSS one ;-) (https://css-tricks.com/snippets/css/rounded-corners/, for example, but the trickiest part might be the positioning, really, especially in a [responsive](https://en.wikipedia.org/wiki/Responsive_web_design) way.)

PhilipOakley:
RFC third version now at https://github.com/git-for-windows/build-extra/pull/290 uses 'button' as the floating supra-title.
(the GitHub branch name juggling (or not) for versions isn't helping here 🤷 :)

dscho:
TBH I'm surprised that you didn't simply force-push to this PR and instead appended pseudo version numbers to the branch names. Not very Git-like.

dscho:
I believe this has been addressed in the meantime, in a different way.

PR #290, opened 2020-05-26
[RFC] ReleaseNotes: Show 'current' version for "Changes Since" titles

PhilipOakley:
Third version, based on most recent master.

Opinions, especially the colours! 
(Helped by colleague @strath.ac.uk who knows a bit more about some of the HTML fu)
Also, should a title at the bottom of the page auto scroll a little when opened, to show something happening..

P.

The Git for Windows Release Notes show the "Changes Since" title lines
but don't tell the user the current version they refer to.

Convert the Arrow indication to a floating button at the top left
of the header showing the version, and an expansion arrow.

Currently the button is a default grey shade, with a contrasting
green font lettering (valuely matching the logo's green).

---
RFC. The previous white font is not readable/accessible against
the grey button.

### DIFF
--- a/render-release-notes.sh
+++ b/render-release-notes.sh
@@ -105,6 +105,13 @@ render_release_notes () {
 			if ((/^<h2>Changes since (Git for Windows ([^ <]*)|(Win)?Git-([^ <]*))/)) {
 				$previous_version = $2 ? $2 : "v$4";
 
+				if ($v eq $previous_version) {
+					# in-between version
+					$v = "snapshot"
+				}
+
+				s/^<h2>Changes since(.*)/<h2>Changes in $v <small>since \1<\/small>/;
+
 				if (!$latest) {
 					s/>[^<]*/><a name="latest"$&<\/a>/;
 					$latest = 1;
@@ -116,10 +123,7 @@ render_release_notes () {
 				$nr = 0 if (!$nr);
 				$nr++;
 				$id = $v;
-				if ($v eq $previous_version) {
-					# in-between version
-					$id = "snapshot"
-				}
+
 				s/^<h2>/<h2 id="$id" nr="$nr" class="collapsible"> /;
 				$v = $previous_version;
 				s/.*/<\/div>$&<div>/;
--- a/render-release-notes.sh
+++ b/render-release-notes.sh
@@ -110,7 +110,7 @@ render_release_notes () {
 					$v = "snapshot"
 				}
 
-				s/^<h2>Changes since(.*)/<h2>Changes in $v <small>since \1<\/small>/;
+				s/^<h2>Changes since (?:Git(?: for Windows|-))(.*)/<h2>Changes in $v <small>since \1<\/small>/;
 
 				if (!$latest) {
 					s/>[^<]*/><a name="latest"$&<\/a>/;


dscho:
For lurkers: this is how it looks over here:

image

PhilipOakley:
@dscho Thank you for the picture. I hadn't even thought to do that. 

Obviously, the `<h1> Known Issues` heading 'button' probably also needs a height tweak somehow.

PhilipOakley:
I've pushed an updated version of the Release Notes processing, with an updated position/shape for the collapsible region's button. (image in a moment)

I found that the `$ sdk build git-and-installer` had stopped working for me (last used a few days ago), with 
```
Generating file list to be included in the installer ...
../make-file-list.sh: line 75: pactree: command not found
```

Googling the error, showed that it (pactree) in the not too distant past, had been moved to `pacman-contrib`. I had been having repeated updates while using the `sdk build`, typically with
```
warning: git-extra-1.1.464.9434eb4-1 is up to date -- reinstalling
 there is nothing to do
:: Starting full system upgrade...
```
Which sounds like something is out of sync.

This may, or may not have anything to do with this `sdk build` failure.

Eventually, after following the [pacman help page](https://wiki.archlinux.org/index.php/pacman) tip, I installed the `pacman-contrib` (which then indicated that only `vi` had actually been installed). 

The next build (apart from reinstalling `git-extra` yet again) immediately succeeded and generated this screen grab.

RelNotes

Opinions? (most likely far too late for v2.27.0)

dscho:
I have looked at this screenshot a couple of times over the past few days, and I can't say that I find those buttons pleasing. They're actually pretty obtrusive.

We could of course consider a completely different approach where we _change_ the section titles from `Changes since <version>` to `Changes in <version> (<small>since <version></small>)`. I could get behind that.

PhilipOakley:
Hi @dscho The button colour (and hence font colour choice) was just the default colour. It could be changed to the same background as the heading, _if_ that would match your initial suggestion a while back about some form of 'floating' version number tab at the top of the title (I was just guessing as to what you hoped that would actually look like).

I've not been able to put much time into Git in the last week or so as I was doing some remote sensing / photon imaging study work. 

The `Changes in <version> (<small>since <version></small>)` is also doable. 

Should I even try the button colour change (so that it looks like a tab)?

PhilipOakley:
Just added a test patch (3cf56ac) to see about my first suggestion that continued the button/tab approach (for the record). with picture
image

I'll leave that one for the moment (it would still need slight tweaks on the edging if the idea was acceptable). I'll get on to the alternate wording approach (and based on V2.27).

PhilipOakley:
The last (force) push uses the suggested `Changes in <version> (<small>since <version></small>)` approach.

It looks like... gahh

The `snapshot` code needs moving earlier (or the date line wrap moved later), so that the most a snapshot recent version shows as such!

PhilipOakley:
Updated and force pushed code to ensure the 'snapshot' designation is included in the `Changes` line where appropriate.

image

The earlier `-previewyyyymmdd` version are a bit garish, but then they are old (well below the display fold)

image

dscho:
Hmm. Maybe we do not even have to repeat the "Git for Windows" all the time? Like, "Changes in v2.27.0 since v2.26.2" or even shorter "Changes in v2.27.0 vs v2.26.2"?

dscho:
(I do like the latest look much more than the buttons, and the diff is pleasant to read.)

PhilipOakley:
Ok, I can remove the "Git for Windows" part (i.e. just use the $ variables). 

The one minor hiccup is is that there are one or two early "preview" versions which had the same semver version but different release dates which would look 'odd' without the date part. Not sure about the best way to handle those? 
e.g. (a) remove 'preview' and leave the '-yyyymmdd' part, (b) remove both (i.e. the hiccup), (c) detect the apparent duplication and add a suffix (but if dated versions we're stuck again). Maybe start with (a)

dscho:
> The one minor hiccup is is that there are one or two early "preview" versions which had the same semver version but different release dates which would look 'odd' without the date part. Not sure about the best way to handle those?

It's so old that it does not matter, I don't think.

> e.g. (a) remove 'preview' and leave the '-yyyymmdd' part, (b) remove both (i.e. the hiccup), (c) detect the apparent duplication and add a suffix (but if .3. dated versions we're stuck again). Maybe start with (a)

Why remove the "preview" part? This was Git for Windows v1.x, I highly doubt that it is worth the electrons we're expending to talk about this.

dscho:
> can remove the "Git for Windows" part (i.e. just use the $ variables).

I just did that.

PhilipOakley:
+1, Thanks.

PR #292, opened 2020-06-05
Pull option fix

rimrul:
We currently don't save the `fast-forward only` pull behaviour correctly.

Let's fix that.

This fixes git-for-windows/git#2668


dscho:
Thank you!

PR #293, opened 2020-06-11
Add bash-completion packages in Git for Windows 

glhez:
This should fix https://github.com/git-for-windows/git/issues/2562 

The introduction of `bash-completion` should also fix `vim` completion which depends on `_filedir` function defined only with `bash-completion`.

This will also affect `_get_comp_words_by_ref`: the implementation found in Git (https://github.com/git/git/blob/master/contrib/completion/git-completion.bash#L261) differ from bash-completion (https://github.com/scop/bash-completion/blob/133790c464c68fa728b2957a66ee1fa012181a88/bash_completion#L369):

The following completion will not work as intended:

```bash
#!/bin/bash

_my_completion() {
  local lf=$'\n'
  local s="COMP_WORDBREAKS=[${COMP_WORDBREAKS}]${lf}"
  for exclude in '@=' '=' ; do
    local cur='' prev='' cword=''
    local -a words=()
    _get_comp_words_by_ref -n "${exclude}" cur prev words cword
    s+="exclude: ${exclude}"
    s+=", cur=[$cur]"
    s+=", prev=[$prev]"
    s+=", cword=[$cword]"
    s+=", words[${#words[@]}]:"
    for (( i = 0; i < ${#words[@]}; ++i)); do
      s+=" [${words[$i]}]"
    done
    s+="${lf}"
  done
  echo "${lf}$s"
}

complete -o bashdefault -o nospace -o dirnames -o filenames -F _my_completion x
```

On Linux (5.4.28-gentoo-x86_64 with bash-completion):

```text
  $ x --abricot @ad ppp @A<TAB>
  COMP_WORDBREAKS=[\n"'><=;|&(:]
  exclude: @=, cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
  exclude: = , cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
  $ export COMP_WORDBREAKS="${COMP_WORDBREAKS}@"
  $ x --abricot @ad ppp @A<TAB>
  COMP_WORDBREAKS=[\n"'><=;|&(:@]
  exclude: @=, cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
  exclude: = , cur=[A] , prev=[@]  , cword=[6], words[7]: [x] [--abricot] [@] [ad] [ppp] [@] [A]
```

As we can see, the `_get_comp_words_by_ref` function work as intended: the `@A` is kept as is when exclude is `@=` or when `COMP_WORDBREAKS` does not contains `@`.

On Windows, with git version 2.27.0.windows.1, it behaves differently because we use Git definition of `_get_comp_words_by_ref`: we observe that `COMP_WORDBREAKS` contains `@` and that's when the `_get_comp_words_by_ref` fail to work correctly:

```
  $ x --abricot @ad ppp @A<TAB>
  COMP_WORDBREAKS=[\n"'@><=;|&(:]
  exclude: @=, cur=[ppp@A], prev=[--abricot@ad], cword=[2], words[3]: [x] [--abricot@ad] [ppp@A]
  exclude: = , cur=[A]    , prev=[@], cword=[6], words[7]: [x] [--abricot] [@] [ad] [ppp] [@] [A]
  $ export COMP_WORDBREAKS="${COMP_WORDBREAKS//@/}"
  $ x --abricot @ad ppp @A<TAB>
  COMP_WORDBREAKS=[\n"'><=;|&(:]
  exclude: @=, cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
  exclude: = , cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
```

### DIFF
--- a/make-file-list.sh
+++ b/make-file-list.sh
@@ -49,6 +49,11 @@ then
 	UTIL_PACKAGES="$UTIL_PACKAGES tmux libevent"
 fi
 
+if test -z "$MINIMAL_GIT"
+then
+	UTIL_PACKAGES="$UTIL_PACKAGES bash-completion"
+fi
+
 this_script_dir="$(cd "$(dirname "$0")" && pwd -W)" ||
 die "Could not determine this script's dir"
 


make-file-list.sh
```diff
@@ -13,7 +13,7 @@ SH_FOR_REBASE=dash
 PACKAGE_EXCLUDES="db info heimdal git util-linux curl git-for-windows-keyring
 	mingw-w64-p11-kit"
 EXTRA_FILE_EXCLUDES=
-UTIL_PACKAGES="sed awk grep findutils coreutils"
+UTIL_PACKAGES="sed awk grep findutils coreutils bash-completion"
```
dscho, line 16: Since `bash-completion` is firmly in the "interactive usage" space, it is incorrect to include it even in MinGit.

dscho:
> Introduce bash-completion in Git for Windows

We have Bash completion in Git for Windows. Therefore, this commit message title is quite confusing. Could you fix that?

> This should fix [git-for-windows/git#2562](https://github.com/git-for-windows/git/issues/2562)

I don't actually think so. That ticket specifically complains thusly:

> In standard Msys2 or standard Linux, bash_completion directory structure is like this:
>
> /usr/share/bash-completion/bash_completion is a file that contains completion functions.
> /usr/share/bash-completion/completion is a directory that contains many scripts for different commands, including git
>
> But in git-bash,
>
> /usr/share/bash-completion/bash_completion does not exists
> /usr/share/bash-completion/completion is a directory that contains only several scripts for svn, vim, ..., but without git
/mingw64/share/git/completion/git-completion.bash: correspond to /usr/share/bash-completion/completion/git
/mingw64/share/git/completion/git-prompt.sh: only in git-for-windows

So the complaint was that Git's completion is not at the expected location. And including `bash-completion` won't fix this, I don't think, as the following command came up empty:

```sh
$ pacman -Ql bash-completion | grep git
```

In any case, whenever we add things to the Git for Windows installer, we must weigh the cost vs the benefit. Each package we add increases the size of the installer, and it is already relatively large with ~46MB.

Now, running this:

```sh
$ (cd / && tar cf - $(pacman -Ql bash-completion | sed -n 's/^bash-completion \/\(.*[^\/]\)$/\1/p') | xz -9 | wc)
    614    3546  156392
```

suggests that it would add only ~150 kilobytes to the size. That might sound like almost nothing, especially compared with the current installer size. But it is a cost, and since you seem to be the first person reporting problems that _might_ be solved by including this, the benefit for the median Git for Windows user is almost certainly zero.

Even more: looking at the contents of the `bash-completion` package, I see completions for many commands that are definitely not shipped in Git for Windows, and most of them probably will never make sense, anyway, such as `yum-arch`.

> The introduction of `bash-completion` should also fix `vim` completion which depends on `_filedir` function defined only with `bash-completion`.

Can you add the specific symptom of that problem?

> The following completion will not work as intended:
> 
> ```shell
> #!/bin/bash
> 
> _my_completion() {
>   local lf=$'\n'
>   local s="COMP_WORDBREAKS=[${COMP_WORDBREAKS}]${lf}"
>   for exclude in '@=' '=' ; do
>     local cur='' prev='' cword=''
>     local -a words=()
>     _get_comp_words_by_ref -n "${exclude}" cur prev words cword
>     s+="exclude: ${exclude}"
>     s+=", cur=[$cur]"
>     s+=", prev=[$prev]"
>     s+=", cword=[$cword]"
>     s+=", words[${#words[@]}]:"
>     for (( i = 0; i < ${#words[@]}; ++i)); do
>       s+=" [${words[$i]}]"
>     done
>     s+="${lf}"
>   done
>   echo "${lf}$s"
> }
> 
> complete -o bashdefault -o nospace -o dirnames -o filenames -F _my_completion x
> ```
> 
> On Linux (5.4.28-gentoo-x86_64 with bash-completion):
> 
> ```
>   $ x --abricot @ad ppp @A<TAB>
>   COMP_WORDBREAKS=[\n"'><=;|&(:]
>   exclude: @=, cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
>   exclude: = , cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
>   $ export COMP_WORDBREAKS="${COMP_WORDBREAKS}@"
>   $ x --abricot @ad ppp @A<TAB>
>   COMP_WORDBREAKS=[\n"'><=;|&(:@]
>   exclude: @=, cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
>   exclude: = , cur=[A] , prev=[@]  , cword=[6], words[7]: [x] [--abricot] [@] [ad] [ppp] [@] [A]
> ```
> 
> As we can see, the `_get_comp_words_by_ref` function work as intended: the `@A` is kept as is when exclude is `@=` or when `COMP_WORDBREAKS` does not contains `@`.
> 
> On Windows, with git version 2.27.0.windows.1, it behaves differently because we use Git definition of `_get_comp_words_by_ref`: we observe that `COMP_WORDBREAKS` contains `@` and that's when the `_get_comp_words_by_ref` fail to work correctly:
> 
> ```
>   $ x --abricot @ad ppp @A<TAB>
>   COMP_WORDBREAKS=[\n"'@><=;|&(:]
>   exclude: @=, cur=[ppp@A], prev=[--abricot@ad], cword=[2], words[3]: [x] [--abricot@ad] [ppp@A]
>   exclude: = , cur=[A]    , prev=[@], cword=[6], words[7]: [x] [--abricot] [@] [ad] [ppp] [@] [A]
>   $ export COMP_WORDBREAKS="${COMP_WORDBREAKS//@/}"
>   $ x --abricot @ad ppp @A<TAB>
>   COMP_WORDBREAKS=[\n"'><=;|&(:]
>   exclude: @=, cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
>   exclude: = , cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
> ```

I would like to believe that I am an POSIX shell script expert, yet even so, this part reads pretty much like complete gibberish to me. I have no idea what the `_my_completion` sample is supposed to do, not on a higher level, nor on a technical level.

The only piece of information I extract from that wall of text is that Git's Bash completion defines a `_get_comp_words_by_ref` that is apparently incompatible with other implementations of that function, and you need maybe a fixed version thereof?

Please help me (and others) understand the problem so that we can work on an acceptable solution.

glhez:
> > Introduce bash-completion in Git for Windows
>
> We have Bash completion in Git for Windows. Therefore, this commit message title is quite confusing. Could you fix that?

Fixed.

>
> > This should fix [git-for-windows/git#2562](https://github.com/git-for-windows/git/issues/2562)
>
> I don't actually think so. That ticket specifically complains thusly:
>
> > In standard Msys2 or standard Linux, bash_completion directory structure is like this:
> > /usr/share/bash-completion/bash_completion is a file that contains completion functions.
> > /usr/share/bash-completion/completion is a directory that contains many scripts for different commands, including git
> > But in git-bash,
> > /usr/share/bash-completion/bash_completion does not exists
> > /usr/share/bash-completion/completion is a directory that contains only several scripts for svn, vim, ..., but without git
> > /mingw64/share/git/completion/git-completion.bash: correspond to /usr/share/bash-completion/completion/git
> > /mingw64/share/git/completion/git-prompt.sh: only in git-for-windows
>
> So the complaint was that Git's completion is not at the expected location. And including `bash-completion` won't fix this, I don't think, as the following command came up empty:
>

I think you misunderstood the complaint: @linquize speaks about standard MSYS2 and Linux, and gives the current installation, not the package. I think that in MSYS2 and Linux, `bash-completion` is pretty much included by default. `bash-completion` is different than Git completion: perhaps Git completion was written to support wider range of Linux distribution, supporting Bash 3 (at least, it is advertised like that in `git-completion.bash`), or simply because `bash-completion` may be completely absent.

`bash-completion` provides supports to help writing completion whereas `git-completion.bash` simply support git own completion.

You may use some of those helper in Git for Windows, but they are intended for Git completion, and may contains bugs not relevant to Git completion (such as mine).

As an example, let's assume you have a `--help` option to some `sdk` command: `complete -o bashdefault -o nospace -o dirnames -o filenames -f _longopt sdk`: now you will have completion based on result of `sdk --help`. That is what's done for `ls`, and other command.

Do note that even without my pull request integrated, if @linquize or me, want to use `bash_completion`, the easiest way is simply to clone `bash-completion`  git repository and source manually `bash_completion`.

This is done by `/etc/profile.d/bash_completion.sh` in the case of `bash-completion` packages; I assume that MSYS2 packages maintainer may apply fixes to make `bash-completion` work correctly with MSYS2.

> ```shell
> $ pacman -Ql bash-completion | grep git
> ```

I don't know pacman, but if that list the files contained in the package, I think that's pretty much normal: there is [no git completion file][GH-BASHCOMP-LIST] in `bash-completion`.

>
> In any case, whenever we add things to the Git for Windows installer, we must weigh the cost vs the benefit. Each package we add increases the size of the installer, and it is already relatively large with ~46MB.
>
> Now, running this:
>
> ```shell
> $ (cd / && tar cf - $(pacman -Ql bash-completion | sed -n 's/^bash-completion \/\(.*[^\/]\)$/\1/p') | xz -9 | wc)
>     614    3546  156392
> ```
>
> suggests that it would add only ~150 kilobytes to the size. That might sound like almost nothing, especially compared with the current installer size. But it is a cost, and since you seem to be the first person reporting problems that _might_ be solved by including this, the benefit for the median Git for Windows user is almost certainly zero.
>
> Even more: looking at the contents of the `bash-completion` package, I see completions for many commands that are definitely not shipped in Git for Windows, and most of them probably will never make sense, anyway, such as `yum-arch`.
>

The `bash_completion` itself can be useful to users: you have completion on commands such as `ls`, `find`, `grep`, ... And if user installs Git for Windows, I think they do want a functional shell: otherwise they would use Tortoise Git, or the other GUI based client (why may be more user friendly).

Those users are perhaps even Linux user that would want such shell to work seamlessly on Windows: the completion is part of this.

Anyway I suppose I could try to prune the completions that won't work because of Windows. I think that `bash_completion` script is pretty much stand-alone.

However without or without introducing `bash-completion`, this work should probably done _today_: look at `/usr/share/bash-completion/completions`, the command below does not seems to work or completion fails:

- `gapplication`: does the command works? `gapplication list-apps` output: *No such file or directory + Failed to load module: /usr/lib/gio/modules/msys-giofam.dll*
- `gresource`: command not found.
- `sdk`: fails, requires the `sdk` command, which is not present.
- `vim`: `vim gre<TAB>` fails, requires `_filedir` found in `bash-completion`.

I don't know what the final decision will be, but if you want me to prune those files, please confirm me what's the best place (I've seen that `make-file-list.sh` is grepping a lot of path, is it where files are pruned?).

> > The introduction of `bash-completion` should also fix `vim` completion which depends on `_filedir` function defined only with `bash-completion`.
>
> Can you add the specific symptom of that problem?
>

Easy:

```
[21:41:06] /e/git [11.0.7]
$ . /usr/share/bash-completion/completions/vim
```

Now, try to complete vim:

```
[22:40:35] /e/git [11.0.7]
$ vim po<TAB>bash: _filedir: command not found
```

`_filedir` is defined in `bash-completion`.

In fact, I've seen this problem 3 months ago when playing with dynamic completion loading and seen that `vim` completion was present. The completion found in the `/usr/share/bash-completion/completions` folder are never loaded.

Apart from deleting the completion file, or adding the `bash-completion` package, the other solution is to define a `_filedir`. But that would only work for those interested by vim completion (sadly on Windows, I don't think many people would want to use vim... :( ).

> > The following completion will not work as intended:
> > ```shell
> > #!/bin/bash
> >
> > _my_completion() {
> >   local lf=$'\n'
> >   local s="COMP_WORDBREAKS=[${COMP_WORDBREAKS}]${lf}"
> >   for exclude in '@=' '=' ; do
> >     local cur='' prev='' cword=''
> >     local -a words=()
> >     _get_comp_words_by_ref -n "${exclude}" cur prev words cword
> >     s+="exclude: ${exclude}"
> >     s+=", cur=[$cur]"
> >     s+=", prev=[$prev]"
> >     s+=", cword=[$cword]"
> >     s+=", words[${#words[@]}]:"
> >     for (( i = 0; i < ${#words[@]}; ++i)); do
> >       s+=" [${words[$i]}]"
> >     done
> >     s+="${lf}"
> >   done
> >   echo "${lf}$s"
> > }
> >
> > complete -o bashdefault -o nospace -o dirnames -o filenames -F _my_completion x
> > ```
> >
> >
> > On Linux (5.4.28-gentoo-x86_64 with bash-completion):
> > ```
> >   $ x --abricot @ad ppp @A<TAB>
> >   COMP_WORDBREAKS=[\n"'><=;|&(:]
> >   exclude: @=, cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
> >   exclude: = , cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
> >   $ export COMP_WORDBREAKS="${COMP_WORDBREAKS}@"
> >   $ x --abricot @ad ppp @A<TAB>
> >   COMP_WORDBREAKS=[\n"'><=;|&(:@]
> >   exclude: @=, cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
> >   exclude: = , cur=[A] , prev=[@]  , cword=[6], words[7]: [x] [--abricot] [@] [ad] [ppp] [@] [A]
> > ```
> >
> >
> > As we can see, the `_get_comp_words_by_ref` function work as intended: the `@A` is kept as is when exclude is `@=` or when `COMP_WORDBREAKS` does not contains `@`.
> > On Windows, with git version 2.27.0.windows.1, it behaves differently because we use Git definition of `_get_comp_words_by_ref`: we observe that `COMP_WORDBREAKS` contains `@` and that's when the `_get_comp_words_by_ref` fail to work correctly:
> > ```
> >   $ x --abricot @ad ppp @A<TAB>
> >   COMP_WORDBREAKS=[\n"'@><=;|&(:]
> >   exclude: @=, cur=[ppp@A], prev=[--abricot@ad], cword=[2], words[3]: [x] [--abricot@ad] [ppp@A]
> >   exclude: = , cur=[A]    , prev=[@], cword=[6], words[7]: [x] [--abricot] [@] [ad] [ppp] [@] [A]
> >   $ export COMP_WORDBREAKS="${COMP_WORDBREAKS//@/}"
> >   $ x --abricot @ad ppp @A<TAB>
> >   COMP_WORDBREAKS=[\n"'><=;|&(:]
> >   exclude: @=, cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
> >   exclude: = , cur=[@A], prev=[ppp], cword=[4], words[5]: [x] [--abricot] [@ad] [ppp] [@A]
> > ```
>
> I would like to believe that I am an POSIX shell script expert, yet even so, this part reads pretty much like complete gibberish to me. I have no idea what the `_my_completion` sample is supposed to do, not on a higher level, nor on a technical level.

That's probably because this is pure bash, and not POSIX :)

Programmable completion is probably not POSIX at all (can't say) and debugging completion is a challenge in itself (at least, for me). I do use a lot of `echo` with string concatenation (the `+=`).

>
> The only piece of information I extract from that wall of text is that Git's Bash completion defines a `_get_comp_words_by_ref` that is apparently incompatible with other implementations of that function, and you need maybe a fixed version thereof?
>
> Please help me (and others) understand the problem so that we can work on an acceptable solution.

For my use case (with a `@`), I can say that `_get_comp_words_by_ref` is bugged when it deals with `@`.

Best for me is to quote the Bash documentation:

> [From the manual][BASH-COMP]: When the command or function is invoked, the COMP_LINE, COMP_POINT, COMP_KEY, and COMP_TYPE variables are assigned values as described above (see Bash Variables).  If a shell function is being invoked, the COMP_WORDS and COMP_CWORD variables are also set. When the function or command is invoked, the first argument ($1) is the name of the command whose arguments are being completed, the second argument ($2) is the word being completed, and the third argument ($3) is the word preceding the word being completed on the current command line

I can only explain what I understand, so: `readline` is splitting the line into tokens, based on `COMP_WORDBREAKS`. The `COMP_CWORD` is the word being completed in [`COMP_WORDS`][BASH-VAR-COMP_WORDS]. The `COMP_POINT` is the position of the cursor in `COMP_LINE` (this is irrelevant here).

The command `complete -o bashdefault -o nospace -o dirnames -o filenames -F _my_completion x` simply register `_my_completion` to complete the command `x`.

Producing a completion is mostly about reading the current word, and matching it against a list of possible completions: `_my_completion` will take some parameters passed by the completion subsystem (see quote above). Normally, you would add some `COMPREPLY+=( "something" )` to produce a completion result. 

But in my example, I simply decided to `echo` a string to debug the completion.

With Git for Windows 2.25.1/2.26.2/2.27.0 (I decided to investigate more, hence the other versions): the `g` command is used instead of `x`, but it does not change the core problem.

- With `COMP_WORDBREAKS` to `"'@><=;|&(:`
- Bash will position
  - `$1` (command): will be `g`
  - `$2` (current word): will be `@z`
  - `$3` (word preceding): will be `@`
  - `COMP_WORDS` to [`g`, `@`, `z`]
- `_get_comp_words_by_ref -n @= cur prev words` will position:
  - `cur` to `@z`
  - `prev` to `g`
  - `words` to [`g`, `@z`]
- `_get_comp_words_by_ref -n = cur prev words` will position:
  - `cur` to `z`
  - `prev` to `@`
  - `words` to [`g`, `@`, `z`]

If now I go for `g --abricot @ad ppp @A<TAB>`, then the result is quite different:

- With `COMP_WORDBREAKS` to `"'@><=;|&(:`
- Bash will position
  - `$1` (command): will be `g`
  - `$2` (current word): will be `@A`
  - `$3` (word preceding): will be `@`
  - `COMP_WORDS` to `[0]="g" [1]="--abricot" [2]="@" [3]="ad" [4]="ppp" [5]="@" [6]="A"`
- `_get_comp_words_by_ref -n @= cur prev words` will position:
  - `cur` to `ppp@A`
  - `prev` to `--abricot@ad`
  - `words` to [`g`, `--abricot@ad`, `ppp@A`]
- `_get_comp_words_by_ref -n = cur prev words` will position:
  - `cur` to `A`
  - `prev` to `@`
  - `words` to [`g`, `--abricot`, `@`, `ad`, `ppp`, `@`, `A`]

The same with `bash-completion` using the PortableGit: this is how I expect it to work, and how it work on Linux (and on Linux, the git own completion uses [bash-completion] function).

- With `COMP_WORDBREAKS` to `"'@><=;|&(:` (bash-completion seems to remove the `@`; I could not find where, but before sourcing it or using PortableGit with my PR, you would first have to do `declare -p COMP_WORDBREAKS` to display the current value of `COMP_WORDBREAKS`).
- Bash will position
  - `$1` (command): will be `g`
  - `$2` (current word): will be `A` (this is not a mistake: even what bash provide changes, however this seems more valid to me).
  - `$3` (word preceding): will be `@`
  - `COMP_WORDS` to `[0]="g" [1]="--abricot" [2]="@" [3]="ad" [4]="ppp" [5]="@" [6]="A"`
- `_get_comp_words_by_ref -n @= cur prev words` will position:
  - `cur` to `@A`
  - `prev` to `ppp`
  - `words` to [`g`, `--abricot`, `@ad`, `ppp`, `@A`]
- `_get_comp_words_by_ref -n = cur prev words` will position:
  - `cur` to `A`
  - `prev` to `@`
  - `words` to [`g`, `--abricot`, `@`, `ad`, `ppp`, `@`, `A`]

**TL;DR:** I could simply remove `@` from `COMP_WORDBREAKS` but this is not something a completion script should do. This is less problematic in Git for Windows, and for my own script, so I could also do that (or clone `bash-completion`). 

> Since bash-completion is firmly in the "interactive usage" space, it is incorrect to include it even in MinGit.

I changed the part where it was defined: I check for `MINIMAL_GIT_WITH_BUSYBOX` and `MINIMAL_GIT` to be both empty.

And correct me if I'm wrong, is there a specific command to build those?

Do I simply have to execute `MINIMAL_GIT=1 ./portable/release.sh 2.27.1-test`?


[BASH-COMP]: https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion.html#Programmable-Completion
[BASH-VAR-COMP_WORDS]: https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-COMP_005fWORDS
[GENTOO-USE]: https://gentoo.org/support/use-flags/
[GENTOO-PKG]: https://packages.gentoo.org/packages/app-shells/bash-completion
[GH-BASHCOMP]: https://github.com/scop/bash-completion
[GH-BASHCOMP-LIST]: https://github.com/scop/bash-completion/tree/master/completions

dscho:
> I don't know what the final decision will be, but if you want me to prune those files, please confirm me what's the best place (I've seen that `make-file-list.sh` is grepping a lot of path, is it where files are pruned?).

A better alternative might be to identify the minimal set of files you absolutely need at the minimum.

> Do I simply have to execute `MINIMAL_GIT=1 ./portable/release.sh 2.27.1-test`?

The MinGit can be built via `./mingit/release.sh 0-test`.

linquize:
Why closed?

dscho:
> Why closed?

Because nothing happened in response to my suggestion, for well over half a year.

glhez:
I did not had the time to update, and with time I forgot about it. 

I also found that my problem was due to `hostcomplete` option - enabled by default - and which was adding `@` in `COMP_WORDBREAKS`. Since the `_get_comp_words_by_ref` in Git for Windows fails to reparse correctly the prev/next words when `COMP_WORDBREAKS` contains `@` (unless someone patched the completion upstream - but I'd say this is pretty unlikely : most linux distribution would have bash/zsh/whateversh completion AND git), that was sufficient for me...

@linquize do note that if you want `bash-completion`, you can still clone the `bash-completion` repo and load it in your ~/.bashrc or ~/.bash_profile. When investigating my `@` problem, this worked.

And to quote @dscho 

> But it is a cost, and since you seem to be the first person reporting problems that might be solved by including this, the benefit for the median Git for Windows user is almost certainly zero.

The `bash-completion` package certainly is useful, but you won't get all command for those completion to be a really nice an experience worth the additional ~150k. And you would be best in using MSYS2 directly, because it provides the command that Git for Windows won't provide (for example, you have `unzip` but not `zip` :)). 

I intended to add `bash-completion` in this PR because of my `@`, not because `bash-completion` provided something I lacked.

That's another reason why I did not insist much.

If I were to redo my PR (spoiler : I won't), that would be to:

- remove non working completions, which amount to `gapplication`, `gresource`, `sdk` and `vim` (this one really depends on bash-completion and `_filedir`...  
- disable hostname completion (`set -u hostcomplete`) in default `/etc/bash_profile`. I think this completion is pretty useless without a [HOSTFILE](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-HOSTFILE) which is unset by default. `bash-completion` unset it directly to replace it with a function : https://github.com/scop/bash-completion/blob/133790c464c68fa728b2957a66ee1fa012181a88/bash_completion#L1528

PR #294, opened 2020-06-12
Add support for Git Credential Manager Core as a credential helper option

mjcheetham:
Add Git Credential Manager (GCM) Core to the Git for Windows installer.
GCM Core is the cross-platform credential helper that will eventually
replace the exiting GCM for Windows (and GCM for Mac/Linux) helpers.

Move the "Enable Git Credential Manager" option to its own, new page
where you can select between the three options: none, GCM, or GCM Core.

Reusing the same "Use Credential Manager" replay choice key, we can
ensure that existing choices for GCM for Windows are carried forwards to
the new three-option choice. Enabled = GCM for Windows, Disabled = None,
and a new option Core = GCM Core.

Because GCM Core is still in beta we do not make it the default yet; GCM
for Windows is the default.

image

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -352,10 +352,14 @@ const
     GP_GitPullRebase  = 2;
     GP_GitPullFFOnly  = 3;
 
+    // Git Credential Manager settings.
+    GCM_None          = 1;
+    GCM_Classic       = 2;
+    GCM_Core          = 3;
+
     // Extra options
     GP_FSCache        = 1;
-    GP_GCM            = 2;
-    GP_Symlinks       = 3;
+    GP_Symlinks       = 2;
 
 #ifdef WITH_EXPERIMENTAL_BUILTIN_DIFFTOOL
 #define HAVE_EXPERIMENTAL_OPTIONS 1
@@ -446,6 +450,10 @@ var
     // Wizard page and variables for the `git pull` options.
     GitPullBehaviorPage:TWizardPage;
     RdbGitPullBehavior:array[GP_GitPullMerge..GP_GitPullFFOnly] of TRadioButton;
+
+    // Wizard page and variables for the credential manager options.
+    GitCredentialManagerPage:TWizardPage;
+    RdbGitCredentialManager:array[GCM_None..GCM_Core] of TRadioButton;
 
     // Wizard page and variables for the extra options.
     ExtraOptionsPage:TWizardPage;
@@ -796,6 +804,7 @@ begin
                     'credential.helper':
                         case Value of
                             'manager': RecordInferredDefault('Use Credential Manager','Enabled');
+                            'manager-core': RecordInferredDefault('Use Credential Manager','Core');
                         else RecordInferredDefault('Use Credential Manager','Disabled');
                         end;
                     'core.symlinks':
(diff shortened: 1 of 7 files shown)


PhilipOakley:
Is there a wider justification to be added to the commit - it's worth starting the commit explanation even if it's still wip. 

Don't assume everyone is familiar with the abbreviations, or the potential benefits, and don't forget to include any documentation improvement that will guide the users to use the benefits...

dscho:
I agree that there could be a justification (it's easy to see: Git Credential Manager for Windows is limited to Windows, whereas the Core version can replace that totally stale Java version of the Git Credential Manager).

Also, could you make this a radio button choice between GCM Windows, GCM Core and "leave unconfigured", moving it to a new page?

mjcheetham:
Hi @PhilipOakley, sorry for not being more descriptive with this change. It is indeed a work-in-progress, and was directed at @dscho for some initial comments which we also had offline - I realise this lack of context for others on GitHub is not a good thing. I hoped creating the PR as draft would also help indicate as such.

The benefits are that we plan to deprecate the existing Git Credential Manager for Windows and replace it with Git Credential Manager Core, the cross-platform replacement which will receive more support, fixes and features in the future.

I wanted to get Git Credential Manager Core into the Git for Windows installer as an experimental option at first to help provide a limited rollout of the component that is yet to see wide-spread testing.

@dscho I agree that a separate page for selecting a credential helper ("None/unset", "GCM for Windows", or "GCM Core [experimental]") would probably be better. We should also be able to reuse the existing stored preference for "Use Credential Manager" and extend it to include a "core" value.

installer/install.iss
```diff
@@ -2788,16 +2810,28 @@ begin
         GitSystemConfigSet('pull.ff','only')
     end;
 
+    {
+        Configure credential helper
+    }
+
+    if RdbGitCredentialManager[GCM_None].checked then begin
+        GitSystemConfigSet('credential.helper',#0);
```
mjcheetham, line 2818: I'm not sure if we want the option to be "None" which forcibly removes any configured helper, or change it to "Unset" which does nothing?

dscho: Since we're talking about the system config, I think `None` is appropriate (and technically, the same as `Unset`): it's not like we leave previously-configured settings in there, otherwise you would not be able to turn off GCM during a reinstallation.

installer/install.iss
```diff
@@ -2788,16 +2810,28 @@ begin
...
+        GitSystemConfigSet('credential.https://dev.azure.com.useHttpPath',#0);
+    end else if RdbGitCredentialManager[GCM_Classic].checked then begin
+        GitSystemConfigSet('credential.helper','manager');
+        GitSystemConfigSet('credential.https://dev.azure.com.useHttpPath',#0);
```
dscho, line 2822: Sneaky ;-)

mingw-w64-git-credential-manager-core/PKGBUILD
```diff
@@ -0,0 +1,30 @@
...
+license=('MIT')
+groups=('VCS')
+
+source=("${zip_url}" "$src_zip_url")
+
+sha256sums=('a9d424bb56e5571a9a80e195376aa7c7fcf5f890fe1029f45eb81be04eb9d264'
+            '7b07bf3317ae4e47446a93febe2b6fcde6f409bd3ab3b78358b6dfcc13ab5592')
+
+package() {
+  builddir=build-${MINGW_CHOST}
```
dscho, line 24: We don't have a build directory, right?

mingw-w64-git-credential-manager-core/PKGBUILD
```diff
@@ -0,0 +1,30 @@
+# Maintainer: Johannes Schindelin/Matthew J Cheetham
+
+_realname="git-credential-manager-core"
+pkgbase="mingw-w64-${_realname}"
+pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
+pkgver=2.0.164.25618
+pkgrel=1
```
dscho, line 7: Would you mind setting this to `0` in the beginning? The convention in Git for Windows' SDK is to increase the `pkgrel` (or increment `pkgver` and reset `pkgrel`) as part of the build that deploys the package.

While at it, could I ask you to [separate this new package definition out](https://git-scm.com/docs/git-rebase#_splitting_commits) into its own commit?

please.sh
```diff
@@ -2927,6 +2936,41 @@ upgrade () { # [--directory=<artifacts-directory>] [--only-mingw] [--no-build] [
...
+		 srcdir2="$(unzip -l $zip_prefix$version.zip | sed -n \
+		   's/^.\{28\} *\(.*\/\)\?git-credential-manager-core.exe/\1/p')" &&
+		 sed -i -e 's/^\(  srcdir2=\).*/\1"${srcdir}\/'$srcdir2'"/' \
+			PKGBUILD &&
+		 maybe_force_pkgrel "$force_pkgrel" &&
+		 git commit -s -m "Upgrade $package to $version${force_pkgrel:+-$force_pkgrel}" PKGBUILD &&
+		 create_bundle_artifact) &&
+		url=https://github.com/$repo/releases/tag/$tag_name &&
+		release_notes_feature='Comes with [Git Credential Manager Core v'$version']('"$url"').'
+		;;
```
dscho, line 2973: Great!

Ideally, I would like to see this change to `please.sh` as a second of three commits, teaching the administrative script to build and deploy the package.

mingw-w64-git-credential-manager-core/PKGBUILD
mjcheetham: Removed.

please.sh
mjcheetham: Done!

mingw-w64-git-credential-manager-core/PKGBUILD
mjcheetham: Done and done!

installer/install.iss
mjcheetham: Good sneaky or bad sneaky? This setting is required for GCM Core since it doesn't include the "walk the process tree" workaround that GCM for Windows does.

dscho: Ah... That's a good explanation ;-)

dscho:
Awesome job, @mjcheetham!

yyjdelete:
@mjcheetham @dscho 
Question: Seems the option is not shown when `Only show new options` is enabled in new released git-for-windows 2.28.0-x64?

rimrul:
Yes. That's due to the way `Only show new options` works. It doesn't realise there's a new answer for the credential manager. It only realises that we previously asked you about the credential manager.

nkjackzhang:
When i use `git help credential-manager-core`, i get error like ` Can't find file C:/Program Files/Git/mingw64/share/doc/git-doc/git-credential-manager-core.html`, is it missing?

rimrul:
I don't think such a help file exists.

PR #295, opened 2020-06-12
Fix recall of experimental pseudo console support setting

mjcheetham:
The Pseudo Console support experimental option was not being recalled
correctly on subsequent installs. If the option was unset it would
default to unchecked (correct), but if the option was set to 'disabled'
then it would become default checked!


dscho:
Thank you @mjcheetham!

PR #297, opened 2020-06-25
sdk create-desktop-icon: use correct path to Desktop

dscho:
We used to hard-code the default Desktop path, but that can be overridden (and often is, e.g. when the user follows OneDrive's suggestion to back up the Desktop folder on OneDrive).

This fixes https://github.com/git-for-windows/git-sdk-64/issues/22

### DIFF
--- a/git-extra/git-sdk.sh
+++ b/git-extra/git-sdk.sh
@@ -71,7 +71,8 @@ sdk () {
 		x86_64) bitness=" 64-bit";;
 		*) bitness=;;
 		esac &&
-		desktop_icon_path="$USERPROFILE/Desktop/Git SDK$bitness.lnk" &&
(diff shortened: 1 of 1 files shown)


dscho:
>  I wasn't sure that I understood the sed substitution with the quoting and escaping, or the `'y/...` command.

The `sed` invocation first replaces backslashes by forward slashes (`y/\\/\//` is the `y/<from>/<to>/` command that essentially does the same as the `tr` command), and then only prints out the part after `REG_SZ` (and only that line).

Even if you did not set up OneDrive (or, _especially_ because you did not), it would be good to have confirmation that it manages to recreate the Desktop link for you...

PhilipOakley:
Ok, so downloaded and checked out your branch. It both compiled the `Git-0-test-64-bit.exe` (after a an upgrade or two), and also the `sdk create-desktop-icon` worked (once I'd renamed my old ones)
👍 
`

PR #298, opened 2020-06-26
Installer: drop special treatment for Windows versions prior to Vista

rimrul:
We've dropped support for Windows versions older than Windows Vista
years ago. There's no reason to keep these special code paths  in the
installer just to work around missing features in these old versions.

Since we error out early on old Windows versions, this code was unreachable.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -553,66 +553,26 @@ begin
         Exit;
     end;
 
-    GetWindowsVersionEx(Version);
-
-    // Use the Restart Manager API when installing the shell extension on Windows Vista and above.
-    if Version.Major>=6 then begin
-        SetArrayLength(Modules,17);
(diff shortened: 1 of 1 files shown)


installer/install.iss
```diff
@@ -2353,24 +2313,13 @@ begin
...
-                ,   IDNO
-                )=IDYES);
-            end;
+			Result:=(SuppressibleMsgBox(
+				'If you continue without closing the listed applications they will be closed and restarted automatically.' + #13 + #13 +
+				'Are you sure you want to continue?'
+			,   mbConfirmation
+			,   MB_YESNO
+			,   IDYES
+			)=IDYES);
```
rimrul, line 2322: I've messed up here. should be spaces, not tabs.

PR #299, opened 2020-07-02
installer: allow Pseudo Console support to be re-disabled

dscho:
As noted by @satsuper in https://github.com/git-for-windows/git/issues/2729#issuecomment-653090026, when a user enabled support for Pseudo Consoles and then reinstalled to disable it, by mistake the Pseudo Console support would _still_ be enabled.

The reason? We only write `/etc/git-bash.config` to enable the Pseudo Console support, but not when we disable it (because it is disabled by default, anyway). However, we failed to remove that file upon uninstall/upgrade.


dscho:
I'll just merge this and let the Git Artifacts Pipeline build new installers, then test ;-)

PR #301, opened 2020-08-11
Fix persisting the choice in the credential helper selector

dscho:
Apparently with the upgrade to mingw-w64's GCC v10.x, we have to be a bit more strict with our use of the string format specifier (`%s` will render UTF-8 strings, even when specified e.g. in an `swprintf()` format, we need to use `%S` for wide Unicode strings).

This fixes https://github.com/git-for-windows/git/issues/2776

While at it, we now support toggling the checkbox via keyboard (<kbd>Alt</kbd>+<kbd>A</kbd>).


dscho:
@bbolli thank you!

PR #302, opened 2020-08-12
Added known issue about command hints for non-POSIX shells.

WildCard65:
Closes git-for-windows/git#2785

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -23,7 +23,7 @@ See [http://git-scm.com/](http://git-scm.com/) for further details about Git inc
 * As Git for Windows is shipped without Python support, `git p4` (which is backed by a Python script) is not supported.
 * The Quick Launch icon will only be installed for the user running setup (typically the Administrator). This is a technical restriction and will not change.
 * Older versions of the Windows Explorer do *not* calculate Git for Windows' on-disk size correctly, as it is unaware of hard links. Therefore, it might look like Git for Windows takes up 1.5GB when in reality it is about a third of that.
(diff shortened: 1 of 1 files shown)


ReleaseNotes.md
```diff
@@ -23,6 +23,7 @@ See [http://git-scm.com/](http://git-scm.com/) for further details about Git inc
 * As Git for Windows is shipped without Python support, `git p4` (which is backed by a Python script) is not supported.
 * The Quick Launch icon will only be installed for the user running setup (typically the Administrator). This is a technical restriction and will not change.
 * Older versions of the Windows Explorer do *not* calculate Git for Windows' on-disk size correctly, as it is unaware of hard links. Therefore, it might look like Git for Windows takes up 1.5GB when in reality it is about a third of that.
+* GIT command hints are designed for a POSIX shell, this can lead to issues when using them **AS IS** in non-POSIX shells like PowerShell. ([GH-2785](https://github.com/git-for-windows/git/issues/2785))
```
dscho, line 26: Would you mind adjusting the spelling of "Git" to how it is spelled elsewhere? It is not _really_ an acronym. Also, I would like to have "as is" lower-case (let's not be _that_ alarming), and I would much rather use "such as this ticket" as link text than "GH-2785").

WildCard65:
@dscho Done.

dscho:

For future record, we do prefer real names in the Signed-off-by: lines (for code contributions on Git's source code it is actually required), and we also recommend amending and force-pushing over piling fixup commits on top.

But in this instance, it seems overkill to require all that stringency.

ReleaseNotes.md
danielFreitasExternal: politeness class :)

PR #303, opened 2020-09-05
installer: only use SW_HIDE on custom editors that use a wrapper script

rimrul:
We're currently assuming that directly called editors ignore our request
to hide their initial window, but some editors honor that request.
This causes the installer to get stuck waiting for user input in a
hidden window.

This fixes git-for-windows/git#2684.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1483,6 +1483,8 @@ var
     InputText,OutputText:AnsiString;
     TmpFile:String;
     Res:Longint;
+    CustomEditorExt:String;
+    Show:Integer;
 begin
     if not PathIsValidExecutable(CustomEditorPath) then begin
         Wizardform.NextButton.Enabled:=False;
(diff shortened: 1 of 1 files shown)


installer/install.iss
```diff
@@ -1494,7 +1496,13 @@ begin
     InputText:='Please modify this text, e.g. delete it, then save it and exit the editor.'
     SaveStringToFile(TmpFile,InputText,False);
 
-    if not ShellExec('',CustomEditorPath,CustomEditorOptions+' "'+TmpFile+'"','',SW_HIDE,ewWaitUntilTerminated,Res) then begin
+    CustomEditorExt:=ExtractFileExt(CustomEditorPath);
+    if (CompareText(CustomEditorExt,'.bat')=0) or (CompareText(CustomEditorExt,'.cmd')=0) then
```
dscho, line 1500: Do we need to do this for `.ps1` scripts, too?

rimrul: No. Powershell launches them fine even with `SW_HIDE`. They aren't on `PATHEXT` by default, though and their default associaton is to open them with `notepad`.

dscho: Thank you for clarifying!

PR #305, opened 2020-09-28
Upgrade users to Git Credential Manager Core

dscho:
As per [GitHub's announcement](https://github.blog/changelog/2019-08-08-password-based-http-basic-authentication-deprecation-and-removal), password-based authentication is deprecated and will be removed in the near future.

Git Credential Manager for Windows (which is currently still the recommended default in Git for Windows) only supports that method when authenticating with GitHub, and it is unlikely that it will ever learn any other method: Git Credential Manager Core is the designated successor and all of the development effort goes into that project.

So let's upgrade Git Credential Manager for Windows users to Git Credential Manager Core in Git for Windows v2.29.0; upgrading in any later version would most likely be missing the date when GitHub turns off password-based authentication, and that in turn would most likely cause a lot of turmoil with our users.

### DIFF
--- a/git-extra/git-credential-helper-selector.c
+++ b/git-extra/git-credential-helper-selector.c
@@ -491,7 +491,7 @@ out_of_memory:
 	selected_helper = 0;
 
 	if (!previously_selected_helper)
-		previously_selected_helper = L"manager";
+		previously_selected_helper = L"manager-core";
 
 	for (p = env; *p; ) {
 		WCHAR *q = wcschr(p, L';');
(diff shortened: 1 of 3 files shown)


mjcheetham (approved):
Looks good to me! One question, do we want to swap the order of GCMW and GCMCore on the "select a helper" page?

dscho:
> One question, do we want to swap the order of GCMW and GCMCore on the "select a helper" page?

Good idea! Done. It looks like this now: 
image

mjcheetham (approved):
:shipit:

PR #308, opened 2020-10-02
Prevent missing DLLs in Git for Windows

dscho:
A recent snapshot was missing `libzstd.dll` although it was required to load `libcurl-4.dll`. With this script, which I intend to hook up in an automated build, we should be able to safeguard against similar issues in the future.

While at it, clean up the exclusion patterns in `make-file-list.sh` a bit more.

### DIFF
--- a/check-for-missing-dlls.sh
+++ b/check-for-missing-dlls.sh
@@ -1,0 +1,63 @@
+#!/bin/sh
+
+die () {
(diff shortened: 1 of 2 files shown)


jeffhostetler (approved):
Kind of amazing that you can do this much wizardry with so little code...

PR #309, opened 2020-10-05
Exclude unneeded DLLs from the installer/MinGit

dscho:
This PR enhances the script we recently introduced to verify that we're not missing any `.dll` files, to _also_ verify that we're not shipping superfluous `.dll` files either.

### DIFF
--- a/check-for-missing-dlls.sh
+++ b/check-for-missing-dlls.sh
@@ -30,6 +30,9 @@ else
 	next_line='\n'
 fi
 
+used_dlls_file=/tmp/used-dlls.$$.txt
+trap "rm \"$used_dlls_file\"" EXIT
+
 all_files="$(export ARCH BITNESS && "$thisdir"/make-file-list.sh | tr A-Z a-z)" &&
 usr_bin_dlls="$(echo "$all_files" | grep '^usr/bin/[^/]*\.dll$')" &&
 mingw_bin_dlls="$(echo "$all_files" | grep '^mingw'$BITNESS'/bin/[^/]*\.dll$')" &&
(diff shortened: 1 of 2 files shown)


check-for-missing-dlls.sh
```diff
@@ -67,3 +72,19 @@ do
...
+	grep '\.dll$' |
+	grep -v \
+		-e "$used_dlls_regex" \
+		-e '^usr/lib/perl5/' \
+		-e '^usr/lib/gawk/' \
+		-e '^usr/lib/openssl/engines' \
+		-e '^usr/lib/sasl2/' \
+		-e '^mingw../libexec/git-core/\(atlassian\|azuredevops\|bitbucket\|github\|microsoft\|newtonsoft\)\.' \
+		-e '^mingw../lib/\(engines\|reg\|thread\)' |
+	sed 's/^/unused dll: /' >&2
```
jeffhostetler, line 90: GH is complaining about no EOL on the last line.

jeffhostetler (approved):
I added a `set +x +v` locally to `check-for-missing-dlls.sh` and looked thru the output and
got a pretty good understanding of the magic you're casting here.  I didn't dig thru each
third-party directory exclusion that you've listed at the bottom, but I would think that if
one of them were wrong, we'd see lots of messages about it.

So I'm happy with this.

Thanks!!!

check-for-missing-dlls.sh
dscho: That's because it is not needed, and I edited this file in VS Code (which makes it much easier to drop a trailing newline than, say, `vi`).

dscho:
> I didn't dig thru each
> third-party directory exclusion that you've listed at the bottom, but I would think that if
> one of them were wrong, we'd see lots of messages about it.

Yep, that's the point of the `check-for-missing-dlls.sh` automation: we would know _very_ quickly about it ;-)

PR #310, opened 2020-10-06
installer: default to GCM Core (including fresh installs, not just upgrades)

dscho:
I completely missed the fact that we still install GCM(W) by default... 😠


mjcheetham (approved):
This looks correct, but I haven't built and tested myself.

PR #314, opened 2020-10-16
installer: two fixes regarding the default branch name choice

dscho:
Yesterday, after I had released Git for Windows v2.29.0-rc2, I realized that I had been asked to choose whether I want to override the default branch name, even if I had checked the "Only show new options" checkbox, and even if I had previously said that I want to stay with Git's default.

Turns out that this choice was not recorded at all, due to an under-documented feature of the `SetPreviousData(Key,Value)` function: when specifying the empty string as value, it is not recorded at all; Instead, any previously recorded data for that key is deleted.

While working around this feature, I also noticed how easy it is to specify an invalid branch name, so I added code to prevent that from being accepted.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1899,7 +1899,8 @@ begin
     // Restore the setting chosen during a previous install.
     Data:=ReplayChoice('Default Branch Option','');
     case Data of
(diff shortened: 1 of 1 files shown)


dscho:
I've verified manually that this works.

PR #315, opened 2020-10-18
[RFC PATCH] Installer: Suggest also '.gitattributes' EOL setting

PhilipOakley:
fixes: #2808

Inform users that cross-platform repositories should also have a
`.gitattributes` file containing "* text=auto" as described in
https://www.edwardthomson.com/blog/git_for_windows_line_endings.html.

Hence Git-for-Windows Windows users will have text files converted from
Windows style line endings (\r\n) to Unix style line endings (\n) when
they’re added to the repository, even if they have not set "core.autocrlf".

This is an untested draft to follow up G-f-W #2808

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -2068,7 +2068,7 @@ begin
     CRLFPage:=CreatePage(PrevPageID,'Configuring the line ending conversions','How should Git treat line endings in text files?',TabOrder,Top,Left);
 
     // 1st choice
-    RdbCRLF[GC_CRLFAlways]:=CreateRadioButton(CRLFPage,'Checkout Windows-style, commit Unix-style line endings','Git will convert LF to CRLF when checking out text files. When committing'+#13+'text files, CRLF will be converted to LF. For cross-platform projects,'+#13+'this is the recommended setting on Windows ("core.autocrlf" is set to "true").',TabOrder,Top,Left);
+    RdbCRLF[GC_CRLFAlways]:=CreateRadioButton(CRLFPage,'Checkout Windows-style, commit Unix-style line endings','Git will convert LF to CRLF when checking out text files. When committing'+#13+'text files, CRLF will be converted to LF. For cross-platform projects,'+#13+'this is the recommended setting on Windows ("core.autocrlf" is set to "true").'+#13+#13+'Also consider including a ".gitattributes" file'+#13+'containing "* text=auto" in all cross-platform repositories.',TabOrder,Top,Left);
 
     // 2nd choice
     RdbCRLF[GC_LFOnly]:=CreateRadioButton(CRLFPage,'Checkout as-is, commit Unix-style line endings','Git will not perform any conversion when checking out text files. When'+#13+'committing text files, CRLF will be converted to LF. For cross-platform projects,'+#13+'this is the recommended setting on Unix ("core.autocrlf" is set to "input").',TabOrder,Top,Left);


dscho:
It's a good change, but hardly a fix for git-for-windows/git#2808... As I had pointed out in that ticket, there is much work to do to detect changes in the eol settings of a file after it was written.

dscho:
Oh, and: could you paste a screenshot of the new page? (use `installer/release.sh -d CRLF` for testing)

dscho:
FWIW I intend tn merge this PR before releasing Git for Windows v2.29.0 (probably tonight).

PhilipOakley:
Hi @dscho 

Screenshot:
image

May need a bit more space (height) on the dialog box. ..

I was/am having problems with the Msys2/Mingw part of the `$ sdk build git-and-installer` at the point where it switches to doing various `mingw` bits
```
make: Leaving directory '/usr/src/git'
error: mingw32: key "4A6129F4E4B84AE46ED7F635628F528CF3053E04" is unknown
:: Import PGP key 4A6129F4E4B84AE46ED7F635628F528CF3053E04? [Y/n]
```


I tried the tricks at https://www.msys2.org/news/#2020-06-29-new-packagers but no luck for the git-and-installer build, though your `installer/release.sh -d` did work for the screen shot. PGP issues isn't something I know much about.

dscho:
You'll have to use `pacman-key --recv-keys <key>` and then `... --lsign-key ...` to resolve that GPG problem.

About that missing space... that's a real problem. I don't want any existing information to be lost.

PhilipOakley:
> About that missing space... that's a real problem. I don't want any existing information to be lost.

True, I'm trying to understand how the inno-setup `CreateRadioButton` works with respect to the `Top` variable and how (if) that affects the spacing. At the moment it's a bit unclear. 

I was hoping that there was a direct parameter asking for the required amount of space (I think it (`Top`) drops through to https://github.com/git-for-windows/build-extra/blob/503688af95491122b79d0375df63666c1042a37d/installer/install.iss#L1408), but what that does I'm not sure. And It's not clear what the 'right' way to tweak it would be.

rimrul:
If all else fails we could do a dropdown and the description of the selected element, like with the editors.

PhilipOakley:
> About that missing space... that's a real problem. I don't want any existing information to be lost.

True, I'm trying to understand how the inno-setup `CreateRadioButton` works with respect to the `Top` variable and how (if) that affects the spacing. At the moment it's a bit unclear. 

I was hoping that there was a direct parameter asking for the required amount of space 

- Argh above text never posted.

maybe

@@ -1402,7 +1402,7 @@ begin
     Result.Left:=ScaleX(Left);
     Result.Top:=ScaleY(Top);
     Result.Width:=ScaleX(405);
-    Result.Height:=ScaleY(17);
+    Result.Height:=ScaleY(21);
     Result.TabOrder:=TabOrder;
     TabOrder:=TabOrd

Maybe needed to tweak the dialog height (Untested, guessed change!)

At the moment I'm stuck on the build because (I think) Msys2_32bit has been deprecated, and the packager has changed so the old PGP key is invalid 
`error: key "4A6129F4E4B84AE46ED7F635628F528CF3053E04" could not be looked up remotely` so all the sdk auto builds are failing at that step.

PhilipOakley:
Despite the GPG errors, I eventually got the hack to compile, but no luck, the dialog box size hadn't changed.  

The InnoSetup help is no help, so it's back to StackOverfow and friends for clues as to chat to change to increase the dialog box height. (`SurfaceHeight` is the current property being investigated)

image

dscho:
> the dialog box size hadn't changed.

I don't think that there is an easy way to do that.

And I am starting to be rather doubtful that this is the right spot to tell users about this.

In any case, I think we should expend more energy changing Git itself so that its behavior is more helpful in the first place.

PhilipOakley:
> > the dialog box size hadn't changed.
> 
> I don't think that there is an easy way to do that.

That's a shame. I couldn't find anything either.

> And I am starting to be rather doubtful that this is the right spot to tell users about this.

I'm still pro the idea of improving the active documentation (stuff users see), so, apart from the current lack of space, I'd still be pro the advice hint.

> In any case, I think we should expend more energy changing Git itself so that its behavior is more helpful in the first place.

I suspect that is a long uphill task, though if there is an outline plan then that would make it easier.

In summary, it looks like InnoSetup's limitations have halted this suggestion.

dscho:
The problem with bombarding users with information is that they stop reading very quickly.

dscho:
@PhilipOakley could we put this suggestion maybe into an advice in the Git executable proper, detecting the situation at hand? That would probably have a much higher impact that adding some notes to the installer that people read (if the read them at all) when installing Git for Windows rather than when they work in their repositories?

dscho:
> 
> 
> @PhilipOakley could we put this suggestion maybe into an advice in the Git executable proper, detecting the situation at hand? That would probably have a much higher impact that adding some notes to the installer that people read (if the read them at all) when installing Git for Windows rather than when they work in their repositories?

@PhilipOakley ping?

PhilipOakley:
> @PhilipOakley could we put this suggestion maybe into an advice in the Git executable proper, detecting the situation at hand? That would probably have a much higher impact that adding some notes to the installer that people read (if the read them at all) when installing Git for Windows rather than when they work in their repositories?

Hi @dscho, Sorry, I haven't had much Git time recently, what with Covid etc.

Looking back at the discussion I see that in https://github.com/git-for-windows/git/issues/2808#issuecomment-709628741 you said 

>  the index does not record the `autocrlf` setting with which entries were checked out (and as a consequence, Git cannot realize that `core.autocrlf` has changed in the meantime and warn the user about it.

.. so I'm not sure how you hoped that we'd detect the original problem at source. Did you have an outline of what you were thinking of?

I had thought (when I saw the ping) that this was a follow on to https://github.com/git-for-windows/git/issues/2994#issuecomment-766740284 had a similar underlying EOL problem.

dscho:
> > the index does not record the `autocrlf` setting with which entries were checked out (and as a consequence, Git cannot realize that `core.autocrlf` has changed in the meantime and warn the user about it.
> 
> .. so I'm not sure how you hoped that we'd detect the original problem at source. Did you have an outline of what you were thinking of?

The idea was to change Git in such a manner that it _would_ be possible to detect the change. Like, store the "last-seen `core.autocrlf`" in an index extension.

dscho:
@PhilipOakley have you looked into the possibility to store the EOL setting that was used in an index extension?

PhilipOakley:
> @PhilipOakley have you looked into the possibility to store the EOL setting that was used in an index extension?

Sorry, I've not looked at that approach (i.e what it might mean to code up an extension to the index and use it in a consistent manner). 
I'm not sure it is a complete solution for cases where the index is 'temporary' (i.e. gets deleted or rebuilt), whereby the record of the checkout is lost so we'd be back to square one in those [edge?] cases.

Thinking a bit more as I compose this, maybe it needs to be thought of as a (type of) clean/smudge filter problem, where we are detecting that the checked out version has been smudged (maybe with some percentage factor if there is a diff involved as well) . In this case it ("what's the smudge?") being that there is an core.autocrlf setting, and we compare to `lf->crlf`. Thus the index extn would note if a file was 'smudged' in some way. Though code doesn't solve the human social/learning problems.

But overall the whole issue is a can of worms.

dscho:
> I'm not sure it is a complete solution for cases where the index is 'temporary' (i.e. gets deleted or rebuilt), whereby the record of the checkout is lost so we'd be back to square one in those [edge?] cases.

I have troubles understanding how the temporary nature of an index has anything to do with line endings. The way I am thinking about this, the idea would work with temporary indices just the same (leaving aside the fact that most times you have a temporary index, no files get checked out, actually, and the entire line-ending issue is moot).

> Thinking a bit more as I compose this, maybe it needs to be thought of as a (type of) clean/smudge filter problem, where we are detecting that the checked out version has been smudged (maybe with some percentage factor if there is a diff involved as well) .

This line of thinking is not fruitful because all it does is keep us in the theoretical space, with no code being implemented, and nothing to test.

> But overall the whole issue is a can of worms.

We could attempt to close that can by closing this PR ;-)

dscho:
> We could attempt to close that can by closing this PR ;-)

@PhilipOakley do you agree?

PhilipOakley:
The appropriate actions for managing EOL conversion on Git-for-Windows and other cross platform repository situations is still an open ended, long-standing issue. Let's leave the users to their own local solutions and I'll close this PR.

PR #319, opened 2020-12-10
A couple fixes for the credential helper selector

dscho:
In a conversation with @mjcheetham I realized that neither `%s` nor `%S` are appropriate format specifiers for wide-character string arguments in `printf()`/`wprintf()` calls. Instead, `%ls` is the best option.

After fixing this, I noticed in my tests that the helper selector constructed a command line with improperly quoted paths. That's the other fix included in this PR.


mjcheetham (approved):
Looks good to me!

abuwely (commented):
abuwely

Ollodriss:
Très très important

PR #320, opened 2020-12-11
Update aliases.sh

michkot:
Recent windows versions have the improved console which reports as xterm-256color (and also Windows Terminal reports the same), and they definitely do not need winpty to translate anything. This PR fixes to check so that it applies solely to mintty. Also, given that mintty is not shipped with GfW (anymore?) maybe remove this file completely?

### DIFF
--- a/git-extra/aliases.sh
+++ b/git-extra/aliases.sh
@@ -5,8 +5,8 @@
 alias ls='ls -F --color=auto --show-control-chars'
 alias ll='ls -l'
 
-case "$TERM" in
-xterm*)
+case "$TERM_PROGRAM" in
+mintty*)
 	# The following programs are known to require a Win32 Console
 	# for interactive usage, therefore let's launch them through winpty
 	# when run inside `mintty`.


dscho:
`mintty.exe` is still shipped, and it is the default terminal for Git Bash in the Start menu.

So... render me puzzled.

rimrul:
They're saying that there is no need for WinPTY when Git-Bash is configured to use Conhost or when running `usr/bin/bash` from Windows terminal.

They also say that Windows Terminal and Conhost on up-to-date Windows 10 report `TERM` as some `xterm` variant.

They thus changed the `aliases.sh` to not check wether `TERM` contains `xterm` but instead check `TERM_PROGRAM` for `mintty`.

I don't know if this works or not, but the gist of it is "If not running Mintty, don't configure Winpty aliases".

dscho:
@michkot care to clarify?

michkot:
Sure, I will clarify!
First, sorry, I didn't really check the "mintty is not shipped with git for windows" in depth, I just skimmed trough the directories and did not notice mintty is under /usr/bin (and I forgot I, probably, choose to not use it be default in the installer). 
And yes, exactly as rimrul summarized: Even the update/release notes for version 2.27.0 since 2.26.2 https://github.com/git-for-windows/build-extra/blob/main/ReleaseNotes.md#changes-since-git-for-windows-v2262-april-20th-2020 states:
> This release comes with a Git Bash that optionally uses Windows-native pseudo consoles. Meaning: finally, Git Bash can accommodate console programs like node.exe, Python or PHP, without using the winpty helper (see Known Issues above). Note that this is still a very new feature and is therefore known to have some corner-case bugs.
However the file in question of this PR makes the choice whether to add alias to user's bash based on the value of $TERM containing "xterm"...however native console terminal, from Windows 10 1809+ I think, reports as an xterm-variant too. 
Switching to $TERM_PROGRAM would be better.

dscho:
The most important (and still unadressed) question for me is: how reliable is that `TERM_PROGRAM` tell-tale? I.e. Are there circumstances when it is not set? Or are there any scenarios where it is set but no MinTTY is running (such as: Windows Terminal/VS Code being called _from MinTTY_)?

michkot:
> 
> 
> The most important (and still unadressed) question for me is: how reliable is that `TERM_PROGRAM` tell-tale? I.e. Are there circumstances when it is not set? Or are there any scenarios where it is set but no MinTTY is running (such as: Windows Terminal/VS Code being called _from MinTTY_)?

Cross-terminal call is a thing I did not consider. I've did a brief examination and it seems that:
- both TERM and TERM_PROGRAM are passed (including their values), as expected, from mintty-hosted bash to cmd/wt
- ~~I do not understand the typical process of TERM getting set to some value,~~ the xterm-256color gets actually set by MSYS2 now: https://github.com/msys2/msys2-runtime/blob/6642d4755392b2c99c9fd20b9e409bf2d9b4aa3d/winsup/cygwin/environ.cc#L916-L942 ; per that code, WindowsTerminal or conhost (includes VScode) hosted bash only gets TERM set to "xterm-256color" if it **does not exists beforehand**
- - => If I start mintty-bash, then from it, I start "cmd-bash", then from it, I start mintty-bash, I will get TERM=xterm and TERM_PROGRAM=mintty in **all** bash instances :-/

Conclusion: 
TERM=xterm and(or) TERM_PROGRAM=mintty means "this either runs from inside mintty **or** runs in something else (including native console/wt/vscode) but has been started indirectly from mintty". 
TERM=xterm-256colors **might** mean "is running in latest conhost/openconsole(WinTerminal)", but it definitely should not be mintty.

Recommendation: 
change the aforementioned check to either check for "xterm" explicitly (did NOT tested nor researched that) or to check for TERM_PROGRAM. Either of those seem to be a better alternative to the one currently used.

dscho:
I'm still slightly concerned that `git -c alias.n="!echo $TERM" n` when run in, say, Git CMD shows a value (`cygwin`) while `git -c alias.n="!echo $TERM_PROGRAM" n` shows an empty value. Meaning: `TERM` is typically overridden, but `TERM_PROGRAM` is usually left unchanged when run outside of MinTTY.

Having said that, I am willing to risk the fall-out if you (@michkot) update the commit message to stop claiming that MinTTY is no longer shipped, as well as including a summary of this PR's discussion  in the commit message (extra brownie points if you wrap the commit message at <=76 columns/line).

dscho:
>  I am willing to risk the fall-out if you (@michkot) update the commit message to stop claiming that MinTTY is no longer shipped, as well as including a summary of this PR's discussion in the commit message (extra brownie points if you wrap the commit message at <=76 columns/line).

@michkot are you still interested in this PR?

michkot:
Yes, I've actually forgot about it for a while...
 But I got back to the topic 3 days ago and I think I should do a bit more
testing, since (not) using winpty generated some awkward behaviour when
using a specific python2 script. I will definitely incorporate the commit
message changes.

dscho:
I think it will need more than just commit message massaging: the environment variable seems not to be unset/overridden in all scenarios when we need it to be.

dscho:
This PR has become stale.

PR #323, opened 2021-01-22
portable: add arm64 artifact option

dennisameling:
Builds further on https://github.com/git-for-windows/build-extra/pull/321. Very much a proof of concept, just sharing here for inspiration.

**Example release based on this PR**: https://github.com/dennisameling/git/releases/tag/2.30.0-beta2

Results in an actual working portable installer for ARM64 when taking the following steps:

- `sdk build mingw-w64-git`
- `pacman -U mingw-w64-i686-git-2.30.0.2.f8cbc844b8-1-any.pkg.tar.xz` (replace the filename with the actual filename that's generated)
- `sdk build git-extra` (so that `/etc/profile` gets overwritten to include ARM64)
- `pacman -U git-extra-1.1.513.20ab9a5-1-i686.pkg.tar.xz` (replace the filename with the actual filename that's generated)
- ([create the arm64 binaries](https://github.com/git-for-windows/git/pull/2915#issuecomment-739538291) now and output to `C:\git-sdk-32\arm64`)
- Go into `/usr/src/build-extra/portable` and run  `mkdir -p arm64 && ./release.sh --include-arm64-artifacts=/arm64 --output=./arm64 2.30.0`

It then looks like this on an ARM64 device:

image

### DIFF
--- a/please.sh
+++ b/please.sh
@@ -4770,12 +4770,13 @@ build_mingw_w64_git () { # [--only-32-bit] [--only-64-bit] [--skip-test-artifact
 }
 
 # This function does not "clean up" after installing the packages
-make_installers_from_mingw_w64_git () { # [--pkg=<package>[,<package>...]] [--version=<version>] [--installer] [--portable] [--mingit] [--mingit-busybox] [--nuget] [--nuget-mingit] [--archive]
+make_installers_from_mingw_w64_git () { # [--pkg=<package>[,<package>...]] [--version=<version>] [--installer] [--portable] [--mingit] [--mingit-busybox] [--nuget] [--nuget-mingit] [--archive] [--include-arm64-artifacts=<path>]
 	modes=
 	install_package=
 	output=
 	output_path=
 	version=0-test
+	include_arm64_artifacts=
 	while case "$1" in
 	--pkg=*)
 		install_package="${install_package:+$install_package }$(echo "${1#*=}" | tr , ' ')"
@@ -4820,6 +4821,9 @@ make_installers_from_mingw_w64_git () { # [--pkg=<package>[,<package>...]] [--ve
 		output_path="$(cygpath -am "${1#-?}")"
 		output="--output=$output_path" || exit
 		;;
+	--include-arm64-artifacts=*)
+		include_arm64_artifacts="$1"
+		;;
 	-*) die "Unknown option: %s\n" "$1";;
 	*) break;;
 	esac; do shift; done
@@ -4855,7 +4859,7 @@ make_installers_from_mingw_w64_git () { # [--pkg=<package>[,<package>...]] [--ve
 		test installer != $mode ||
 		extra="${extra:+$extra }--window-title-version=$version"
 
-		sh -x "${this_script_path%/*}/$mode/release.sh" $output $extra $version
+		sh -x "${this_script_path%/*}/$mode/release.sh" $output $extra $include_arm64_artifacts $version
 	done
 }
 
--- a/portable/release.sh
+++ b/portable/release.sh
@@ -22,6 +22,9 @@ do
 	--include-pdbs)
 		include_pdbs=t
 		;;
+	--include-arm64-artifacts=*)
+		arm64_artifacts_directory="${1#*=}"
+		;;
 	-*)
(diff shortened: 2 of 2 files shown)


portable/release.sh
```diff
@@ -134,6 +144,12 @@ test -z "$include_pdbs" || {
 } ||
 die "Could not unpack .pdb files"
 
+# ARM64 Windows handling
+if test -n "$cross_compiling_arm64"
+then
+	cp -ar /arm64 "$SCRIPT_PATH/root"
```
dscho, line 150: If we allow specifying an argument (like `--output <directory>`), we do not have to hard-code `/arm64` here.

I think I like that idea.

And if we go that route, the option should probably be renamed to `--include-arm64-artifacts=<directory>` or some such.

dennisameling: Changed in https://github.com/git-for-windows/build-extra/pull/323/commits/6e5a2e74b5fbe829a7c93c82c0cab9487a8d5ad0. Can now run `./release.sh --include-arm64-artifacts=/arm64 --output=./arm64 2.30.0` (the `--output` option was already there and I simply use it to easily find back the generated `.7z.exe` file 😊)

Output:

```
$ ./release.sh --include-arm64-artifacts=/arm64 --output=./arm64 2.30.0
Including ARM64 artifacts from /arm64
7za is /usr/bin/7za
Creating archive

7-Zip (a) [32] 16.02 : Copyright (c) 1999-2016 Igor Pavlov : 2016-05-21
p7zip Version 16.02 (locale=nl_NL.UTF-8,Utf16=on,HugeFiles=on,32 bits,8 CPUs Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz (906ED),ASM,AES-NI)

Scanning the drive:
92 folders, 6209 files, 708249591 bytes (676 MiB)

Creating archive: /tmp.7z

Items to compress: 6301


Files read from disk: 6203
Archive size: 48514387 bytes (47 MiB)
Everything is Ok
Success! You will find the new installer at "./arm64/PortableGit-2.30.0-arm64.7z.exe".
It is a self-extracting .7z archive.
```

dennisameling:
Just updated `please.sh` to add an `--extra-args` argument, which is then passed to the appropriate `release.sh` script - details in https://github.com/git-for-windows/build-extra/pull/323/commits/dc69222faab114018b8a10a6d418b0c857ddadc8

Example invocation in https://github.com/dennisameling/git/runs/1799017569?check_suite_focus=true:

```
/usr/src/build-extra/please.sh make_installers_from_mingw_w64_git --extra-args="--include-arm64-artifacts=${{github.workspace}}/arm64" --version=`$(cat pkg-i686/ver) -o artifacts --${{matrix.artifact.name}} --pkg=pkg-i686/mingw-w64-i686-git-[0-9]*.tar.xz --pkg=pkg-i686/mingw-w64-i686-git-doc-html-[0-9]*.tar.xz 
```

... which correctly results in a PortableGit executable that includes the `/arm64` binaries 🎉

please.sh
```diff
@@ -4820,6 +4821,9 @@ make_installers_from_mingw_w64_git () { # [--pkg=<package>[,<package>...]] [--ve
 		output_path="$(cygpath -am "${1#-?}")"
 		output="--output=$output_path" || exit
 		;;
+	--extra-args=*)
```
dscho, line 4824: How about calling this `--include-arm64-artifacts=*` instead? I do not necessarily want that `make_installers_from_mingw64_w64_git` function to learn to pass arbitrary flags to the `release.sh` scripts. Something like

```sh
--include-arm64-artifacts=*)
    include_arm64_artifacts="$1"
    ;;
[...]
sh -x "${this_script_path%/*}/$mode/release.sh" $output $extra $include_arm64_artifacts $version
```

dennisameling: Sure! Thought it'd be convenient to have a more generic function, but I get that it might add complexity/potential for issues. Done 👍🏼

dennisameling:
@dscho just applied your suggestion and squashed the commits 👍🏼 should be good to go now!

dscho:
I adjusted this only slightly before merging. Thank you!

PR #324, opened 2021-01-23
mingit: add arm64 artifact option

dennisameling:
Related to https://github.com/git-for-windows/build-extra/pull/323, needs https://github.com/git-for-windows/build-extra/pull/321 & #323 

Example release based on this PR: https://github.com/dennisameling/git/releases/tag/2.30.0-beta2

This probably needs optimization. The generated ZIP is 207MB 😅 any tips/guidance are/is welcome!

### GitHub Desktop / Dugite Native / Dugite
The mingit packages are used by [desktop/dugite-native](https://github.com/desktop/dugite-native) > [desktop/dugite](https://github.com/desktop/dugite) > [desktop/desktop](https://github.com/desktop/desktop) (GitHub Desktop), so I tested with those as well.

⚠️ Note: `dugite-native` is still on Git 2.26.2, so the failing tests might be due to the bump to 2.30.0 as well

dugite-native: https://github.com/desktop/dugite-native/pull/347
dugite: https://github.com/desktop/dugite/pull/421

PR #325, opened 2021-01-31
archive: add arm64 artifact option

dennisameling:
Needs https://github.com/git-for-windows/build-extra/pull/323

Adds an --include-arm64-artifacts argument to the archive script, just like for portable (#323) and mingit (#324)


dscho:
Thanks!

PR #326, opened 2021-02-04
installer: add arm64 artifact option

dennisameling:
Adds ARM64 support to the installer 🚀 

Currently installs ARM64 Git into `C:\Program Files (x86)\Git` by default.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -34,7 +34,11 @@ LZMAUseSeparateProcess=yes
 OutputBaseFilename={#FILENAME_VERSION}
 OutputDir={#GetEnv('TEMP')}
 #else
+#ifdef INSTALLER_FILENAME_SUFFIX
+OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION+'-'+INSTALLER_FILENAME_SUFFIX}
+#else
 OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION}-{#BITNESS}-bit
+#endif
 #ifdef OUTPUT_DIRECTORY
 OutputDir={#OUTPUT_DIRECTORY}
 #else
--- a/installer/release.sh
+++ b/installer/release.sh
@@ -58,6 +58,13 @@ do
 	--include-pdbs)
 		include_pdbs=t
 		;;
+	--include-arm64-artifacts=*)
+		case "${1#*=}" in
+		/*) ;; # absolute path, okay
+		*) die "Need an absolute path: $1";;
+		esac
+		arm64_artifacts_directory="${1#*=}"
+		;;
 	*)
 		break
 	esac
@@ -255,6 +262,17 @@ test -z "$include_pdbs" || {
 		>> file-list.iss
 } ||
 die "Could not include .pdb files"
+
+test -z "$arm64_artifacts_directory" || {
+	echo "Including ARM64 artifacts from $arm64_artifacts_directory" &&
+	inno_defines="$inno_defines$LF#define INSTALLER_FILENAME_SUFFIX 'arm64'" &&
+	mixed="$(cygpath -m "$arm64_artifacts_directory")" &&
+	find "$arm64_artifacts_directory" -type f |
+	sed -e "s|^$arm64_artifacts_directory\\(/.*\)\?/\([^/]*\)$|Source: \"$mixed\\1/\\2\"; DestDir: {app}/arm64\\1; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore|" \
+		-e 's|/|\\|g' \
+		>> file-list.iss
+} ||
+die "Could not include ARM64 artifacts"
 
 etc_gitconfig_dir="${etc_gitconfig%/gitconfig}"
 printf "%s\n%s\n%s\n%s\n%s%s" \


installer/release.sh
```diff
@@ -218,6 +221,23 @@ test -z "$GITCONFIG_PATH" || {
...
+# ARM64 Windows handling
+IS_ARM64=
+if test -n "$arm64_artifacts_directory"
+then
+	echo "Including ARM64 artifacts from $arm64_artifacts_directory";
+	IS_ARM64=1
+	# If directory starts with a slash (i.e. /arm64), strip it, because InnoSetup can't handle it.
+	if [[ $arm64_artifacts_directory == *\/* ]];
+	then
+		ARM64_FILES=$(find ${arm64_artifacts_directory} -type f | cut -c2-)
```
dennisameling, line 233: You probably have a much cleaner solution, @dscho 😊 it does work with this setup though. The installer is created successfully and includes the ARM64 artifacts 🚀

installer/release.sh
```diff
@@ -218,6 +221,23 @@ test -z "$GITCONFIG_PATH" || {
...
 }
 
+# ARM64 Windows handling
+IS_ARM64=
+if test -n "$arm64_artifacts_directory"
+then
+	echo "Including ARM64 artifacts from $arm64_artifacts_directory";
+	IS_ARM64=1
+	# If directory starts with a slash (i.e. /arm64), strip it, because InnoSetup can't handle it.
+	if [[ $arm64_artifacts_directory == *\/* ]];
```
dscho, line 231: Do you really want to catch paths starting with a `/`? If so, let's do this instead:

```sh
case "$arm64_artifacts_directory" in
/*)
    arm64_artifacts_directory=${arm64_artifacts_directory#/}
    ;;
esac
```

But since the `${...#<prefix>}` construct will strip the given prefix only when found, that `case` is unnecessary. Even better: we do not need to modify the variable, but use that construct instead below.

installer/release.sh
```diff
@@ -218,6 +221,23 @@ test -z "$GITCONFIG_PATH" || {
...
+	then
+		ARM64_FILES=$(find ${arm64_artifacts_directory} -type f | cut -c2-)
+	else
+		ARM64_FILES=$(find ${arm64_artifacts_directory} -type f)
+	fi
+
+	LIST="${LIST}"$'\n'"${ARM64_FILES}"
```
dscho, line 238: Let's do it directly:

```sh
LIST="$LIST$LF$(find "$arm64_artifacts_directory#/" -type f)"
```

installer/install.iss
```diff
@@ -34,7 +34,11 @@ LZMAUseSeparateProcess=yes
 OutputBaseFilename={#FILENAME_VERSION}
 OutputDir={#GetEnv('TEMP')}
 #else
+#if IS_ARM64==1
+OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION}-arm64
+#else
 OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION}-{#BITNESS}-bit
+#endif
```
dscho, line 41: Hmm. It would probably make more sense to make this more generic: it is simply a file name suffix. Something like this:

```c
#ifdef INSTALLER_FILENAME_SUFFIX
 OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION-INSTALLER_FILENAME_SUFFIX}
#else
OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION}-{#BITNESS}-bit
#endif
```

And then `#define INSTALLER_FILENAME_SUFFIX 'arm64'` in `config.iss` instead.

dscho (changes requested):
Great progress!

dscho:
There is one more concern that I forgot to mention earlier: if we try to include artifacts from a directory outside of the pseudo root, the given approach won't work.

It would probably make more sense to imitate the `$include_pdbs` approach (but you don't need to copy, you just need to set the `Source:` attribute to the actual (Windows) path).

I admit that the `sed` call in the `$include_pdbs` clause is not for the faint of heart ;-)

dennisameling:
Alright, just applied your suggestions! Should be good to go now 😊

installer/release.sh
```diff
@@ -58,6 +58,9 @@ do
 	--include-pdbs)
 		include_pdbs=t
 		;;
+	--include-arm64-artifacts=*)
+		arm64_artifacts_directory="${1#*=}"
```
dscho, line 62: I just realized that due to the `cd` above, we cannot allow relative paths here. Therefore, I would like to recommend adding something like this to that clause:

```sh
case "${1#*=}" in
/*) ;; # absolute path, okay
*) die "Need an absolute path: $1";;
esac
```

installer/release.sh
```diff
@@ -256,6 +259,20 @@ test -z "$include_pdbs" || {
...
+test -z "$arm64_artifacts_directory" || {
+	echo "Including ARM64 artifacts from $arm64_artifacts_directory" &&
+	inno_defines="$inno_defines$LF#define INSTALLER_FILENAME_SUFFIX 'arm64'" &&
+	rm -rf root &&
+	mkdir root &&
+	cp -ar $arm64_artifacts_directory/* root &&
+	find root -type f |
+	sed -e 's|/|\\|g' \
+		-e 's|^root\\\([^\\]*\)$|Source: "{#SourcePath}\\root\\\1"; DestDir: {app}\\arm64; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore|' \
+		-e 's|^root\\\(.*\)\\\([^\\]*\)$|Source: "{#SourcePath}\\root\\\1\\\2"; DestDir: {app}\\arm64\\\1; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore|' \
```
dscho, line 271: I think something like this should work, without copying any files:

```sh
mixed="$(cygpath -m "$arm64_artifacts_directory")" &&
find "$arm64_artifacts_directory" -type f |
sed -e "s|^$arm64_artifacts_directory\\(/.*\)\?/\([^/]*\)$|Source: \"$mixed\\1/\\2\"; DestDir: \"{app}/arm64\\1; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore|" \
    -e 's|/|\\|g'
```

dscho (changes requested):
Getting there! ;-)

dennisameling:
Thanks! Just applied your suggestions @dscho. Tested and just had to apply a very minimal change to the `sed` command. Works now without copying the files first. Didn't know you can use `cygpath` to convert a Unix to a Windows path; learning something new every day 😊

dscho (approved):
Great!

PR #327, opened 2021-02-06
[WIP] installer: Update InnoSetup from 5.6.1 to 6.1.2

dennisameling:
https://github.com/git-for-windows/build-extra/pull/326 installs Git for Windows in the 32-bit Program Files folder (`C:\Program Files (x86)`), but InnoSetup 6 added support for installing in the 64-bit folder for Windows on ARM:

> 6.0: Support for 64-bit mode on Windows 10 on ARM.

While not a hard requirement, it would be nice to have future versions of Git for Windows on ARM64 install in the 64-bit app directory by default.

They also [bumped the minimal Windows version to Vista](https://jrsoftware.org/isdl.php#stable), but that should be fine given the fact that Git for Windows [also requires Vista](https://gitforwindows.org/requirements.html) or later.

Most important changes are listed at https://jrsoftware.org/isdl.php#stable, and the full changelog is listed at https://jrsoftware.org/files/is6-whatsnew.htm

## Testing on Windows on ARM

After applying this PR and building the installer, Git will be installed to `C:\Program Files\Git` instead of `C:\Program Files (x86)\Git` on Windows on ARM:

image

One thing I noticed is that on Windows on ARM, both Git Credential Manager options are greyed out:

image


installer/install.iss
```diff
@@ -87,7 +87,7 @@ WizardImageBackColor=clWhite
 WizardImageStretch=no
 WizardImageFile={#SourcePath}\git.bmp
 WizardSmallImageFile={#SourcePath}\gitsmall.bmp
-MinVersion=0,5.01sp3
+MinVersion=6.0
```
dennisameling, line 90: Bumped the MinVersion to 6.0 (Vista)

installer/install.iss
```diff
@@ -3366,7 +3366,6 @@ begin
         Form.Caption:='Git Uninstall: Removing in-use files';
         Form.ClientWidth:=ScaleX(500);
         Form.ClientHeight:=ScaleY(256);
-        Form.Center;
```
dennisameling, line 3369: Had to remove this due to the following error:

```
Launching Inno Setup compiler ...
Error on line 3369 in C:\git-sdk-32\usr\src\build-extra\installer\install.iss: Column 14:
Unknown identifier 'CENTER'
```

installer/update-inno-setup.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # Download the most recent Inno Setup version.
-installer="is-unicode.exe"
+installer="is.exe"
```
dennisameling, line 4: There's no longer an `is-unicode.exe` file. [Details](https://jrsoftware.org/isdl.php#stable):

> Change in default behavior: Starting with Inno Setup 6 there's only one version available: Unicode Inno Setup. Unicode Inno Setup has been available for 9 years but in case you have not yet updated to it: please see the Unicode Inno Setup topic in the help file for more information. Basically, unless you're using [Code] to make DLL calls with string parameters you shouldn't have to make any changes to your script.

installer/install.iss
rimrul: I guess that means we could drop our own check wether we're running on vista or later.

dennisameling: It needs to stay there, otherwise InnoSetup [will default to](https://jrsoftware.org/isdl.php) `6.1sp1`:

> **Change in default behavior**: Starting with Inno Setup 6.1 the [Setup] section directive MinVersion defaults to 6.1sp1, so **by default Setup will not run on Windows Vista or on versions of Windows 7 and Windows Server 2008 R2 which have not been updated**. Setting MinVersion to 6.0 to allow Setup to run on Windows Vista is supported but not recommended: Windows Vista doesn't support some of Setup's security measures against potential DLL preloading attacks so these have to be removed by the compiler if MinVersion is below 6.1 making your installer less secure on all versions of Windows.

rimrul: No, i meant since InnoSetups requirement and this `MinVersion=6.0` check are in place, we don't need this check anymore: https://github.com/git-for-windows/build-extra/blob/dbd9c2cf7dc82558922d97a0261e7fab150979e6/installer/install.iss#L1000-L1006

rimrul:
>One thing I noticed is that on Windows on ARM, both Git Credential Manager options are greyed out:

Seems like we can't [detect .NET Framework](https://github.com/git-for-windows/build-extra/blob/dbd9c2cf7dc82558922d97a0261e7fab150979e6/installer/install.iss#L1083-L1090) 4.5.1 or newer

https://github.com/git-for-windows/build-extra/blob/dbd9c2cf7dc82558922d97a0261e7fab150979e6/installer/install.iss#L1083-L1090

https://github.com/git-for-windows/build-extra/blob/dbd9c2cf7dc82558922d97a0261e7fab150979e6/installer/install.iss#L2172-L2179

dennisameling:
@rimrul thanks for that info! I just checked and this is the reason it fails:

**x64 machine**
This machine has two registry keys, one for 32-bit and one for the 64-bit .NET framework:

32-bit: `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\NET Framework Setup\NDP\v4\Full`
64-bit: `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full`

**ARM64 machine**
This machine only has one registry key, because .NET Framework 4 isn't available natively for ARM64 and 64-bit (x64) emulation is only available in prerelease versions of Windows 10:

32-bit: `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\NET Framework Setup\NDP\v4\Full`
64-bit: _MISSING_

InnoSetup will attempt to access the 64-bit registry [in 64-bit mode](https://jrsoftware.org/ishelp/index.php?topic=32vs64bitinstalls), so it makes sense that the value isn't found. .NET Framework _is_ installed on ARM64 though, only the 32-bit version.

To keep things clean, it might be an option to separate this PR into two:
1) Update InnoSetup to 6.1.2
2) Add 64-bit support for Windows ARM64

Otherwise it might become a bit challenging to review this PR. What do you think?

installer/install.iss
dscho: Heh, I did the same in dscho/build-extra@674f32017f43d596f9e49c3a390598aea0c92c07...

installer/update-inno-setup.sh
dscho: I did the same in dscho/build-extra@commit 114305fcd781c6ea3149436896faa53307dab92c, and I also added `//CURRENTUSER`, no idea why, I forgot the details (probably because it's been over a year that I dabbled with the InnoSetup upgrade).

dscho:
> To keep things clean, it might be an option to separate this PR into two:
> 
>     1. Update InnoSetup to 6.1.2, not changing any other functionality
> 
>     2. Add 64-bit support for Windows ARM64
> 
> 
> Otherwise it might become a bit challenging to review this PR.

I think that it's a fine plan to split this work up into two separate PRs.

dennisameling:
Closing this PR as it's superseded by https://github.com/git-for-windows/build-extra/pull/327

PR #329, opened 2021-02-24
installer: bump NetFx version requirement from 4.5.1 -> 4.6.1

y2kbugger:
We already require Fx4.6.1, but the installer checks for Fx4.5.1

Signed-off-by: Zak Kohler <git@y2kbugger.com>

I used the _lower_ of the two 4.6.1 checks as that is how 4.5.1 was done previously.

@mjcheetham 

> .NET Framework 4.6.1
> On Windows 10 November Update systems: 394254
> On all other Windows operating systems (including Windows 10): 394271

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -2287,12 +2287,12 @@ begin
     // No credential helper
     RdbGitCredentialManager[GCM_None]:=CreateRadioButton(GitCredentialManagerPage,'None','Do not use a credential helper.',TabOrder,Top,Left);
 
-    // Restore the settings chosen during a previous install, if .NET 4.7.2
+    // Restore the settings chosen during a previous install, if .NET Framework 4.7.2
     // or later is available.
     if DetectNetFxVersion()<461808 then begin
         RdbGitCredentialManager[GCM].Checked:=False;
         RdbGitCredentialManager[GCM].Enabled:=False;
-        RdbGitCredentialManager[GCM].Caption:=RdbGitCredentialManager[GCM].Caption+' (requires .NET >= 4.7.2)'
+        RdbGitCredentialManager[GCM].Caption:=RdbGitCredentialManager[GCM].Caption+' (requires .NET Framework >= 4.7.2)'
     end else begin
         case ReplayChoice('Use Credential Manager','Enabled') of
             'Disabled': RdbGitCredentialManager[GCM_None].Checked:=True;
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -2292,6 +2292,7 @@ begin
     if DetectNetFxVersion()<378675 then begin
         RdbGitCredentialManager[GCM].Checked:=False;
         RdbGitCredentialManager[GCM].Enabled:=False;
+        RdbGitCredentialManager[GCM].Caption:=RdbGitCredentialManager[GCM].Caption+' (requires .NET >= 4.5.1)'
     end else begin
         case ReplayChoice('Use Credential Manager','Enabled') of
             'Disabled': RdbGitCredentialManager[GCM_None].Checked:=True;
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1174,8 +1174,8 @@ end;
 
 function DetectNetFxVersion:Cardinal;
 begin
-    // We are only interested in version v4.5.1 or later, therefore it
-    // is enough to only use the 4.5 method described in
+    // We are only interested in version v4.7.2 or later, therefore it
+    // is enough to only use the "4.5 and later" method described in
     // https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed
     if IsWin64 then begin
         if (
@@ -2287,12 +2287,12 @@ begin
     // No credential helper
     RdbGitCredentialManager[GCM_None]:=CreateRadioButton(GitCredentialManagerPage,'None','Do not use a credential helper.',TabOrder,Top,Left);
 
-    // Restore the settings chosen during a previous install, if .NET 4.5.1
+    // Restore the settings chosen during a previous install, if .NET 4.7.2
     // or later is available.
-    if DetectNetFxVersion()<378675 then begin
+    if DetectNetFxVersion()<461808 then begin
         RdbGitCredentialManager[GCM].Checked:=False;
         RdbGitCredentialManager[GCM].Enabled:=False;
-        RdbGitCredentialManager[GCM].Caption:=RdbGitCredentialManager[GCM].Caption+' (requires .NET >= 4.5.1)'
+        RdbGitCredentialManager[GCM].Caption:=RdbGitCredentialManager[GCM].Caption+' (requires .NET >= 4.7.2)'
     end else begin
         case ReplayChoice('Use Credential Manager','Enabled') of
             'Disabled': RdbGitCredentialManager[GCM_None].Checked:=True;


dscho:
> .NET Framework 4.6.1
> On Windows 10 November Update systems: 394254
> On all other Windows operating systems (including Windows 10): 394271

Where does that information come from?

Also, could you amend the commit message to start with "installer: bump NetFx version requirement from 4.5.1 -> 4.6.1", and then continue by specifying _what_ requires v4.6.1, accompanied by a link to the evidence (e.g. to the code in GCM Core or GCM for Windows that shows how it won't work with older versions)?

dscho:
Oh, sorry, I forgot to thank you for this valuable contribution!

I'm not the domain expert on this, so I'll leave it to @mjcheetham to give their 👍 before merging.

y2kbugger:
Version detection spec comes from here:
https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed

To be fair, @mjcheetham noticed the discrepancy, mentioned in an email when asking about another issue:

> One interesting thing to note.. the Git for Windows installer is checking for .NET Framework 4.5.1, but actually it should be checking for 4.6.1 (which is what GCM Core requires on Windows) – this is a bug

dscho:
> 
> 
> Version detection spec comes from here:
> https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed
> 
> To be fair, @mjcheetham noticed the discrepancy, mentioned in an email when asking about another issue:
> 
> > One interesting thing to note.. the Git for Windows installer is checking for .NET Framework 4.5.1, but actually it should be checking for 4.6.1 (which is what GCM Core requires on Windows) – this is a bug

Excellent! Could you add that information (together with a link to @mjcheetham's comment) into the commit message? Then we're golden.

installer/install.iss
```diff
@@ -2166,14 +2166,14 @@ begin
...
 
     // Git Credential Manager for Windows
-    RdbGitCredentialManager[GCM_Classic]:=CreateRadioButton(GitCredentialManagerPage,'Git Credential Manager','(DEPRECATED) The <A HREF=https://github.com/Microsoft/Git-Credential-Manager-for-Windows>Git Credential Manager for Windows</A> handles credentials e.g.'+#13+'for Azure DevOps and GitHub (requires .NET framework v4.5.1 or later).',TabOrder,Top,Left);
+    RdbGitCredentialManager[GCM_Classic]:=CreateRadioButton(GitCredentialManagerPage,'Git Credential Manager','(DEPRECATED) The <A HREF=https://github.com/Microsoft/Git-Credential-Manager-for-Windows>Git Credential Manager for Windows</A> handles credentials e.g.'+#13+'for Azure DevOps and GitHub (requires .NET framework v4.6.1 or later).',TabOrder,Top,Left);
```
rimrul, line 2169: This seems untrue. This text is about the deprecated [Git Credential Manager for Windows](https://github.com/microsoft/Git-Credential-Manager-for-Windows), not the newer [Git Credential Manager Core](https://github.com/microsoft/Git-Credential-Manager-Core) you mentioned in the commit message. This GCM Classic does seem to [only require .NET Framework 4.5.1](https://github.com/microsoft/Git-Credential-Manager-for-Windows/blob/master/Installer/Setup.iss#L211) as stated in the original Text.

The GCM Core text that should probably mention this reqjuirement is 3 lines above.

installer/install.iss
```diff
@@ -2166,14 +2166,14 @@ begin
...
 
     // No credential helper
     RdbGitCredentialManager[GCM_None]:=CreateRadioButton(GitCredentialManagerPage,'None','Do not use a credential helper.',TabOrder,Top,Left);
 
-    // Restore the settings chosen during a previous install, if .NET 4.5.1
+    // Restore the settings chosen during a previous install, if .NET 4.6.1
     // or later is available.
-    if DetectNetFxVersion()<378675 then begin
+    if DetectNetFxVersion()<394254 then begin
```
rimrul, line 2176: It might even make sense to split this check into 2 checks, to only disable GCM core, but not GCM classic if the .NET version is >= 4.51, but < 4.6.1

rimrul (dismissed):

y2kbugger:
Since this change isn't trivial like the first, I want to build and test that it works as expected. I'm looking into options for this.

installer/install.iss
```diff
@@ -2163,34 +2163,36 @@ begin
...
-    end else begin
-        case ReplayChoice('Use Credential Manager','Core') of
-            'Disabled': RdbGitCredentialManager[GCM_None].Checked:=True;
-            'Enabled': RdbGitCredentialManager[GCM_Classic].Checked:=True;
-            'Core': RdbGitCredentialManager[GCM_Core].Checked:=True;
-        else
-            RdbGitCredentialManager[GCM_Core].Checked:=True;
-        end;
-        // Auto-upgrade GCM to GCM Core in version v2.29.0
-        if RdbGitCredentialManager[GCM_Classic].Checked and ((PreviousGitForWindowsVersion='') or IsUpgrade('2.29.0')) then begin
```
dscho, line 2190: Why not keep this block here, but enhance the conditionals so that e.g. the auto-upgrade won't happen if `RdbGitCredentialManager[GCM_Core].Enabled` evaluated to `True`?

y2kbugger: My reason is that denesting the conditionals reduces complexity/ improves maintainability. The tradeoff is that the upgrade/replay execute regardless, and are the repealed by the version checks. but I think that is better than having a combinatory nested ifs.

If you look at structure of final code (rather than just diff) it is:

- Replay previous
- Upgrade to Core
- Version Guard for GCM Core
- Version Guard for GCM Classic


Did notice one thing that might need changed (based on testing): if both guards get triggered (and unchecked and disabled), do we need to manually set `[GCM_None].Checked:=True`, or do radios enforce that once option is chosen.

mjcheetham:
> I'm not the domain expert on this, so I'll leave it to @mjcheetham to give their 👍 before merging.

This is correct. GCM Core requires .NET Framework 4.6.1 be installed to run. GCM "classic" only requires 4.5.1.

Heads-up.. a future version (the next one?) of GCM Core will require .NET Framework 4.7.2.

Going even further in to the future, GCM Core will not require any version of .NET Framework (we will bundle/statically link the .NET Core CLR and libraries).

y2kbugger:
> My reason is that denesting the conditionals reduces complexity/ improves maintainability. The tradeoff is that the upgrade/replay execute regardless, and are the repealed by the version checks. but I think that is better than having a combinatorially nested ifs, as that would involve duplicating at least some of the checks.
> 
> If you look at structure of final code (rather than just diff) it is:
> 
> * Replay previous
> * Upgrade to Core
> * Version Guard for GCM Core
> * Version Guard for GCM Classic
> 
> Did notice one thing that might need changed (based on testing): if both guards get triggered (and unchecked and disabled), do we need to manually set `[GCM_None].Checked:=True`, or do radios enforce that once option is chosen.

I think i made the required changes and unless my reasoning(or code) is either wrong or the other style is just preferred this is ready to be tested.

installer/install.iss
dscho: My rationale: if you put the version guard _before_ replying, you can automatically reject an invalid replay.

dscho:
> a future version (the next one?) of GCM Core will require .NET Framework 4.7.2.

@mjcheetham this happened with the latest release, correct?

mjcheetham:
> > a future version (the next one?) of GCM Core will require .NET Framework 4.7.2.
> 
> @mjcheetham this happened with the latest release, correct?

Yes. The current latest GCM Core release on Windows targets .NET Framework 4.7.2.
https://github.com/microsoft/Git-Credential-Manager-Core/blob/3fc6791abf670dd08c0620e8a6c642f7104c128a/src/windows/Payload.Windows/Payload.Windows.csproj#L6

installer/install.iss
dscho: On that note, @y2kbugger, have you tested what happens if `GCM_Core` is checked during a replay, but then disabled because of the version check (using a bogus minimum version to force it to be disabled would be the easiest way to test that, I think)?

dscho:
I guess now that we no longer ship the old Git Credential Manager for Windows, this PR would become substantially simpler?

mjcheetham:
> > I guess now that we no longer ship the old Git Credential Manager for Windows, this PR would become substantially simpler?
> 
> @mjcheetham I guess that was a question for you?

GCM "Core" still requires .NET Framework 4.7.2.

dscho:
> > > I guess now that we no longer ship the old Git Credential Manager for Windows, this PR would become substantially simpler?
> > 
> > 
> > @mjcheetham I guess that was a question for you?
> 
> GCM "Core" still requires .NET Framework 4.7.2.

Oh, so we have to change https://github.com/git-for-windows/build-extra/blob/19c17738d6ef48cbfaa4c3d286a9fed4e01ca9aa/installer/install.iss#L2290-L2292, but to what value?

rimrul:
> Oh, so we have to change
> 
> https://github.com/git-for-windows/build-extra/blob/19c17738d6ef48cbfaa4c3d286a9fed4e01ca9aa/installer/install.iss#L2290-L2292
> , but to what value?

Probably 461808. See https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#detect-net-framework-45-and-later-versions

installer/install.iss
```diff
@@ -2287,11 +2287,12 @@ begin
...
     RdbGitCredentialManager[GCM_None]:=CreateRadioButton(GitCredentialManagerPage,'None','Do not use a credential helper.',TabOrder,Top,Left);
 
-    // Restore the settings chosen during a previous install, if .NET 4.5.1
+    // Restore the settings chosen during a previous install, if .NET 4.7.2
     // or later is available.
-    if DetectNetFxVersion()<378675 then begin
+    if DetectNetFxVersion()<461808 then begin
         RdbGitCredentialManager[GCM].Checked:=False;
         RdbGitCredentialManager[GCM].Enabled:=False;
+        RdbGitCredentialManager[GCM].Caption:=RdbGitCredentialManager[GCM].Caption+' (requires .NET >= 4.7.2)'
```
mjcheetham, line 2295: You'll want to specify .NET _Framework_ here rather than just ".NET" which is really the evolution of .NET Core after 3.x (Core being a separate product from Framework).

Otherwise someone may install ".NET 6" and wonder why it's still not working.

```suggestion
        RdbGitCredentialManager[GCM].Caption:=RdbGitCredentialManager[GCM].Caption+' (requires .NET Framework >= 4.7.2)'
```

Core CLR|Desktop CLR
-|-
.NET Core 1<br/>.NET Core 2.x<br/>.NET Core 3.x<br/>.NET 5<br/>.NET 6<br/>.NET 7|.NET Framework 1<br/>.NET Framework 2<br/>.NET Framework 3<br/>.NET Framework 3.5<br/>.NET Framework 4.0<br/>.NET Framework 4.5.x<br/>.NET Framework 4.6.x<br/>.NET Framework 4.7.x<br/>.NET Framework 4.8

GCM still needs the Desktop flavours of the CLR(!)

mjcheetham (changes requested):
Technically looks good (correct version check etc), but I'd change the UI to specify "Framework".

installer/install.iss
```diff
@@ -2292,6 +2292,7 @@ begin
     if DetectNetFxVersion()<378675 then begin
         RdbGitCredentialManager[GCM].Checked:=False;
         RdbGitCredentialManager[GCM].Enabled:=False;
+        RdbGitCredentialManager[GCM].Caption:=RdbGitCredentialManager[GCM].Caption+' (requires .NET >= 4.5.1)'
```
rimrul: This should say `.NET Framework`. It can't really be confused for any other .NET product with these twi version numbers in Question, but it's better to use accurate names with how many different .NET products/standards exist nowadays.

dscho: Thanks, addressed in https://github.com/git-for-windows/build-extra/pull/329/commits/5db650c0aec90ecd7e3c193bec51a18ad5c5ed90

installer/install.iss
dscho: Thanks, addressed in https://github.com/git-for-windows/build-extra/pull/329/commits/5db650c0aec90ecd7e3c193bec51a18ad5c5ed90 (I also touched up the code comment)

rimrul:
On a side-note: The CI seems to struggle a lot with installing the `build-installers` flavor of the SDK. I think it might be related to the fact that when I try to download the SDK artifact manually, it reliably slows to a crawl after around 20MB. This doesn't seem to happen with the `full-sdk` artifact, though.

mjcheetham (approved):
LGTM!

dscho:
> The CI seems to struggle a lot with installing the `build-installers` flavor of the SDK

Right, this is actually tracked in https://github.com/git-for-windows/setup-git-for-windows-sdk/issues/336... I just didn't have time yet to look into it more closely.

> I think it might be related to the fact that when I try to download the SDK artifact manually, it reliably slows to a crawl after around 20MB.

Potentially.

I think we should look into using partial clones, and parallel checkout, anyway. We could even just initialize an empty repository, get the commit SHA as well as the contents of `.sparse/` via the API, and then fetch the objects in a batch. It's a bit tricky to get this right so that the performance does not tank, but I think this might be better than going forward with these `git-sdk-64-extra-artifacts` job artifacts.

PR #332, opened 2021-03-16
Improve error message in silent mode when Git for Windows' files are in use

dscho:
In silent mode, we do not see the "Processes" page to which the error message refers:

image

Let's change the error message in silent mode, to be more helpful. Example:

An alternative I would have _loved_ to implement would have shown the wizard form via `WizardForm.Show()`, i.e. by turning the silent installer non-silent. However, clicking the <kbd>OK</kbd> button of the error message will cause the installer to exit anyway, i.e. it won't let the user interact with the wizard form before exiting, so that is not an option.

This fixes https://github.com/git-for-windows/git/issues/3118

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -2431,14 +2431,33 @@ begin
                         Log(Msg);
                     ExitEarlyWithSuccess();
                 end;
-                if WizardSilent() then begin
-                    Msg:='Silent setup failed because the following process(es) use Git for Windows:'+#13+#10;
-                    for j:=i to GetArrayLength(Processes)-1 do
-                        if not Processes[j].Restartable then
-                            Msg:=Msg+#13+#10+Processes[j].Name+' (PID '+IntToStr(Processes[j].ID)+')';
-                end else
+                if WizardSilent() then
+                    while True do begin
+                        Msg:='';
+                        for j:=i to GetArrayLength(Processes)-1 do
+                            if not Processes[j].Restartable then
+                                Msg:=Msg+#13+#10+Processes[j].Name+' (PID '+IntToStr(Processes[j].ID)+')'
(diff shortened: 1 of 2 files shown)


Tragetaschen:
Maybe I misunderstand, but

This only changes the error message, but the installer still exits and I have to wait for the toast to reappear or download and run the installer manually, right?

rimrul:
I was thinking about `SupressibleMsgBox(Msg, mbCriticalError, MB_RETRYCANCEL, IDCANCEL)` and then checking again, if the result is `IDRETRY`.

dscho:
> 
> 
> Maybe I misunderstand, but
> 
> This only changes the error message, but the installer still exits and I have to wait for the toast to reappear or download and run the installer manually, right?

Almost correct. The installer still exits, but you do not have to wait for the toast to reappear, you can run `git update-git-for-windows` again, or download manually.

There are obviously quite a few ways how we could make this more convenient. For example, we could show a Yes/No message in a loop that exits if either Yes was clicked or if (after refreshing) the process list contains no more processes.

Or we could offer to re-schedule. Or we could offer to run a non-silent installer from within that error message.

Plenty of options. But I think this PR is a first good step.

rimrul:
>But I think this PR is a first good step.

Definitely. We're not leading the user on an invisible wizard page chase anymore, and instead straight up tell them the offending processes.

rimrul:
>But I think this PR is a first good step.

Definitely. We're not leading the user on an invisible wizard page chase anymore, and instead straight up tell them the offending processes.

dscho:
How about this?

Tragetaschen:
> How about this?

Maybe include one of the options in https://github.com/git-for-windows/build-extra/pull/332#issuecomment-800192715 like the `git update-git-for-windows` command line in the text as well?

dscho:
I wasn't happy with the line breaks (and the single words per line), so I reworded it thusly:

PR #330, opened 2021-03-02
installer: Update Inno Setup from 5.6.1 to 6.1.2

dennisameling:
Successor of https://github.com/git-for-windows/build-extra/pull/327. As discussed in that PR, we'll make separate PRs for the Inno Setup update and later on for adding Windows ARM64 64-bit support (currently installs in `Program Files (x86)`). This PR focuses on the Inno Setup upgrade, without changing any other functionality. It's split up into multiple commits for easier reviewing.

CC @dscho

PR #333, opened 2021-03-21
installer: add 64-bit support for ARM64

dennisameling:
Builds further on https://github.com/git-for-windows/build-extra/pull/330

The Git for Windows ARM64 installation ideally should be installed in the 64-bit Program Files folder. Inno Setup 6.0 added support for 64-bit on Windows ARM64. This commit leverages this support.

The installation directory is correctly set to `C:\Program Files\Git`:

image

The Git Credential Manager radio buttons are enabled correctly in the 64-bit ARM64 installer:

image

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -26,6 +26,10 @@
 #define DEFAULT_BRANCH_NAME 'master'
 #endif
 
+#ifndef INSTALLER_FILENAME_SUFFIX
+#define INSTALLER_FILENAME_SUFFIX ''
+#endif
+
 [Setup]
 ; Compiler-related
 Compression=lzma2/ultra64
@@ -34,7 +38,7 @@ LZMAUseSeparateProcess=yes
 OutputBaseFilename={#FILENAME_VERSION}
 OutputDir={#GetEnv('TEMP')}
 #else
-#ifdef INSTALLER_FILENAME_SUFFIX
+#if INSTALLER_FILENAME_SUFFIX!=''
 OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION+'-'+INSTALLER_FILENAME_SUFFIX}
 #else
 OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION}-{#BITNESS}-bit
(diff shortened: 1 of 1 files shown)


installer/install.iss
```diff
@@ -1076,11 +1080,20 @@ end;
 
 function DetectNetFxVersion:Cardinal;
 begin
+#if INSTALLER_FILENAME_SUFFIX=='arm64'
+    // Windows ARM64 is only available starting with Windows 10 1803,
+    // which has .NET Framework 4.7.2 built-in, but only 32-bit (which
+    // causes the registry check below to return 0). Therefore, we
+    // hardcode the version here.
+    // https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies
+    Result:=461808;
```
rimrul, line 1089: Could AMD64 Windows also have only a 32 bit .NET framework installed? Should we try to additionally discover 32 bit .NET framework on 64 bit systems?

This workaround works for now, but will break once GCM core [moves their windows build to .NET 5](https://github.com/microsoft/Git-Credential-Manager-Core/issues/113).

dscho: Good point.

Is there maybe a way to check the 32-bit version of the registry as a fall-back, once the 64-bit test failed?

dscho: It [looks](https://jrsoftware.org/ishelp/index.php?topic=registrysection) as if `HKLM32` and `HKLM64` can be used for this purpose, and obviously their use should be guarded by `if IsWin64() then [...]`.

dennisameling: > Could AMD64 Windows also have only a 32 bit .NET framework installed?

I think there's a very small chance this will happen, because Windows 10 has both 32-bit and 64-bit .NET framework built-in on AMD64. When [downloading the runtime manually](https://dotnet.microsoft.com/download/dotnet-framework/net472), it installs 32-bit and 64-bit in one go.

However, it's rather easy to build the check with the suggestions @dscho mentioned. I got it to work, just updated this PR accordingly.

I just downloaded a Windows 7 32-bit VM for testing ([from the official Microsoft website](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/)), and with .NET framework installed it works:

image

When I remove .NET framework it correctly greys out the options:

image

Let me know if there's anything you want me to change in the logic 👍🏼

rimrul: I don't like how it has none of the radio buttons checked, but next still enabled in the no .NET case, but that's entirely unrelated to the changes and would happen like that on otherintel based cpus as well. It looks good to me.

rimrul (approved):

dscho (approved):
At first, I was a bit concerned about defining `INSTALLER_FILENAME_SUFFIX` when we don't want any suffix, but it does seem like the cleanest solution (because unlike a proper C preprocessor, `#if defined(INSTALLER_FILENAME_SUFFIX) && INSTALLER_FILENAME_SUFFIX!=''` would still result in the error message `Undeclared identifier: INSTALLER_FILENAME_SUFFIX.`).

dscho:
@dennisameling thank you so much!

PR #335, opened 2021-04-07
Don't modify HKCU\Console root key

adrianghc:
There is no reason why the installer should not only create the `HKCU\Console\Git Bash` and `HKCU\Console\Git CMD` keys, but also modify the `HKCU\Console` key.
It needlessly overwrites user settings and is redundant because of the Git-specific keys.

Signed-off-by: Adrian Georg Herrmann <adrian.ghc@outlook.com>

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -107,7 +107,7 @@ Name: ext\guihere; Description: Git GUI Here; Types: default
 Name: gitlfs; Description: Git LFS (Large File Support); Types: default; Flags: disablenouninstallwarning
 Name: assoc; Description: Associate .git* configuration files with the default text editor; Types: default
 Name: assoc_sh; Description: Associate .sh files to be run with Bash; Types: default
-Name: consolefont; Description: Use a TrueType font in all console windows
+Name: consolefont; Description: Use a TrueType font in all console windows; OnlyBelowVersion: 10.0
 Name: autoupdate; Description: Check daily for Git for Windows updates
 
 [Run]


rimrul (changes requested):
If we remove the "consolefont" feature, we should at least remove it fully.

adrianghc:
Yes, that makes sense. Is it better now?

dscho:
> There is no reason why the installer should not only create the `HKCU\Console\Git Bash` and `HKCU\Console\Git CMD` keys, but also modify the `HKCU\Console` key.

I have to admit that I am having a real hard time to parse this. Could you express this without a double negative?

Also, I do not understand the rationale. You are saying that there is no `Console` key in HKCU? But the comment "There is no "Console" key in HKLM" (which is left stale by this PR, by the way) seems to suggest that it exists _somewhere_, and that somewhere was most likely intended to be HKCU...

As to the commit removing the `consolefont` feature: its commit message does a pretty good job at saying what it does. However, commit messages' job is to say _why_ they do what they do, the _what_ is usually pretty obvious from their diff already.

For that reason, I would like to ask for "making a case" (i.e. explain the rationale well) in the commit message as to why this is a good change that we should accept.

adrianghc:
I feel that I owe you an apology, as I have realized upon closer inspection that my rationale for this pull request was based on a misunderstanding. Let me explain it to you:

For some time I have been bugged by the fact that my installer would always revert global console settings that were inside 
`HKCU\Console`. Today I thought I would look into it. Searching the repository, I found where those keys were set. And for sure, remove that part from the installer, no more overwriting! But the thing is that I attributed the overwriting of keys to the wrong installer option. I thought it was related to the setting for using the Windows terminal instead of the MinTTY terminal!

I think the reason I attributed this wrongly is that when I first used Git during Windows 8.1 days, the default was still a rasterized font. I guess for some reason I assumed that not ticking the box in the Git installer would not do nothing, but instead enable the rasterized font. So I started ticking that box by default, and didn't give it any further thought soon enough.

So how did I miss what was the right setting that modified those keys when making this pull request? Put plainly, I overlooked the `Components: consolefont` parameter. This is also why at first I didn't remove this feature from the rest of the file, and only did so when @rimrul called it to my attention, and also why I used such a non-descriptive commit message for that: The removal of that feature per se was not the actual rationale behind my changes.

I guess there are several lessons for me here. I was too eager to fix my problem and wasn't as diligent as I should have been. I thought I should give you this explanation rather than just walk away, leaving you to wonder what the heck happened. Perhaps with my explanation my initial description also makes more sense.

In any case, I think one could make a case for what is even the point of this installer option, seeing how TrueType fonts have been the default on Windows since Windows 10 1507. Insofar, you could probably make a case for its removal. However, this case is obviously diminished by the fact that the option actually does exactly as it says - as long as you do read it correctly, of course. 😅 So if you feel it's appropriate, you can go ahead and close this issue.

dscho:
Thank you for the explanation, and: no apology needed at all!

And your analysis strikes me as correct: at this time and age, when even Windows 7 is past its end of life, we should finally be able to remove those console options.

However, I am slightly uncomfortable making that call, in case anybody can think of a compelling reason why we should still keep them (maybe they know of enough users who simply cannot move away from Windows 7 yet?). So let's let this simmer for a while before deciding?

rimrul:
Sinve we still support various old versions of windows, how about just not showing this option on Windows 10 >=1507 (and probably the corresponding server versions)?

dscho:
> 
> 
> Since we still support various old versions of windows, how about just not showing this option on Windows 10 >=1507 (and probably the corresponding server versions)?

Can we do that?

*clicketyclick*

I guess we [can](https://jrsoftware.org/ishelp/index.php?topic=commonparams#OnlyBelowVersion): just add an appropriate `OnlyBelowVersion` to the components in question...

adrianghc:
> > Since we still support various old versions of windows, how about just not showing this option on Windows 10 >=1507 (and probably the corresponding server versions)?
> 
> Can we do that?
> 
> _clicketyclick_
> 
> I guess we [can](https://jrsoftware.org/ishelp/index.php?topic=commonparams#OnlyBelowVersion): just add an appropriate `OnlyBelowVersion` to the components in question...

That looks very good. If you give me the okay, I will add this and modify the commit accordingly (and use a better commit message, too 😅).

dscho:
I'll gladly give you my okay! :-)

adrianghc:
Apologies for the delay. I've updated the PR, as far as I can tell, this should be all there is to it, shouldn't it?

dscho (approved):
I like it!

rimrul (approved):
LGTM.

dscho:
Thank you, @adrianghc!

PR #336, opened 2021-04-07
installer: ignore when the previous Git installation was deleted

dscho:
It was reported in https://github.com/git-for-windows/git/issues/3161 that the Git installer will throw errors when the previous Git installation was deleted manually.

Let's work around this by simply pretending that the system config was empty.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -740,8 +740,13 @@ begin
 
     TmpFile:=ExpandConstant('{tmp}\git-config-get.txt');
     if not Exec(ExpandConstant('{cmd}'),'/D /C .\{#MINGW_BITNESS}\bin\git.exe config -l -z '+ExtraOptions+' >'+#34+TmpFile+#34,
-                AppDir,SW_HIDE,ewWaitUntilTerminated,i) then
-        LogError('Unable to get system config');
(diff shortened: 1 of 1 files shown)


installer/install.iss
```diff
@@ -740,8 +740,13 @@ begin
 
     TmpFile:=ExpandConstant('{tmp}\git-config-get.txt');
     if not Exec(ExpandConstant('{cmd}'),'/D /C .\{#MINGW_BITNESS}\bin\git.exe config -l -z '+ExtraOptions+' >'+#34+TmpFile+#34,
-                AppDir,SW_HIDE,ewWaitUntilTerminated,i) then
-        LogError('Unable to get system config');
+                AppDir,SW_HIDE,ewWaitUntilTerminated,i) then begin
+        if FileExists(AppDir+ExpandConstant('\{#MINGW_BITNESS}\bin\git.exe')) then
```
rimrul, line 744: It feels a little odd, that we try to execute `{#MINGW_BITNESS}\bin\git.exe` and check wether it exists afterwards, but it should work fine.

rimrul (approved):

installer/install.iss
dscho: Right. I did want to favor the common code path, though, and that would be: the call succeeds.

PR #338, opened 2021-04-28
git update-git-for-windows: optionally update from a fork

dscho:
For the upcoming ARM64 support, as well as for the Microsoft fork of Git for Windows, it would be helpful if there was a way to update from the GitHub Releases of the respective fork instead of Git for Windows' own fork. This PR makes that easy.T o ask the updater to look for new versions at, say, https://github.com/dennisameling/git/releases, run the following command
as administrator:

	git config \
	   -f "$(git --exec-path)/git-update-git-for-windows.config" \
	   update.fromFork dennisameling/git

Cc: @dennisameling

### DIFF
--- a/git-extra/git-update-git-for-windows
+++ b/git-extra/git-update-git-for-windows
@@ -174,7 +174,26 @@ update_git_for_windows () {
 		;;
 	esac
 
-	latest_tag_url=https://gitforwindows.org/latest-tag.txt
+	test -f "$0.config" &&
+	fork="$(git config -f "$0.config" update.fromFork)" &&
+	test -n "$fork" ||
+	fork=
+
+	if test -n "$fork"
+	then
+		git_label="$(git config -f "$0.config" update.gitLabel)"
+		test -n "$git_label" || git_label=Git
+
+		releases_url=https://api.github.com/repos/$fork/releases
+		latest_tag_url=$releases_url/latest
+		latest_eval='latest=${latest_tag#*\"tag_name\": \"}; latest=${latest%%\"*}'
+	else
+		git_label="Git for Windows"
+		releases_url=https://api.github.com/repos/git-for-windows/git/releases
+		latest_tag_url=https://gitforwindows.org/latest-tag.txt
+		latest_eval='latest=${latest_tag#v}'
+	fi
+
 	latest_tag=$(http_get $latest_tag_url) ||
 	case $?,"$proxy" in
 	7,)
@@ -190,13 +209,13 @@ update_git_for_windows () {
 		;;
 	esac
 
-	latest=${latest_tag#v}
+	eval "$latest_eval"
 	# Did we ask about this version already?
 	recently_seen="$(get_recently_seen)"
 	test -n "$quiet" && test "x$recently_seen" = "x$latest" && return
 
 	version=$(git --version | sed "s/git version //")
-	echo "Git for Windows $version (${bit}bit)" >&2
+	echo "$git_label $version (${bit}bit)" >&2
 	if test -z "$testing" && test "$latest" = "$version"
 	then
 		echo "Up to date" >&2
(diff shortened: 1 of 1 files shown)


dscho:
I should add that I did test this and it works as advertised.

rimrul:
This'll work for microsoft/git, but don't we need a way to specify the installer suffix for ARM64?

dscho:
> 
> 
> This'll work for microsoft/git, but don't we need a way to specify the installer suffix for ARM64?

Good point...

I think we will have to extend the `.config` support to allow specifying installer file names or some such. But I think we can safely do that on top of this PR, when @dennisameling has time (they indicated to be busy elsewhere until May 8th).

git-extra/git-update-git-for-windows
```diff
@@ -174,7 +174,22 @@ update_git_for_windows () {
 		;;
 	esac
 
-	latest_tag_url=https://gitforwindows.org/latest-tag.txt
+	if test -f "$0.config"
+	then
+		fork="$(git config -f "$0.config" update.fromFork)"
+		test -n "$fork" || {
+			echo "Could not find update.fromFork in $0.config" >&2
```
rimrul, line 181: I get that end users shouldn't edit this config, but maybe we should fall back to the same behaviour we use when $0.config doesn't exist.

Wether the file exists or not, it's basically an unset config option.

dennisameling:
> But I think we can safely do that on top of this PR, when @dennisameling has time (they indicated to be busy elsewhere until May 8th).

I'm back! 😊 Created a PR on top of this one to detect ARM64: https://github.com/git-for-windows/build-extra/pull/340 - it simply checks if the `/arm64/bin` folder is present, just like we do in [git-wrapper.c](https://github.com/git-for-windows/MINGW-packages/blob/a723a9beca720c36378bce90e2085e0828d90012/mingw-w64-git/git-wrapper.c#L139-L144). I _think_ that should be okay, but you might be aware of some edge-cases where this won't work.

dennisameling (approved):
Can confirm that this works as described 👍🏼

git-extra/git-update-git-for-windows
dscho: True. I now treat a `.config` file without `update.fromFork` the same as if the `.config` file was absent.

dscho:
> 
> 
> > But I think we can safely do that on top of this PR, when @dennisameling has time (they indicated to be busy elsewhere until May 8th).
> 
> I'm back! 😊 Created a PR on top of this one to detect ARM64: #340 - it simply checks if the `/arm64/bin` folder is present, just like we do in [git-wrapper.c](https://github.com/git-for-windows/MINGW-packages/blob/a723a9beca720c36378bce90e2085e0828d90012/mingw-w64-git/git-wrapper.c#L139-L144). I _think_ that should be okay, but you might be aware of some edge-cases where this won't work.

We do a little bit more in the Git wrapper: we also confirm that we're running on ARM64.

Therefore, it would be _technically_ "more correct" if we introduced, say, `--is-running-on-ARM64` in `/cmd/git.exe` and used that.

But I think that is overkill, so let's go with your solution for now.

> 
> 
> Can confirm that this works as described 👍🏼

Could I ask you to test again? I changed the patch slightly in response to https://github.com/git-for-windows/build-extra/pull/338#pullrequestreview-648091694.

dennisameling:
@dscho Sorry for the late reply, was planning to test earlier, but was firefighting at work in the last few days 😅 

Config:
```
[update]
	fromFork = dennisameling/git
	gitLabel = "Dennis's fork"
```

Output:
```
cmd/git.exe update-git-for-windows --gui
Dennis's fork 2.31.1.windows.1 (64bit)
Update 2.31.2.windows.1 is available
```

Toast also looking good:

image

Thanks! 🚀

dscho:
No worries, I was quite distracted, too. Thanks for testing!

PR #339, opened 2021-05-02
installer: provide Windows Terminal Fragment Extension

Okeanos:
Provides initial support for installing basic Git Bash fragments for Windows Terminal (see git-for-windows/git/issues/3183 for details).
Known Issues:
- Final location of git-bash-global.json and git-bash-local.json in installer repository TBD
- Mix and match of globally installed Terminal and locally installed Git (Bash) and vice versa does not work
- Custom Git installation directories are not supported in the fragments
- Should probably include a pre-generated UUIDv5 according to the [Terminal docs](https://docs.microsoft.com/en-us/windows/terminal/json-fragment-extensions#how-to-determine-the-guid-of-an-existing-profile) for later modification

I am raising this WIP pull request to initiate the discussion on how to fix the aforementioned issues.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -113,6 +113,7 @@ Name: assoc; Description: Associate .git* configuration files with the default t
 Name: assoc_sh; Description: Associate .sh files to be run with Bash; Types: default
 Name: consolefont; Description: Use a TrueType font in all console windows; OnlyBelowVersion: 10.0
 Name: autoupdate; Description: Check daily for Git for Windows updates
+Name: windowsterminal; Description: Add a Git Bash Profile to Windows Terminal; MinVersion: 10.0.18362
 
 [Run]
 Filename: {app}\git-bash.exe; Parameters: --cd-to-home; Description: Launch Git Bash; Flags: nowait postinstall skipifsilent runasoriginaluser unchecked
@@ -133,6 +134,8 @@ Name: "{app}\dev"
 Name: "{app}\dev\mqueue"
 Name: "{app}\dev\shm"
 Name: "{app}\tmp"
+Name: "{commonappdata}\Microsoft\Windows Terminal\Fragments\Git"; Components: windowsterminal; Check: IsAdminLoggedOn
+Name: "{localappdata}\Microsoft\Windows Terminal\Fragments\Git"; Components: windowsterminal; Check: not IsAdminLoggedOn
 
 [Icons]
 Name: {group}\Git GUI; Filename: {app}\cmd\git-gui.exe; Parameters: ""; WorkingDir: %HOMEDRIVE%%HOMEPATH%; IconFilename: {app}\{#MINGW_BITNESS}\share\git\git-for-windows.ico
@@ -273,6 +276,10 @@ Type: dirifempty; Name: {app}
 
 ; Delete Git Bash options
 Type: files; Name: {app}\etc\git-bash.config
+
+; Delete Windows Terminal profile fragments
+Type: files; Name: {commonappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json
+Type: files; Name: {localappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json
 
 [Code]
 #include "helpers.inc.iss"
@@ -2692,6 +2699,32 @@ begin
         LogError(ExpandConstant('Line {#__LINE__}: Unable to remove the Git for Windows updater (output: '+ReadFileAsString(LogPath)+', errors: '+ReadFileAsString(ErrPath)+').'));
 end;
 
+procedure InstallWindowsTerminalFragment;
+var
+    Res:Longint;
+    AppPath,JSONPath:String;
+begin
+    if IsAdminInstallMode() then
+        JSONPath:=ExpandConstant('{commonappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json')
+    else
+        JSONPath:=ExpandConstant('{localappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json');
+    AppPath:=ExpandConstant('{app}');
+    StringChangeEx(AppPath, '\', '/', True)
+    if not SaveStringToFile(JSONPath,
+        '{'+
+        '    "profiles": ['+
+        '      {'+
+        '        "name": "Git Bash",'+
+        '        "commandline": "'+AppPath+'/usr/bin/bash.exe -i -l",'+
+        '        "icon": "'+AppPath+'/{#MINGW_BITNESS}/share/git/git-for-windows.ico",'+
+        '        "startingDirectory": "%USERPROFILE%"'+
+        '      }'+
+        '    ]'+
+        '  }',False) then begin
+        LogError('Line {#__LINE__}: Unable to install Windows Terminal Fragment to '+JSONPath)
+    end;
+end;
+
 {
     Create Cygwin's idea of a symbolic link:
     - a system file
@@ -3142,6 +3175,13 @@ begin
         if not DeleteFile(AppDir+'\{#MINGW_BITNESS}\bin\git-lfs.exe') and not DeleteFile(AppDir+'\{#MINGW_BITNESS}\libexec\git-core\git-lfs.exe') then
             LogError('Line {#__LINE__}: Unable to delete "git-lfs.exe".');
     end;
+
+    {
+        Create the Windows Terminal integration
+    }
+
+    if IsComponentSelected('windowsterminal') then
+        InstallWindowsTerminalFragment();
 
     {
         Set the default Git editor


rimrul (dismissed):
> Final location of git-bash-global.json and git-bash-local.json in installer repository TBD

I think they should go into `git-extra`, but I'm not completely sure.
>Mix and match of globally installed Terminal and locally installed Git (Bash) and vice versa does not work

I'm don't think we actually need to care how Windows terminal was installed.
> Custom Git installation directories are not supported in the fragments

We might be able to disable the component for those unsupported cases with a check wether the value of `{app}` is what we expect.
> Should probably include a pre-generated UUIDv5 according to the Terminal docs for later modification

Unless I'm missreading the documentation, that's only for modifying profiles that Windows Terminal comes with.

>The only profiles that can be modified through fragments are the default profiles, Command Prompt and PowerShell, as well as dynamic profiles.

and "dynamic profiles" refers to these:

>Windows Terminal will automatically create Windows Subsystem for Linux (WSL) and PowerShell profiles for you if you have these shells installed on your machine.

So I think the way we're supposed to update our profile is by updating the `json` file.

installer/git-bash-local.json
```diff
@@ -0,0 +1,10 @@
+{
+    "profiles": [
+      {
+        "name": "Git Bash",
+        "commandline": "%LOCALAPPDATA%/git/usr/bin/bash.exe -i -l",
+        "icon": "%LOCALAPPDATA%/Git/mingw64/share/git/git-for-windows.ico",
+        "startingDirectory": "%USERPROFILE%"
+      }
+    ]
+  }
```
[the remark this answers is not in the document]
Okeanos: I actually considered that based on its other contents such as the `vimrc` but was not sure on whether or not some other part of the installer includes `git-extra` by default.

installer/install.iss
```diff
@@ -124,6 +127,8 @@ Filename: {app}\ReleaseNotes.html; Description: View Release Notes; Flags: shell
 Source: {#SourcePath}\ReleaseNotes.html; DestDir: {app}; Flags: replacesameversion; AfterInstall: DeleteFromVirtualStore
 Source: {#SourcePath}\..\LICENSE.txt; DestDir: {app}; Flags: replacesameversion; AfterInstall: DeleteFromVirtualStore
 Source: {#SourcePath}\NOTICE.txt; DestDir: {app}; Flags: replacesameversion; AfterInstall: DeleteFromVirtualStore; Check: ParamIsSet('VSNOTICE')
+Source: {#SourcePath}\git-bash-global.json; DestDir: {commonappdata}\Microsoft\Windows Terminal\Fragments\Git; DestName: bash.json; Components: windowsterminal\global; Check: IsAdminLoggedOn
```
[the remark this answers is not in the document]
Okeanos: I did build the installer locally and used it to successfully install the profile as described in the commit. I was not entirely sure based on the documentation provided by them, though:

> […] In this case, the JSON files should be added to the folder:
> `C:\ProgramData\Microsoft\Windows Terminal\Fragments\{app-name}`

I took that to mean a specific `app-name` named subfolder and it worked. I am personally fine with either, though.

installer/install.iss
```diff
@@ -113,6 +113,9 @@ Name: assoc; Description: Associate .git* configuration files with the default t
 Name: assoc_sh; Description: Associate .sh files to be run with Bash; Types: default
 Name: consolefont; Description: Use a TrueType font in all console windows; OnlyBelowVersion: 10.0
 Name: autoupdate; Description: Check daily for Git for Windows updates
+Name: windowsterminal; Description: Add a Git Bash Profile to Windows Terminal; MinVersion: 10.0.18362
+Name: windowsterminal\global; Description: Windows Terminal was installed for all users of this PC; MinVersion: 10.0.18362
+Name: windowsterminal\local; Description: Windows Terminal was installed for this user only; MinVersion: 10.0.18362
```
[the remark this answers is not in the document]
Okeanos: Fair point, I didn't check/didn't know that. I'll update this along side the checks for non admin users accordingly.

installer/install.iss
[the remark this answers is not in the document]
rimrul: Right. I've re-read the code, They iterate over the folders in `CascadiaSettings::_ApplyJsonStubsHelper`, and then call `CascadiaSettings::_AccumulateJsonFilesInDirectory` and iterate over the files within. so your approach here is correct.

Okeanos:
General side question: I presume the merge request should be a single commit in the end, i.e. all fixes for the issues squashed into a single commit with a proper description and signed-off-by line?

rimrul:
We might also be able to avoid the install directory issues if we dynamically generate the `json` file like we do with the `xml` file for the auto updater in [InstallAutoUpdater](https://github.com/git-for-windows/build-extra/blob/main/installer/install.iss#L2642-L2682). So something like 

```pascal
procedure InstallWindowsTerminalFragment;
var
    Res:Longint;
    AppPath,XMLPath:String;
begin
    if IsAdminInstallMode() then
        JSONPath:=ExpandConstant('{commonappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json')
    else
        JSONPath:=ExpandConstant('{localappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json');
    AppPath:=ExpandConstant('{app}');
    SaveStringToFile(JSONPath,
        '{'+
        '    "profiles": ['+
        '      {'+
        '        "name": "Git Bash",'+
        '        "commandline": "'+AppPath+'/usr/bin/bash.exe -i -l",'+
        '        "icon": "'+AppPath+'/mingw64/share/git/git-for-windows.ico",'+
        '        "startingDirectory": "%USERPROFILE%"'+
        '      }'+
        '    ]'+
        '  }',False);
end;
```

might work. We might also need to change the `mingw64` part for 32 bit installs.

rimrul:
> I presume the merge request should be a single commit in the end, i.e. all fixes for the issues squashed into a single commit with a proper description and signed-off-by line?

Yes, exactly. just force push your updated commit and this PR will be updated accordingly.

Okeanos:
> We might also be able to avoid the install directory issues if we dynamically generate the `json` file like we do with the `xml` file for the auto updater in [InstallAutoUpdater](https://github.com/git-for-windows/build-extra/blob/main/installer/install.iss#L2642-L2682). So something like
> 
> ```pascal
> procedure InstallWindowsTerminalFragment;
> var
>     Res:Longint;
>     AppPath,XMLPath:String;
> begin
>     if IsAdminInstallMode() then
>         JSONPath:=ExpandConstant('{commonappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json')
>     else
>         JSONPath:=ExpandConstant('{localappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json');
>     AppPath:=ExpandConstant('{app}');
>     SaveStringToFile(JSONPath,
>         '{'+
>         '    "profiles": ['+
>         '      {'+
>         '        "name": "Git Bash",'+
>         '        "commandline": "'+AppPath+'\usr\bin\bash.exe -i -l",'+
>         '        "icon": "'+AppPath+'\{#MINGW_BITNESS}\share\git\git-for-windows.ico",'+
>         '        "startingDirectory": "%USERPROFILE%"'+
>         '      }'+
>         '    ]'+
>         '  }',False);
> end;
> ```
> 
> might work.

The good news: it works. The bad news: `AppPath` expands using `\` as separator which breaks JSON compatibility and renders the fragment unusable by Terminal.

rimrul:
Right. This should work around that:

```pascal
procedure InstallWindowsTerminalFragment;
var
    Res:Longint;
    AppPath,XMLPath:String;
begin
    if IsAdminInstallMode() then
        JSONPath:=ExpandConstant('{commonappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json')
    else
        JSONPath:=ExpandConstant('{localappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json');
    AppPath:=StringChangeEx(ExpandConstant('{app}'), '\', '/', True);
    SaveStringToFile(JSONPath,
        '{'+
        '    "profiles": ['+
        '      {'+
        '        "name": "Git Bash",'+
        '        "commandline": "'+AppPath+'/usr/bin/bash.exe -i -l",'+
        '        "icon": "'+AppPath+'/{#MINGW_BITNESS}/share/git/git-for-windows.ico",'+
        '        "startingDirectory": "%USERPROFILE%"'+
        '      }'+
        '    ]'+
        '  }',False);
end;
```

Okeanos:
So, I just pushed the changes you suggested:
- [x] Single option install
- [x] Additional `Checks`
- [x] Custom Procedure with embedded JSON

I tested this with local admin user by installing Git with the local installer (from scratch). Verified installation by opening Terminal and starting the newly registered Git Bash. To finish I also verified the fragment was deleted from disk when Git was uninstalled.

I could set up a local user (non-admin) in my VM as well if desired to test that part of the installation but not today anymore.

rimrul:
> I could set up a local user (non-admin) in my VM as well if desired to test that part of the installation but not today anymore.

That would be great. Thank you.

Okeanos:
I updated the procedure a little to log errors in case the file could not be saved. That aside the tests (clean installs for admin and non-admin) users as described before worked perfectly fine and the Fragment was successfully detected by Windows Terminal.

dscho (approved):
Most excellent!

installer/install.iss
```diff
@@ -2692,6 +2699,32 @@ begin
...
+    else
+        JSONPath:=ExpandConstant('{localappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json');
+    AppPath:=ExpandConstant('{app}');
+    StringChangeEx(AppPath, '\', '/', True)
+    if not SaveStringToFile(JSONPath,
+        '{'+
+        '    "profiles": ['+
+        '      {'+
+        '        "name": "Git Bash",'+
+        '        "commandline": "'+AppPath+'/usr/bin/bash.exe -i -l",'+
```
TBBle, line 2718: I'm late to the party on this, but should this be `"'+AppPath+'/bin/bash.exe -i -l"`? That's what https://github.com/microsoft/terminal/issues/1394#issuecomment-740300541 says, referencing https://stackoverflow.com/questions/56839307/adding-git-bash-to-the-new-windows-terminal#comment110596050_57369284

dscho: Excellent point. Would you kindly open a PR (I can't right now, am on a phone).

TBBle: PR created: #348

dscho: And merged!

jeremyd2019:
> > Should probably include a pre-generated UUIDv5 according to the Terminal docs for later modification
> 
> Unless I'm missreading the documentation, that's only for modifying profiles that Windows Terminal comes with.
> 

I *think* providing a pre-generated UUIDv5 would allow the user to modify/override/disable the profile shipped here in their own settings json.  That was my reading of the docs anyway

DHowett:
You're correct! Providing a stable guid of any sort will allow the user to customize git-for-windows _once_ and have their customizations stick around even if you change the name of the profile. It will also make sure that you collide with yourself (which may or may not be good, depending on your view.)

jeremyd2019:
So in this case should it be in the "guid" property like in the normal settings, or is the uuid supposed to be called "updates" in the fragment extensions schema?

315mohsen:
Mohsen2021

dscho:
So how would such a GUID be provided? Could you open a PR to do that?

jeremyd2019:
From https://github.com/microsoft/terminal/issues/10374#issuecomment-857830194

> **UGH**. Our documentation here is not correct¹.
> 
> The namespace GUID is actually based on the name of the folder/fragment name you're using, and it will be used to automatically generate a GUID if you haven't specified one.
> 
> Okay -- we can't change this because now we'll break compatibility with all user customizations on top of Fragments (grrr).
> 
> Here is the code that will generate the same GUID as Git Bash is already getting. You'll want to use this guid specifically, because doing anything else will separate the user's settings from your upstream document.
> 
> ```
> import uuid
> 
> # CALCULATE NAMESPACE BASED ON FRAGMENT FOLDER NAME
> u = uuid.uuid5(uuid.UUID("{f65ddb7e-706b-4499-8a50-40313caf510a}"), "Git".encode("UTF-16LE").decode("ASCII"))
> 
> # CALCULATE FINAL BASED ON NAMESPACE
> uuid.uuid5(u, "Git Bash".encode("UTF-16LE").decode("ASCII"))
> 
> # For Git/Git Bash: UUID('2ece5bfe-50ed-5f3a-ab87-5cd4baafed2b')
> ```
> 
> ¹ The namespace GUID it provides is actually for the internal "Windows.Terminal.Wsl" namespace... which due to legacy reasons doesn't follow the same rules. It was authored in v0.5, and we didn't want to break user customizations back then either. 😬


So we at least now know what the GUID should be.  I at least am not sure what "property" it should be on, "guid" or "updates", for a fragment.

DHowett:
> So we at least now know what the GUID should be. I at least am not sure what "property" it should be on, "guid" or "updates", for a fragment.

I'll also make sure the docs clear this up :smile:

When the fragment is creating a totally new profile, it should have a `guid` or a `name` (the `guid` will be automatically generated based on the namespace rules above if `guid` is not provided.). Providing a `guid` is a way to guarantee that no matter what you do -- whether you move the fragment to another place or rename it -- it will be stable with regards to the user's customizations.

When the fragment is modifying an existing Terminal profile that came from another source (came with Terminal, detected via the Windows Subsystem for Linux detector, etc.) that was **not** the user¹, it should contain an `updates` guid. _If the profile it is "updating" does not exist, the fragment will be ignored._

¹ This is so that a fragment cannot manipulate the user's personal profiles, no matter how hard it tries.

DHowett:
In a diagram (which I'll formalize), it's like this:

```
-
|+ 0 - THE BEGINNING OF TIME
|
|+ 1 - The oceans form
|
|+ 2 - Built-in Terminal profiles (PowerShell, Command Prompt) are loaded
|+ 3 - Dynamic Terminal profiles (PowerShell 6+, WSL) are loaded
|
|+ 4 - Fragments are loaded
|      Fragments with a `guid` become full-fledged profiles
|      Fragments with an `updates` are applied to profiles in [1], [2], [3]
|         If they do not match, they are deleted.
|
|+ 5 - User profiles are loaded
|      Profiles that have the same GUID as ones in 1,2,3,4 are overlaid on
|      top as user-specific customizations.
|      If the user has a profile that came from [3] or [4] that can no longer
|      be found, it is deleted (So, if they uninstall Git for Windows the
|      profile doesn't linger)
v
```

installer/install.iss
```diff
@@ -2692,6 +2699,32 @@ begin
...
+    Res:Longint;
+    AppPath,JSONPath:String;
+begin
+    if IsAdminInstallMode() then
+        JSONPath:=ExpandConstant('{commonappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json')
+    else
+        JSONPath:=ExpandConstant('{localappdata}\Microsoft\Windows Terminal\Fragments\Git\git-bash.json');
+    AppPath:=ExpandConstant('{app}');
+    StringChangeEx(AppPath, '\', '/', True)
+    if not SaveStringToFile(JSONPath,
```
Nirmal4G, line 2713: If the `{AppData}\Microsoft\Windows Terminal\Fragments\Git` path doesn't exist, then the json file won't be written. Better to make sure that the path exists and make one if not.

Testing the latest released version as of now, the installer shows the following error…

```log
Unable to install Windows Terminal Fragment to C:\ProgramData\Microsoft\Windows Terminal\Fragments\Git\git-bash.json
```

Nirmal4G: @dscho

rimrul: We do create the directory before writing a file to it, but there was an oversight with upgrade installations. See git-for-windows/git#3312 for details and #363 for the proposed fix.

PR #340, opened 2021-05-08
git update-git-for-windows: download ARM64 Git for Windows if applicable

dennisameling:
Builds further on #338 
Needed for https://github.com/git-for-windows/git/issues/3107

If the `/arm64/bin` folder is present, we assume that the user has the ARM64 version of Git for Windows installed. When a new version is available, we'll then automatically download the ARM64 version. This is similar to the logic in [git-wrapper.c](https://github.com/git-for-windows/MINGW-packages/blob/a723a9beca720c36378bce90e2085e0828d90012/mingw-w64-git/git-wrapper.c#L139-L144) in mingw-w64-git.

Sample release with a fake ARM64 installer (renamed 64-bit installer, just to show that the installer indeed downloads the correct file): https://github.com/dennisameling/git/releases/tag/2.31.2.windows.1

Tested as follows:
1. Ensure `/arm64/bin` folder is present
2. Point the update check script to my fake 2.31.2 release: https://github.com/dennisameling/git/releases/tag/2.31.2.windows.1
3. Run `cmd\git.exe update-git-for-windows --gui` from Git Bash
4. Get the "Update available" notification - it correctly downloaded the ARM64 installer

### DIFF
--- a/git-extra/git-update-git-for-windows
+++ b/git-extra/git-update-git-for-windows
@@ -215,7 +215,14 @@ update_git_for_windows () {
 	test -n "$quiet" && test "x$recently_seen" = "x$latest" && return
 
 	version=$(git --version | sed "s/git version //")
-	echo "$git_label $version (${bit}bit)" >&2
+	if test -d /arm64/bin
+	then
+		arch_bit=ARM64
+	else
+		arch_bit=${bit}-bit
+	fi
+
+	echo "$git_label $version ($arch_bit)" >&2
 	if test -z "$testing" && test "$latest" = "$version"
 	then
 		echo "Up to date" >&2
@@ -237,7 +244,7 @@ update_git_for_windows () {
 	releases=$(http_get $releases_url/latest) || return
 	download=$(echo "$releases" |
 		grep '"browser_download_url": "' |
-		grep "$bit\-bit\.exe" |
+		grep "$arch_bit\.exe" |
 		sed -E 's/.*": "([^"]*).*/\1/')
 	filename=$(echo "$download" | sed -E 's/.*\/([^\/]*)$/\1/')
 	name="$(echo "$releases" | sed -n 's/^  "name": "\(.*\)",$/\1/p')"


git-extra/git-update-git-for-windows
```diff
@@ -190,13 +205,22 @@ update_git_for_windows () {
...
 
 	version=$(git --version | sed "s/git version //")
-	echo "Git for Windows $version (${bit}bit)" >&2
+	version_string="Git for Windows $version (${bit}bit)"
+	filename_for_download="$bit\-bit\.exe"
+
+	if [ -d "/arm64/bin" ]
+	then
+		version_string="Git for Windows $version (ARM64)"
+		filename_for_download="ARM64\.exe"
```
dennisameling, line 220: I assumed that you want to call the ARM64 binaries something like `Git-2.31.2-ARM64.exe` - happy to change if you think there might be a better name though!

git-extra/git-update-git-for-windows
```diff
@@ -211,7 +211,16 @@ update_git_for_windows () {
 	test -n "$quiet" && test "x$recently_seen" = "x$latest" && return
 
 	version=$(git --version | sed "s/git version //")
-	echo "Git for Windows $version (${bit}bit)" >&2
+	version_string="Git for Windows $version (${bit}bit)"
```
dscho, line 214: This should probably be replaced already, too. How about

```sh
if test -d /arm64/bin
then
    arch_bit=ARM64
else
    arch_bit=${bit}-bit
fi

echo "Git for Windows $version ($arch_bit)" >&2

[...]
        grep "$arch_bit\.exe" |
[...]
```

git-extra/git-update-git-for-windows
```diff
@@ -211,7 +211,16 @@ update_git_for_windows () {
...
+	filename_for_download="$bit\-bit\.exe"
+
+	if [ -d "/arm64/bin" ]
```
dscho, line 217: Technically, the Git wrapper does more: it first verifies that we _are_ running on ARM64. Therefore, we could implement the `--is-running-on-ARM64` option in `/cmd/git.exe` and use that. But that sounds like quite a bit of effort for almost no value in return, so I'm fine with the directory check.

To align with Git's coding style, I would like to use `test` instead of `[`, though.

git-extra/git-update-git-for-windows
dscho: No, I concur with you: `Git-2.31.2-ARM64.exe` sounds good.

dscho (changes requested):
I'd like to ask for just minor changes, and for a rebase on top of my (updated) PR.

dscho (approved):

dscho:
I went ahead and made those changes, and merged it. If things do not quite work as expected, we'll just need another PR, eh? 😄

dennisameling:
Thanks for making those changes! Quick test is looking good:

```
$ cmd/git.exe update-git-for-windows --gui
Dennis's fork 2.31.1.windows.1 (ARM64)
Update 2.31.2.windows.1 is available
```

Indeed downloads from the right path: https://github.com/dennisameling/git/releases/download/2.31.2.windows.1/Git-2.31.2-ARM64.exe

Great! This is really exciting 😊

PR #349, opened 2021-05-25
installer: make the Windows Terminal profile option more visible

dscho:
When upgrading from prior versions, the Components page was not shown by default (because the installer thinks that there were no options that had not been shown during earlier installations/upgrades).

Let's fix this.

Cc: @Okeanos


dscho:
@Okeanos did you try it?

Okeanos:
Didn't have time yet, I'll try to squeeze that in today and  give you feedback.

Okeanos:
Okay, from a functionality perspective this appears to work as desired. I used a current release to cleanly install Git then build an installer with your changes and could successfully select & install the fragment.

That was something I didn't actually check during the original merge request (I think I just assumed it'd work out of the box because there's this "show only new features" checkbox in the process) – thanks a lot for fixing that :)

PR #350, opened 2021-05-26
In preparation for supporting contrib/scalar/, teach the installer about it

dscho:
Scalar is a project that tests out new strategies to make Git scale better. Over the years, hits like "partial clone" and "commit-graph" and `git maintenance` were developed in Scalar and then upstreamed into Git.

Over in the Microsoft fork, I am busy at work re-implementing the remaining parts of [Scalar](https://github.com/microsoft/scalar#readme) in C, putting them in `contrib/scalar/`. The idea is to eventually upstream all remaining bits and pieces of Scalar that do not require the GVFS protocol (which are only required because Azure Repos do not support partial clones, at time of writing).

To give this a smooth gliding path, I want to include the `contrib/scalar/` parts in the releases of the Microsoft fork first, given them some real world testing, and then start upstreaming.

To help with this, I want to teach Git for Windows' scripts to build the installer to prepare for this by offering the C Port of Scalar as an experimental option (if the `scalar.exe` artifacts are included in the installer's payload).

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -2850,6 +2850,49 @@ begin
     end;
 end;
 
+function UpgradeFromDotNetBasedScalar:Boolean;
+var
+    RegKey,UninstallScalar,ScalarExe,Cmd:String;
+    Res:Longint;
+    Enlistments:TArrayOfString;
+    i:Integer;
+begin
+    Result:=True;
+
+    RegKey:='SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{82F731CB-1CFC-406D-8D84-8467BF6040C7}_is1';
+    if not RegQueryStringValue(HKEY_LOCAL_MACHINE,RegKey,'UninstallString',UninstallScalar) then
+        // No existing Scalar found; ignore this silently
+        Exit;
+
+    // Check twice that .NET-based Scalar is there
+    ScalarExe:=UninstallScalar;
+    StringChangeEx(ScalarExe,'"','',True); // strip the surrounding double-quote characters
+    StringChangeEx(ScalarExe,'\unins000.exe','\scalar.exe',True);
+    if not FileExists(ScalarExe) then
+        Exit;
+
+    WizardForm.StatusLabel.Caption:='Upgrading from .NET-based Scalar';
+
+    // First, get .NET-based Scalar's idea of the registered enlistments
+    if not ExecSilently('"'+ScalarExe+'" list','scalar-list',ExpandConstant('Line {#__LINE__}: Unable to run `scalar list`')) then
+        Result:=False;
+    LoadStringsFromFile(ExpandConstant('{tmp}\scalar-list.out'),Enlistments);
+
+    // Now, register them with the C-based Scalar
+    for i:=0 to Length(Enlistments)-1 do begin
+        WizardForm.StatusLabel.Caption:='Registering '+Enlistments[i]+' with Scalar';
+        ExecSilentlyAsOriginalUser('"'+AppDir+'\cmd\scalar.exe" register "'+Enlistments[i]+'"','scalar-register-'+IntToStr(i),ExpandConstant('Line {#__LINE__}: Could not register "'+Enlistments[i]+'" with Scalar'));
+    end;
+
+    // Now uninstall the .NET-based Scalar
+    // (leaving C:\ProgramData\Scalar in place, in case
+    // the user needs to downgrade again to get unblocked)
+    WizardForm.StatusLabel.Caption:='Uninstalling .NET-based Scalar';
+    Cmd:='"'+UninstallScalar+'/VERYSILENT /SILENT /NORESTART /SUPPRESSMSGBOXES /LOG"';
+    if (not Exec(ExpandConstant('{sys}\cmd.exe'),'/D /C '+Cmd,'',SW_HIDE,ewWaitUntilTerminated,Res)) or (Res<>0) then
+        LogError('Could not uninstall Scalar. Trying to continue anyway.');
+end;
+
 procedure CurStepChanged(CurStep:TSetupStep);
 var
     DllPath,FileName,Cmd,Msg,Ico:String;
@@ -3249,7 +3292,8 @@ begin
            not DeleteFile(AppDir+'\{#MINGW_BITNESS}\share\doc\git-doc\scalar.html') then begin
             LogError('Line {#__LINE__}: Unable to delete "scalar.exe".');
         end;
-    end;
+    end else
+        UpgradeFromDotNetBasedScalar();
 #endif
 
     {
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -3292,8 +3292,10 @@ begin
            not DeleteFile(AppDir+'\{#MINGW_BITNESS}\share\doc\git-doc\scalar.html') then begin
             LogError('Line {#__LINE__}: Unable to delete "scalar.exe".');
         end;
-    end else
+    end else begin
         UpgradeFromDotNetBasedScalar();
+        ExecSilentlyAsOriginalUser('"'+AppDir+'\cmd\scalar.exe" reconfigure --all','scalar-reconfigure','Line {#__LINE__}: Could not reconfigure Scalar enlistments');
+    end;
 #endif
 
     {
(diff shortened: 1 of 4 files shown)


dscho:
I still have to test the upgrade story of this.

installer/install.iss
```diff
@@ -2245,45 +2250,47 @@ begin
...
 
     // Restore the settings chosen during a previous install
     RdbExperimentalOptions[GP_EnablePCon].Checked:=ReplayChoice('Enable Pseudo Console Support','Auto')='Enabled';
 #endif
 
+#ifdef WITH_EXPERIMENTAL_SCALAR
+    RdbExperimentalOptions[GP_InstallScalar]:=CreateCheckBox(ExperimentalOptionsPage,'Install Scalar (ported to C)','<RED>(NEW!)</RED> Install Scalar (ported from <A HREF=https://github.com/microsoft/scalar#readme>the .NET-based version</A>).',TabOrder,Top,Left);
+
+    // Restore the settings chosen during a previous install
+    RdbExperimentalOptions[GP_InstallScalar].Checked:=ReplayChoice('Install Scalar','Auto')='Disabled';
```
derrickstolee, line 2291: We want this to be enabled by default, at least for `microsoft/git`, right?

installer/install.iss
```diff
@@ -2797,6 +2804,52 @@ begin
...
+        Result:=False;
+    end;
+    LoadStringsFromFile(OutPath,Enlistments);
+
+    // Now, register them with the C-based Scalar
+    for i:=0 to Length(Enlistments)-1 do begin
+        Cmd:='/D /C ""'+AppDir+'\cmd\scalar.exe" register "'+Enlistments[i]+'"  >"'+OutPath+'" 2>"'+ErrPath+'""';
+        if not Exec(ExpandConstant('{sys}\cmd.exe'),Cmd,'',SW_HIDE,ewWaitUntilTerminated,Res) or (Res<>0) then
+            LogError('Could not register "'+Enlistments[i]+'" with Scalar (code: '+IntToStr(Res)+' output: '+ReadFileAsString(OutPath)+', errors: '+ReadFileAsString(ErrPath)+'); ignoring');
+    end;
```
derrickstolee, line 2843: Perhaps this should just be a "add these paths to the global `scalar.repo` config list" and then we want to run `scalar register` on each of those paths during every upgrade, anyway, to get the latest recommended config settings.

dscho:
> 
> 
> I still have to test the upgrade story of this.

I tested this as well as I could, and it seems to work exactly as designed.

installer/install.iss
dscho: Yes, eventually. The way to do this is to pass `/o:InstallScalar=Enabled` to the installer on the command-line. Or to invert the logic where `Auto` would imply `Disabled`.

installer/install.iss
dscho: Hmm. Maybe we can add a new verb for that? Or would it be something like `git for-each-repo --config=scalar.repos scalar register` (I'd have to teach `git for-each-repo` to special-case the argument `scalar` to not mean `git-scalar` but `scalar.exe` instead)?

installer/install.iss
dscho: Or maybe you want me to flip the default to enable it already, in preparation for an experimental release including the native Scalar?

installer/install.iss
derrickstolee: Right, I think I made note of that possible change to `git for-each-repo` to allow a non-git subcommand (`--no-git`?).

installer/install.iss
derrickstolee: Yeah, I'm actually thinking we shouldn't even have the option: it's either included or not. At least, that should be the case for `microsoft/git` and I'm not sure the best course of action for `git-for-windows/git`.

installer/install.iss
dscho: We could take a page out of aliases' book and support `git for-each-repo --config=scalar.repos !scalar register`.

installer/install.iss
dscho: I could also "hide" it under Components, next to Git LFS, defaulting to "on". What do you think?

installer/install.iss
dscho: I think I'd actually be a lot in favor of implementing `scalar upgrade-config [--all | -a]`.

installer/install.iss
derrickstolee: That level of fine-grained detail might be enough for us to be confident it will be installed without alienating power users. Thanks.

installer/install.iss
derrickstolee: That works for me.

derrickstolee (approved):

installer/install.iss
dscho: Done.

dscho: Wait, no, not here.

installer/install.iss
dscho: Done.

installer/install.iss
dscho: I've decided to go with the verb name `reconfigure` (and make `config` a hard-coded alias).

derrickstolee (approved):

installer/install.iss
```diff
@@ -114,6 +114,10 @@ Name: assoc_sh; Description: Associate .sh files to be run with Bash; Types: def
 Name: consolefont; Description: Use a TrueType font in all console windows; OnlyBelowVersion: 10.0
 Name: autoupdate; Description: Check daily for Git for Windows updates
 Name: windowsterminal; Description: "(NEW!) Add a Git Bash Profile to Windows Terminal"; MinVersion: 10.0.18362
+#ifdef WITH_EXPERIMENTAL_SCALAR
+Name: scalar; Description: "(NEW!) Scalar"; Types: default
```
rimrul, line 118: This seems a little short on words. Can we give the user a bit of a hint what scalar is? A hint that this is experimental (and not the same scalar they find by googling "scalar git" (.NET scalar)) might also be helpfull.

dscho: That's a good point.

Unfortunately, I don't think there's an easy way to embed links e.g. to https://github.com/microsoft/scalar#readme.

And also: I am currently a bit busy with -rc2, and then have to fix bugs in this PR (I tested without building the full installer so far, and naturally that hid the bug where `scalar reconfigure --all` must not be run as administrator because it wouldn't find the enlistments).

After that, I will try to come up with a better description.

Please note, however, that the description is not likely to matter a whole lot, as the component will _only_ be shown if the installer does include Scalar's files. And if that is already the case, the idea is to turn this on by default. And the checkbox is there only out of an abundance of caution: I would like users not to turn it off unless they have a _really_ good reason for that (one valid reason would be a bug that forces them to go back to .NET-based Scalar).

PhilipOakley: Maybe steal the very initial "Scalar is a project that tests out new strategies to make Git scale better. " Then something about it meaning very large repositories, and maybe selecting a commonly known one that's just large enough (Chromium, Android, whatever) to count (rather than just being "Micro$oft Mega Repo" ;-)

derrickstolee: For some ideas about how to briefly describe Scalar, perhaps consider [the Philosophy of Scalar](https://github.com/microsoft/scalar/blob/main/docs/philosophy.md), specifically this sentence:

> Scalar takes out the guesswork by automatically configuring your Git repositories to take advantage of the latest and greatest features.

dscho: I don't really think that the Components page has enough space for such a description:



As you can easily see, something like "Scalar is a project that tests out new strategies to make Git scale better" would most likely be way too long, and in any case inconsistent with the existing descriptions.

dscho: How about this?

> Scalar (Git add-on to manage large-scale repositories)

rimrul: Sounds good.

dscho:
> [installer: automatically reconfigure Scalar enlistments](/git-for-windows/build-extra/pull/350/commits/3ef1b52dd6db2870d789c645bcc8c8bd5bfee566)&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; 3ef1b52

Due to this, I think I should wait to merge this PR (even if it is ready) until https://github.com/microsoft/git/pull/359 is merged.

installer/install.iss
dscho: This is how it looks now:

installer/install.iss
dscho: See https://github.com/microsoft/git/pull/359 for the implementation.

derrickstolee (approved):

PR #352, opened 2021-05-28
installer: do show unseen experimental options

dscho:
This fixes a long-standing bug that I only detected while trying to release Git for Windows v2.32.0-rc2: if there were unseen experimental options, they would not be shown by mistake, but skipped when "Only show new options" was checked.


dscho:
I just tested this, and it finally works as expected.

PR #353, opened 2021-05-28
installer: try to deal with running FSMonitor daemons gracefully

dscho:
We now have long-running Git processes, processes that would potentially block upgrades.

However, due to a mistake the installer would not show them: we only looked for processes using `/mingw64/bin/git.exe` (or the equivalent in 32-bit setups), while the FSMonitor daemon typically runs via `/mingw64/libexec/git-core/git.exe`, namely when it was spawned implicitly.

Since the installer missed these processes, they would not be killed, and InnoSetup would "helpfully" require a reboot after the upgrade.

This PR not only plugs that hole by detecting more Git processes, but also brings code to automatically shut down all FSMonitor daemons before the actual upgrade.

The method presented here might not be ideal, but is probably a bit more robust and simpler than the approach implemented inhttps://github.com/git-for-windows/git/pull/3237 (which would still have required a bit of work on the installer in order not to bother the user about needing to stop Git processes manually).

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -572,6 +572,36 @@ begin
     end;
 end;
 
+// Returns true if at least one FSMonitor daemon was shut down successfully
+function ShutdownFSMonitorDaemons():Boolean;
+var
+    FindRec:TFindRec;
+    Path:String;
+    Len:Integer;
+begin
+    Result:=False;
+#ifdef WITH_EXPERIMENTAL_BUILTIN_FSMONITOR
+    if not FindFirst('\\.\pipe\*',FindRec) then
+        Exit;
+    repeat
+        if WildcardMatch(FindRec.Name,'*\fsmonitor--daemon.ipc') or WildcardMatch(FindRec.Name,'*\.git\fsmonitor') then begin
+            // The colon was replaced with an underscore by the FSMonitor daemon
+            Len:=Length(FindRec.Name);
+            if WildcardMatch(FindRec.Name,'*\fsmonitor--daemon.ipc') then
+                Len:=Len-22
+            else
+                Len:=Len-10;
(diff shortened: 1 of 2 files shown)


installer/install.iss
```diff
@@ -578,6 +572,66 @@ begin
...
+            if (Pos(#0+FindRec.Name+#0,AlreadyHandledFSMonitorPaths)>0) then
+                Continue;
+            AlreadyHandledFSMonitorPaths:=AlreadyHandledFSMonitorPaths+FindRec.Name+#0;
+
+            // An earlier `fsmonitor--daemon` iteration called it `--stop`, not `stop`;
+            // Find out which form to use.
+            if (BuiltinFSMonitorStopOption='') then begin
+                BuiltinFSMonitorStopOption:='(huh?)';
+                Path:=ExpandConstant('{tmp}\fsmonitor--help.err');
+                if not Exec(ExpandConstant('{sys}\cmd.exe'),'/D /C ""'+AppDir+'\cmd\git.exe" fsmonitor--daemon -h 2>"'+Path+'""','',SW_HIDE,ewWaitUntilTerminated,i) or (i<>129) then begin
```
dscho, line 604: Hmm. I guess we should be quiet about this (and exit early) if

- `AppDir+'\cmd\git.exe'` does not even exist (exit code 127), or
- FSMonitor was not included in the previously-installed Git for Windows version (i.e. if `git fsmonitor--daemon -h` returns exit code 1.

My current thinking is that the error message should be suppressed if the exit code (i.e. `i`) is either 1 or 127.

dscho: Done.

dscho:
I'll just merge this now, if there are concerns, we can always iterate on top.

PR #355, opened 2021-06-04
installer: Support Sublime Text 4 and (probably) newer

rimrul:
Sublime Text has recently been released as version 4. Along with this new
version the install directory has been changed to no longer include the version
number in the directory name. Change the installer to detect newer versions of
Sublime Text and prefer them over version 3.

This fixes git-for-windows/git#3248


dscho (approved):
Nice!

PR #357, opened 2021-06-10
Warn about switching away from .vfs versions

dscho:
The most likely reasons why users have .vfs versions (i.e. Git for Windows flavors whose `git version` show something like `git version 2.31.1.vfs.0.1`) is that they have downloaded VFS for Git or Scalar and cloned from an Azure Repo.

In such a scenario, it would not be good to switch away from a .vfs version because regular Git for Windows has no support to talk the GVFS protocol (which is used by the .vfs flavor to emulate partial clones), and therefore their VFS for Git or Scalar enlistments would be broken as a consequence.

Let's try to be helpful in such circumstances and give the user a chance to stay with the working setup. This is how it will look like:

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -982,23 +982,34 @@ begin
             Result:=Result+1;
 end;
 
-function IsDowngrade(CurrentVersion:String):Boolean;
 var
-    Path,PreviousVersion:String;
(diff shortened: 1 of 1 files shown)


derrickstolee (approved):
Thank you for this! I think this message will help us a lot.

PR #358, opened 2021-06-21
Use signed version comparisons before recommending upgrade

derrickstolee:
This is not super-important for `git-for-windows/git` because it always uses `-rcX` text in its pre-releases. However, in `microsoft/git` we use regular-versioned tags as prereleases that we hand to dogfooders before promoting that same tag as a full release. This has led to users being prompted for upgrades because their version doesn't match the latest, even though their current version is newer.

Always use a version comparison that compares the components carefully. I tested the new `version_compare` very carefully, but am not sure of how to test the change in the second commit.

### DIFF
--- a/git-extra/git-update-git-for-windows
+++ b/git-extra/git-update-git-for-windows
@@ -31,21 +31,15 @@ version_compare () {
 		test $a1 -le $b1 || { echo 1; return; }
 		test $b1 -le $a1 || { echo -1; return; }
 
-		# Get the next character
-		a1="$(expr "$a" : '^\(.\)')"; a="${a#$a1}"
-		b1="$(expr "$b" : '^\(.\)')"; b="${b#$b1}"
+		# Skip non-numeric prefixes
+		a1="$(expr "$a" : '^\([^0-9]\+\)')"; a="${a#$a1}"
+		b1="$(expr "$b" : '^\([^0-9]\+\)')"; b="${b#$b1}"
 
-		test "x$a1" = "x$b1" || {
-			if test . = "$b1"
-			then
-				echo -1
-			else
-				echo 1
-			fi
-			return
-		}
-
-		test . = "$a1" || { echo 0; return; }
+		case "$a1,$b1" in
+		-rc,-rc) ;; # both are -rc versions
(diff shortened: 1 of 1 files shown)


git-extra/git-update-git-for-windows
```diff
@@ -4,51 +4,131 @@
...
+		return 1
+	fi
+	if [[ "x$b1" < "x$a1" ]]
+	then
+		echo 1
+		return 1
+	fi
+
+	return 0
+}
```
dscho: Hmm. I dislike the complexity of having to do lexicographic comparisons... Let me try whether I can come up with an alternative solution real quick.

dscho: Sorry, I meant to comment immediately but forgot to "finalize my review".

dscho:
@derrickstolee I think I came up with a simpler solution. We do not really need to take care of _all possible_ version number formats, after all, just the ones we use in Git for Windows (and in `microsoft/git`).

What do you think about my alternative solution? (I force-pushed it to your branch so that it is easier to compare the two.)

git-extra/git-update-git-for-windows
```diff
@@ -31,24 +31,43 @@ version_compare () {
...
-		test . = "$a1" || { echo 0; return; }
+		case "$a1,$b1" in
+		-rc,-rc) ;; # both are -rc versions
+		-rc,*) echo -1; return;;
+		*,-rc) echo 1; return;;
+		esac
 	done
 }
 
+test "--test-version-compare" != "$*" || {
```
derrickstolee, line 46: I like this!

derrickstolee (commented):
@dscho I like yours better! Much cleaner, thanks. I'll rebase.

derrickstolee:
> @dscho I like yours better! Much cleaner, thanks. I'll rebase.

Nevermind, you already rebased. Thanks!

PhilipOakley:
(duplicated from the merged commit)
+1
This looks good.
It moves the special .rc* case into the version_compare() function, and will almost certainly resolve my issue git-for-windows/git#1843 (comment)

PR #361, opened 2021-06-30
Reduce MinGit's size a little

dscho:
Together with https://github.com/git-for-windows/MINGW-packages/pull/52, this addresses a couple of concerns raised in https://github.com/git-for-windows/git/issues/1439#issuecomment-865501015 (and subsequent comments on that ticket).


ur4t:
There are three parts in package `diffutils`: binaries, locales and docs. While locales and docs are stripped out, we can exclude the package `diffutils` directly instead of excluding its binaries one by one.

dscho:
> 
> 
> There are three parts in package `diffutils`: binaries, locales and docs. While locales and docs are stripped out, we can exclude the package `diffutils` directly instead of excluding its binaries one by one.

Good point!

PR #367, opened 2021-07-22
Add choice for external OpenSSH

Okeanos:
This pull request addresses git-for-windows/git/issues/1683 and git-for-windows/git/issues/2944 by adding an additional detecting mechanism for Windows OpenSSH to the choice page formerly known as PuTTYPage and allows the user to "skip" the installation of the bundled SSH related binaries that Windows also supplies.

Screenshot 2021-07-22 at 17 17 29
When selected this allows interaction between for example KeePassXC OpenSSH agent integration and Git. This may partially address keepassxreboot/keepassxc/issues/4681 I suppose.

Screenshot 2021-07-22 at 17 27 25
There is a small problem with the choice page, though, at the moment because the installer window is not tall enough to fully display the choices.
I suppose rewording the texts or omitting the PuTTY/Plink option if it wasn't successfully detected would work?

Also I have no 32Bit system to test this with 🤷‍♂️ 

Window supplies the following OpenSSH related binaries as documented [here](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_overview).

Microsoft built OpenSSH independent from Windows is available [here](https://github.com/PowerShell/openssh-portable) and releases are tracked [here](https://github.com/powershell/Win32-OpenSSH).

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -469,7 +469,7 @@ var
     RdbPath:array[GP_BashOnly..GP_CmdTools] of TRadioButton;
 
     // Wizard page and variables for the SSH options.
-    PuTTYPage:TWizardPage;
+    SSHChoicePage:TWizardPage;
     RdbSSH:array[GS_OpenSSH..GS_Plink] of TRadioButton;
     EdtPlink:TEdit;
     TortoisePlink:TCheckBox;
@@ -2063,18 +2063,18 @@ begin
      *)
 
     if RegGetSubkeyNames(HKEY_CURRENT_USER,'Software\SimonTatham\PuTTY\Sessions',PuTTYSessions) and (GetArrayLength(PuTTYSessions)>0) then begin
-        PuTTYPage:=CreatePage(PrevPageID,'Choosing the SSH executable','Which Secure Shell client program would you like Git to use?',TabOrder,Top,Left);
+        SSHChoicePage:=CreatePage(PrevPageID,'Choosing the SSH executable','Which Secure Shell client program would you like Git to use?',TabOrder,Top,Left);
 
         // 1st choice
-        RdbSSH[GS_OpenSSH]:=CreateRadioButton(PuTTYPage,'Use bundled OpenSSH','This uses ssh.exe that comes with Git.',TabOrder,Top,Left);
+        RdbSSH[GS_OpenSSH]:=CreateRadioButton(SSHChoicePage,'Use bundled OpenSSH','This uses ssh.exe that comes with Git.',TabOrder,Top,Left);
 
         // 2nd choice
-        RdbSSH[GS_Plink]:=CreateRadioButton(PuTTYPage,'Use (Tortoise)Plink','To use PuTTY, specify the path to an existing copy of (Tortoise)Plink.exe:',TabOrder,Top,Left);
-        EdtPlink:=TEdit.Create(PuTTYPage);
+        RdbSSH[GS_Plink]:=CreateRadioButton(SSHChoicePage,'Use (Tortoise)Plink','To use PuTTY, specify the path to an existing copy of (Tortoise)Plink.exe:',TabOrder,Top,Left);
+        EdtPlink:=TEdit.Create(SSHChoicePage);
         EdtPlink.Left:=ScaleX(Left+24);
         EdtPlink.Top:=ScaleY(Top);
         with EdtPlink do begin
-            Parent:=PuTTYPage.Surface;
+            Parent:=SSHChoicePage.Surface;
 
             EnvSSH:=GetEnvStrings('GIT_SSH',IsAdminLoggedOn);
             if (GetArrayLength(EnvSSH)=1) and IsPlinkExecutable(EnvSSH[0]) then begin
@@ -2095,11 +2095,11 @@ begin
             TabOrder:=TabOrder;
         end;
         TabOrder:=TabOrder+1;
-        BtnPlink:=TButton.Create(PuTTYPage);
+        BtnPlink:=TButton.Create(SSHChoicePage);
         BtnPlink.Left:=ScaleX(Left+344);
         BtnPlink.Top:=ScaleY(Top);
         with BtnPlink do begin
-            Parent:=PuTTYPage.Surface;
+            Parent:=SSHChoicePage.Surface;
             Caption:='...';
             OnClick:=@BrowseForPuTTYFolder;
             Width:=ScaleX(21);
@@ -2110,12 +2110,12 @@ begin
         Top:=Top+29;
 
         // Add checkbox for tortoise plink
-        TortoisePlink:=TCheckBox.Create(PuTTYPage);
+        TortoisePlink:=TCheckBox.Create(SSHChoicePage);
         TortoisePlink.Left:=ScaleX(Left+24);
         TortoisePlink.Top:=ScaleY(Top);
         with TortoisePlink do begin
             Caption:='Set ssh.variant for Tortoise Plink';
-            Parent:=PuTTYPage.Surface;
+            Parent:=SSHChoicePage.Surface;
             Width:=ScaleX(405);
             Height:=ScaleY(17);
             TabOrder:=TabOrder;
@@ -2135,7 +2135,7 @@ begin
         if (data='true') then
             TortoisePlink.Checked:=True;
     end else begin
-        PuTTYPage:=NIL;
+        SSHChoicePage:=NIL;
     end;
 
     (*
@@ -2486,7 +2486,7 @@ begin
             Wizardform.NextButton.Enabled:=False;
             Exit;
         end;
-    end else if (PuTTYPage<>NIL) and (CurPageID=PuTTYPage.ID) then begin
+    end else if (SSHChoicePage<>NIL) and (CurPageID=SSHChoicePage.ID) then begin
         Result:=RdbSSH[GS_OpenSSH].Checked or
             (RdbSSH[GS_Plink].Checked and FileExists(EdtPlink.Text));
         if not Result then begin
@@ -3158,7 +3158,7 @@ begin
     DeleteMarkedEnvString('GIT_SSH');
     DeleteMarkedEnvString('SVN_SSH');
 
-    if (PuTTYPage<>NIL) then begin
+    if (SSHChoicePage<>NIL) then begin
         GitSystemConfigSet('ssh.variant',#0);
         if RdbSSH[GS_Plink].Checked then begin
             SetAndMarkEnvString('GIT_SSH',EdtPlink.Text,True);
@@ -3460,7 +3460,7 @@ begin
     // Git SSH options.
     Data:='';
     Data2:='false';
-    if (PuTTYPage=NIL) or RdbSSH[GS_OpenSSH].Checked then begin
+    if (SSHChoicePage=NIL) or RdbSSH[GS_OpenSSH].Checked then begin
         Data:='OpenSSH';
     end else if RdbSSH[GS_Plink].Checked then begin
         Data:='Plink';
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -359,8 +359,9 @@ const
     DB_Manual         = 2;
 
     // Git SSH options.
-    GS_OpenSSH        = 1;
-    GS_Plink          = 2;
+    GS_OpenSSH         = 1;
+    GS_Plink           = 2;
+    GS_ExternalOpenSSH = 3;
 
     // Git HTTPS (cURL) options.
     GC_OpenSSL        = 1;
@@ -470,7 +471,7 @@ var
 
     // Wizard page and variables for the SSH options.
     SSHChoicePage:TWizardPage;
-    RdbSSH:array[GS_OpenSSH..GS_Plink] of TRadioButton;
+    RdbSSH:array[GS_OpenSSH..GS_ExternalOpenSSH] of TRadioButton;
     EdtPlink:TEdit;
     TortoisePlink:TCheckBox;
 
@@ -2058,11 +2059,12 @@ begin
     end;
 
     (*
-     * Create a custom page for using (Tortoise)Plink instead of OpenSSH
-     * if at least one PuTTY session is found in the Registry.
+     * Create a custom page for using (Tortoise)Plink or a self-supplied OpenSSH instead of bundled OpenSSH
+     * if at least one PuTTY session is found in the Registry or an OpenSSH binary is found on the PATH.
      *)
 
-    if RegGetSubkeyNames(HKEY_CURRENT_USER,'Software\SimonTatham\PuTTY\Sessions',PuTTYSessions) and (GetArrayLength(PuTTYSessions)>0) then begin
+    if (RegGetSubkeyNames(HKEY_CURRENT_USER,'Software\SimonTatham\PuTTY\Sessions',PuTTYSessions) and (GetArrayLength(PuTTYSessions)>0)) or
+       (FileSearch('ssh.exe', GetEnv('PATH')) <> '') then begin
         SSHChoicePage:=CreatePage(PrevPageID,'Choosing the SSH executable','Which Secure Shell client program would you like Git to use?',TabOrder,Top,Left);
 
         // 1st choice
@@ -2123,9 +2125,16 @@ begin
         TabOrder:=TabOrder+1;
         Top:=Top+17;
 
+        // 3rd choice
+        RdbSSH[GS_ExternalOpenSSH]:=CreateRadioButton(SSHChoicePage,'Use external OpenSSH',
+            'This uses an external ssh.exe. Git will not install its own OpenSSH (related)'+#13+
+            'binaries but use them as found on the PATH.',
+            TabOrder,Top,Left);
+
         // Restore the setting chosen during a previous install.
         case ReplayChoice('SSH Option','OpenSSH') of
             'OpenSSH': RdbSSH[GS_OpenSSH].Checked:=True;
+            'ExternalOpenSSH': RdbSSH[GS_ExternalOpenSSH].Checked:=True;
             'Plink': RdbSSH[GS_Plink].Checked:=True;
         else
             RdbSSH[GS_OpenSSH].Checked:=True;
@@ -2488,6 +2497,7 @@ begin
         end;
     end else if (SSHChoicePage<>NIL) and (CurPageID=SSHChoicePage.ID) then begin
         Result:=RdbSSH[GS_OpenSSH].Checked or
+            (RdbSSH[GS_ExternalOpenSSH].Checked and (FileSearch('ssh.exe', GetEnv('PATH')) <> '')) or
             (RdbSSH[GS_Plink].Checked and FileExists(EdtPlink.Text));
         if not Result then begin
             SuppressibleMsgBox('{#PLINK_PATH_ERROR_MSG}',mbError,MB_OK,IDOK);
@@ -3328,6 +3338,18 @@ begin
     end;
 
     {
+        Optionally "skip" installing bundled SSH binaries conflicting with external OpenSSH:
+    }
+
+#ifdef DELETE_OPENSSH_FILES
+    if RdbSSH[GS_ExternalOpenSSH].Checked then begin
+        WizardForm.StatusLabel.Caption:='Removing bundled Git OpenSSH binaries';
+        if not DeleteOpenSSHFiles() then
+            LogError('Failed to remove OpenSSH file(s)');
+    end;
+#endif
+
+    {
         Set the default Git editor
     }
 
@@ -3462,6 +3484,8 @@ begin
     Data2:='false';
     if (SSHChoicePage=NIL) or RdbSSH[GS_OpenSSH].Checked then begin
         Data:='OpenSSH';
+    end else if RdbSSH[GS_ExternalOpenSSH].Checked then begin
+        Data:='ExternalOpenSSH'
     end else if RdbSSH[GS_Plink].Checked then begin
         Data:='Plink';
         RecordChoice(PreviousDataKey,'Plink Path',EdtPlink.Text);
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -2066,15 +2066,10 @@ begin
         PuTTYPage:=CreatePage(PrevPageID,'Choosing the SSH executable','Which Secure Shell client program would you like Git to use?',TabOrder,Top,Left);
 
         // 1st choice
-        RdbSSH[GS_OpenSSH]:=CreateRadioButton(PuTTYPage,'Use OpenSSH','This uses ssh.exe that comes with Git. The GIT_SSH and SVN_SSH'+#13+'environment variables will not be modified.',TabOrder,Top,Left);
+        RdbSSH[GS_OpenSSH]:=CreateRadioButton(PuTTYPage,'Use bundled OpenSSH','This uses ssh.exe that comes with Git.',TabOrder,Top,Left);
 
         // 2nd choice
-        RdbSSH[GS_Plink]:=CreateRadioButton(PuTTYPage,'Use (Tortoise)Plink',
-            'PuTTY sessions were found in your Registry. You may specify the path'+#13+
-            'to an existing copy of (Tortoise)Plink.exe from the TortoiseGit/SVN/CVS'+#13+
-            'or PuTTY applications. "ssh.variant" will be set in the GIT configuration. '+#13+
-            'The GIT_SSH and SVN_SSH environment variables will be adjusted to point '+#13+
-            'to the following executable:',TabOrder,Top,Left);
+        RdbSSH[GS_Plink]:=CreateRadioButton(PuTTYPage,'Use (Tortoise)Plink','To use PuTTY, specify the path to an existing copy of (Tortoise)Plink.exe:',TabOrder,Top,Left);
         EdtPlink:=TEdit.Create(PuTTYPage);
         EdtPlink.Left:=ScaleX(Left+24);
         EdtPlink.Top:=ScaleY(Top);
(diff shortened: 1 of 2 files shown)


Okeanos:
Another thought … instead of explicitly checking for the Windows supplied OpenSSH binary it might be better to scan the `PATH` for any OpenSSH binary because users can for example use the portable versions from the [official Windows OpenSSH repo](https://github.com/PowerShell/openssh-portable) and manually install them (where ever?).

That'd change the whole thing from "Windows OpenSSH" to "Detected 3rd Party OpenSSH" text wise, functionality wise it'd be the same, just a tad more flexible.

dscho:
> it might be better to scan the `PATH` for any OpenSSH binary

That's a good idea! We would then reword the first option to "Use _bundled_ OpenSSH".



> the installer window is not tall enough to fully display the choices.
> I suppose rewording the texts or omitting the PuTTY/Plink option if it wasn't successfully detected would work?

Yes. I don't think we want to bore the user with implementation details like `ssh.variant` and `GIT_SSH`. Something like this should be plenty sufficient:

```
PuTTY sessions were found. To use them specify the path
to an existing copy of (Tortoise)Plink.exe
```

dscho:
> Also I have no 32Bit system to test this with 🤷‍♂️

I don't think that will be necessary.

installer/install.iss
```diff
@@ -3332,6 +3339,31 @@ begin
...
+            LogError('Line {#__LINE__}: Unable to delete "ssh-add.exe".');
+        if not DeleteFile(AppDir+'\usr\bin\ssh-agent.exe') then
+            LogError('Line {#__LINE__}: Unable to delete "ssh-agent.exe".');
+        if not DeleteFile(AppDir+'\usr\bin\ssh-keygen.exe') then
+            LogError('Line {#__LINE__}: Unable to delete "ssh-keygen.exe".');
+        if not DeleteFile(AppDir+'\usr\bin\ssh-keyscan.exe') then
+            LogError('Line {#__LINE__}: Unable to delete "ssh-keyscan.exe".');
+        if not DeleteFile(AppDir+'\usr\bin\ssh.exe') then
+            LogError('Line {#__LINE__}: Unable to delete "ssh.exe".');
+        if not DeleteFile(AppDir+'\usr\bin\sshd.exe') then
```
dscho, line 3363: I fear that we will need to generate this list from the lines of `pacman -Ql openssh`'s output matching the output of `make-file-list.sh`, i.e. what is available in [the `$LIST` variable](https://github.com/git-for-windows/build-extra/blob/main/installer/release.sh#L158) in `please.sh`.

Every once in a while, a new  `.exe` enters the fray (and of course there is `/usr/lib/ssh/`), and doing it in the way I suggested does not risk `install.iss` going stale.

dscho (changes requested):
This looks awesome! I'd like to ask just for the following changes:

- please wrap the commit message at <=76 columns/line
- we need to dynamically generate the list of OpenSSH files to delete in case of option 3
- Option 3 should be "Use ssh.exe found in the PATH" or something like that

installer/install.iss
Okeanos: I am wondering about that because we specifically need to match against stuff that's already on the `PATH` because someone else supplied it; right now I did it the dumb way and checked the Microsoft supplied list (double checking against my test system) and statically delete the files that should already exist.

I'll see whether I can come up with something but ultimately, yes this should be solved in a maintainable fashion. I am open to suggestions/code snippets, though, because Inno Setup related stuff really isn't my forte 😅.

Okeanos:
So, I fixed some of the outstanding issues; in particular the display and wording of the choice page should now work as desired:

Screenshot 2021-07-24 at 17 50 15

Additionally, I now use the proposed find on path solution to trigger the appearance of the choice page.

I also updated the commit message (<=76 characters per line) to reflect the slightly different scope than initially proposed.

The "what binaries to delete/keep" section in the post install part needs to be updated, still. I am not entirely sure how to do that.

Okeanos:
@dscho I updated the "skip" (deletion) logic – still not fully in the place you want, though. Is there any example of using the `please.sh` output you mentioned within the `install.iss`?

The screenshot below should show that `/usr/bin/*` and `/usr/lib/ssh/*` are now correctly stripped of conflicting binaries.

Screenshot 2021-07-31 at 12 39 01

installer/install.iss
```diff
@@ -3332,6 +3367,25 @@ begin
...
+
+    if RdbSSH[GS_ExternalOpenSSH].Checked then begin
+        WizardForm.StatusLabel.Caption:='Removing bundled Git OpenSSH binaries';
+        Executables:=ListOpenSSHExecutables();
+        for I := 0 to GetArrayLength(Executables) - 1 do
+        begin
+            E := Executables[I];
+            // Only skip installation of binaries that are actually on the path already
+            if (FileSearch(E, GetEnv('PATH')) <> '') and
+               (not
```
dscho, line 3382: Please don't invent a new coding style here, but imitate how things are done elsewhere in `install.iss`. There is never a line break after a `not`.

installer/install.iss
```diff
@@ -548,6 +549,32 @@ begin
...
+    Executables[6]:='ssh-copy-id'; // msys2 only
+    Executables[7]:='ssh-keygen.exe';
+    Executables[8]:='ssh-keyscan.exe';
+    Executables[9]:='ssh-keysign.exe'; // msys2 only
+    Executables[10]:='ssh-pkcs11-helper.exe'; // msys2 only
+    Executables[11]:='ssh-shellhost.exe'; // Win32 only
+    Executables[12]:='ssh-sk-helper.exe'; // msys2 only
+    Executables[13]:='ssh.exe';
+    Executables[14]:='sshd.exe';
+    Result:=Executables;
```
dscho, line 575: The full list of files is a bit longer, and can be obtained in `release.sh` via

```sh
comm -12 <(echo "$LIST" | sort) <(pacman -Ql openssh | sed -n 's|^openssh /\(.*[^/]\)$|\1|p' | sort)
```

When I run it here, it will produce these lines:

```
etc/ssh/moduli
etc/ssh/ssh_config
etc/ssh/sshd_config
usr/bin/findssl.sh
usr/bin/scp.exe
usr/bin/sftp.exe
usr/bin/ssh-add.exe
usr/bin/ssh-agent.exe
usr/bin/ssh-copy-id
usr/bin/ssh-keygen.exe
usr/bin/ssh-keyscan.exe
usr/bin/ssh.exe
usr/bin/sshd.exe
usr/lib/ssh/sftp-server.exe
usr/lib/ssh/ssh-keysign.exe
usr/lib/ssh/ssh-pkcs11-helper.exe
usr/lib/ssh/ssh-sk-helper.exe
usr/share/licenses/openssh/LICENCE
```

If I were you, I would use this invocation as part of writing a function to `config.iss` similar to how `SetSystemConfigDefaults()` is generated. Something along these lines:

```sh
openssh_deletes="$(comm -12 <(echo "$LIST" | sort) \
        <(pacman -Ql openssh | sed -n 's|^openssh /\(.*[^/]\)$|\1|p' | sort) |
    sed -e 'y/\//\\/' -e "s|.*|    if not DeleteFile(AppDir+'\\\\&') then\n        Result:=False;|")"
inno_defines="$inno_defines$LF[Code]${LF}function DeleteOpenSSHFiles():Boolean;${LF}begin$LF    Result:=True;"
inno_defines="$inno_defines$LF$openssh_deletes${LF}end;$LF#define DELETE_OPENSSH_FILES DeleteOpenSSHFiles$LF"
```

And then in `install.iss` at the appropriate point do something like

```innosetup
#ifdef DELETE_OPENSSH_FILES
    if RdbSSH[GS_ExternalOpenSSH].Checked then begin
        WizardForm.StatusLabel.Caption:='Removing bundled Git OpenSSH binaries';
        if not DELETE_OPENSSH_FILES() then
            LogError('Failed to remove OpenSSH file(s)');
end;
```

Note: this is fully untested, therefore I expect that you will have to work on this for a bit until it does what it is expected to do.

installer/install.iss
```diff
@@ -2058,28 +2085,28 @@ begin
...
-    if RegGetSubkeyNames(HKEY_CURRENT_USER,'Software\SimonTatham\PuTTY\Sessions',PuTTYSessions) and (GetArrayLength(PuTTYSessions)>0) then begin
-        PuTTYPage:=CreatePage(PrevPageID,'Choosing the SSH executable','Which Secure Shell client program would you like Git to use?',TabOrder,Top,Left);
+    if (RegGetSubkeyNames(HKEY_CURRENT_USER,'Software\SimonTatham\PuTTY\Sessions',PuTTYSessions) and (GetArrayLength(PuTTYSessions)>0)) or
+       (FileSearch('ssh.exe', GetEnv('PATH')) <> '') then begin
+        SSHChoicePage:=CreatePage(PrevPageID,'Choosing the SSH executable','Which Secure Shell client program would you like Git to use?',TabOrder,Top,Left);
 
         // 1st choice
-        RdbSSH[GS_OpenSSH]:=CreateRadioButton(PuTTYPage,'Use OpenSSH','This uses ssh.exe that comes with Git. The GIT_SSH and SVN_SSH'+#13+'environment variables will not be modified.',TabOrder,Top,Left);
+        RdbSSH[GS_OpenSSH]:=CreateRadioButton(SSHChoicePage,'Use bundled OpenSSH',
+            'This uses ssh.exe that comes with Git.',
```
dscho, line 2098: Please avoid introducing a coding style that is inconsistent with what is already in this file.

installer/install.iss
```diff
@@ -2058,28 +2085,28 @@ begin
...
-        RdbSSH[GS_Plink]:=CreateRadioButton(PuTTYPage,'Use (Tortoise)Plink',
-            'PuTTY sessions were found in your Registry. You may specify the path'+#13+
-            'to an existing copy of (Tortoise)Plink.exe from the TortoiseGit/SVN/CVS'+#13+
-            'or PuTTY applications. "ssh.variant" will be set in the GIT configuration. '+#13+
-            'The GIT_SSH and SVN_SSH environment variables will be adjusted to point '+#13+
-            'to the following executable:',TabOrder,Top,Left);
-        EdtPlink:=TEdit.Create(PuTTYPage);
+        RdbSSH[GS_Plink]:=CreateRadioButton(SSHChoicePage,'Use (Tortoise)Plink',
+            'To use PuTTY, specify the path to an existing copy of (Tortoise)Plink.exe:',
+            TabOrder,Top,Left);
```
dscho, line 2104: Since this changes the PuTTY text, which is independent of the change that adds a third option, I would like to encourage you to refactor this change out into its own commit.

installer/install.iss
```diff
@@ -2058,28 +2085,28 @@ begin
...
+        EdtPlink:=TEdit.Create(SSHChoicePage);
         EdtPlink.Left:=ScaleX(Left+24);
         EdtPlink.Top:=ScaleY(Top);
         with EdtPlink do begin
-            Parent:=PuTTYPage.Surface;
+            Parent:=SSHChoicePage.Surface;
```
dscho, line 2109: Likewise, the change from `PuTTYPage` to `SSHChoicePage` is a bit distracting from the _interesting_ part of this PR. I would therefore prefer it refactored into yet another commit, before the final commit that only adds the third choice.

dscho (commented):
Hopefully this (quite lengthy) description how to generate the function to delete the OpenSSH files works for you.

By the way, I strongly advise not to special-case `ssh-copy-id`: There is no guarantee whatsoever that using a copy of this script from MSYS2's OpenSSH will work with the Win32 variant of OpenSSH at all.

installer/install.iss
Okeanos: Sorry about that; will fix this.

installer/install.iss
Okeanos: Will have a look, thanks!

installer/install.iss
Okeanos: Is a separate commit okay or should I put it into a separate pull request?

Okeanos:
@dscho thank you so much for your input – I believe I have addressed the necessary structural changes and also verified that they work as expected. Actually even checking out a private repository using Win32 SSH with git clone via the git bash.

There are some remnants:
- the ssh folders are not deleted 🤷‍♂️ 
- `ssh-pageant.exe` survives – is that a problem?

Other than that I believe I only have to address the separate commits for the PuTTY description changes, the SSHChoice Page renaming, and finally the actual 3rd option being added with the additional new functionality. Would that be fine in a single pull request or would you like to receive this in different ones?

dscho:
> Other than that I believe I only have to address the separate commits for the PuTTY description changes, the SSHChoice Page renaming, and finally the actual 3rd option being added with the additional new functionality. Would that be fine in a single pull request or would you like to receive this in different ones?

I would prefer the separate commits as "preparatory" ones, i.e. they should be extracted into their own commits that come _before_ the commit that adds that 3rd option (which should be the tip commit of your branch).

Also, could I ask for one more favor? I'd like the commit messages to all begin with `installer:`, so that browsing the output of `git log` is less confusing (there are quite a few more things in `build-extra` than just the installer definition).

dscho:
> I would prefer the separate commits as "preparatory" ones, i.e. they should be extracted into their own commits that come _before_ the commit that adds that 3rd option (which should be the tip commit of your branch).

Right, I did not explicitly state this: I would like the commits to all live in this here PR. Thank you!

installer/install.iss
dscho: A separate commit in the same PR is totally okay.

dscho:
> I only have to address the separate commits for the PuTTY description changes, the SSHChoice Page renaming, and finally the actual 3rd option being added with the additional new functionality.

BTW if you have never done this before, I hope https://git-scm.com/docs/git-rebase#_splitting_commits will be helpful.

Okeanos:
I have now pushed the requested changes (the link did indeed help, thanks!) to this pull request:

1. update descriptions
2. rename page
3. add external ssh support

Each commit is, I think, correctly signed-off and contains only the relevant changes with a 'why' description.

As discussed and acknowledged the empty ssh folder(s) remain(s) and `ssh-pageant.exe` is also left untouched.

I also verified once again that the changes work as expected by building the installer and using it as an upgrade to an existing Git installation. The SSH Choice Page is only displayed if the checkbox for "Only show new options" is **unchecked**, i.e. all pages are shown, including previous choices. Should that be changed as well? If so, how do I do that? Is that lines 1077-1079 in `install.iss`?

Other than that, the integration works as I expected it to: git (bash) picks up the Win32 OpenSSH binaries and uses them, also making use of e.g. an SSH Key managed via KeePassXC in the Win32 ssh-agent. Sadly, on the Git UI side this is a little less well supported. GitHub Desktop (desktop/desktop/issues/5641), SourceTree, etc. don't really support Win32 OpenSSH 🤷‍♂️ . However, on the bright side Visual Studio Code's default Git integration and IntelliJ IDEA both were able to successfully use it :)

dscho (approved):
Thank you for patiently addressing my feedback!

Okeanos:
You are welcome; thank you so much for your support and patience when dealing with my questions as well. Making sure your feedback is addressed was the least I could do :) – so cool this happened!

dscho:
FWIW I quickly added a commit on top that will hide the PuTTY option again, unless at least one PuTTY session was found. I'm in the process of verifying that it works as intended, and will probably push the result shortly.

dscho:
Forgot to send this:

> The SSH Choice Page is only displayed if the checkbox for "Only show new options" is **unchecked**

I think that is still correct because _technically_ the user saw this page, even if there is now a new option available.

dscho:
> FWIW I quickly added a commit on top that will hide the PuTTY option again, unless at least one PuTTY session was found. I'm in the process of verifying that it works as intended, and will probably push the result shortly.

Pushed: 53f76b906

Okeanos:
> Forgot to send this:
> 
> > The SSH Choice Page is only displayed if the checkbox for "Only show new options" is **unchecked**
> 
> I think that is still correct because _technically_ the user saw this page, even if there is now a new option available.

I am not entirely sure: the PuTTY Page only appeared if PuTTY was found to begin with. Now it should appear if PuTTY is found or an alternative ssh.exe exists on the `PATH`. The latter is entirely new and should be shown to all upgrading users, even if they never had PuTTY on their system but do have an alternative ssh.exe. I would fall into the latter category and wouldn't be shown this page unless I explicitly replay all choices, i.e. uncheck "only show new options".

I'll defer to your choice, though.

carenas:
Not sure what might had gone wrong with the integration of this series, but using the rc2 installer as both an upgrade from 2.32.0 and clean install still leaves the bundled ssh files behind even if the option to use the external ssh was selected and it is available in path.

@dscho: I am setting up a development environment to see if I can debug this, but thought it was worth a headsup.

@Okeanos: did you confirm this works with the installer from?:

  https://github.com/git-for-windows/git/releases/tag/v2.33.0-rc2.windows.1

Okeanos:
@carenas I did not try the official RC builds yet but _did_ use SDK built installers as documented in this pull request's comments. I'll certainly check on this later today and give additional feedback here.

dscho:
Oy. The only thing I could think of to go wrong is that maybe the function in `config.inc.iss` isn't generated at all, for whatever reason :-(

@carenas would you have time to analyze the logs of the Git artifacts check of the -rc2 PR? I am on a phone, and technically offline today, which makes it a bit difficult for me to investigate.

carenas:
@dscho: sure, but I am not really sure what to look at
@Okeanos: FWIW I couldn't get it to work either with the rc1 installer

dscho:
> @dscho: sure, but I am not really sure what to look at

(I briefly interrupted my day off to look at this.) What I did was to open https://dev.azure.com/git-for-windows/git/_build/results?buildId=82662, go to the job where the installer was built, then selected the step where the installer was built, then clicked on the magnifying glass symbol above the logs and typed in "openssh", then continued to search until I found the correct spot.

I am fairly certain that the problem lies with https://dev.azure.com/git-for-windows/git/_build/results?buildId=82662&view=logs&j=fd84893c-d8cf-515f-6f6b-748e5f0b9fc1&t=53291ad3-d296-5424-da40-80462f49966c&l=31232:

> ```
> D:/a/1/s/git-sdk-64-build-installers/usr/src/build-extra/installer/release.sh: command substitution: line 284: syntax error near unexpected token `('
> D:/a/1/s/git-sdk-64-build-installers/usr/src/build-extra/installer/release.sh: command substitution: line 284: `comm -12 <(echo "$LIST" | sort) <(pacman -Ql openssh | sed -n 's|^openssh /\(.*[^/]\)$|\1|p' | sort) |'
> ```

dscho:
So the problem lies with `sh.exe` vs `bash.exe` (even if they are identical over here, they must do something slightly different based on their basename):

```console
$ sh -c 'comm -12 <(printf "%s\n" 1 2) <(printf "%s\n" 2 3)'
sh: -c: line 0: syntax error near unexpected token `('
sh: -c: line 0: `comm -12 <(printf "%s\n" 1 2) <(printf "%s\n" 2 3)'

$ bash -c 'comm -12 <(printf "%s\n" 1 2) <(printf "%s\n" 2 3)'
2
```

dscho:
I hope that this is fixed in 863470a56a50d496bf41e6c604219b1f625251a6 and I started https://dev.azure.com/git-for-windows/git/_build/results?buildId=82722&view=results. @carenas could I ask you to monitor that build and test?

rimrul:
It probably tries to follow POSIX more strictly when called as `sh` and the `<(foo)` construct isn't POSIX. We could try to [emulate it with fifos](https://mywiki.wooledge.org/Bashism), though.

dscho:
> It probably tries to follow POSIX more strictly when called as `sh`

That's my guess, too, but this is the first time I see it in action (that I remember).

> and the `<(foo)` construct isn't POSIX.

Indeed!

> We could try to [emulate it with fifos](https://mywiki.wooledge.org/Bashism), though.

I "emulated" it by writing to intermediate files instead. That definitely works always, even if you mix and match MSYS with MINGW executables (the latter cannot handle MSYS-emulated fifos).

dscho:
> 
> 
> I hope that this is fixed in [863470a](https://github.com/git-for-windows/build-extra/commit/863470a56a50d496bf41e6c604219b1f625251a6) and I started https://dev.azure.com/git-for-windows/git/_build/results?buildId=82722&view=results. @carenas could I ask you to monitor that build and test?

Darn it. It failed, of course, with:

> Already tagged: v2.33.0-rc2.windows.1

Which is technically correct.

So I kicked off the GitHub workflow that _should_ be equivalent: https://github.com/git-for-windows/git/runs/3320018811?check_suite_focus=true.

carenas:
@dscho: failed again (x86 first and canceled x86_64), IMHO might be a good idea to use a new tag

dscho:
I kicked off an `installer-x86_64`-only run: https://github.com/git-for-windows/git/actions/runs/1126976271

dscho:
And I pushed a dummy revision so that I could kick off https://dev.azure.com/git-for-windows/git/_build/results?buildId=82726&view=results.

carenas:
@dscho: failed again in a similar way than the last x86 run:

```
+ cp PKGBUILD.v2.33.0-rc2.windows.1-20210813082640 PKGBUILD
cp: cannot stat 'PKGBUILD.v2.33.0-rc2.windows.1-20210813082640': No such file or directory
Error: Process completed with exit code 1.
```

carenas:
> And I pushed a dummy revision so that I could kick off https://dev.azure.com/git-for-windows/git/_build/results?buildId=82726&view=results.

this one still running, and hopefully complete since it has a different "tag"

dscho:
I think this one worked: https://dev.azure.com/git-for-windows/git/_build/results?buildId=82726&view=logs&j=fd84893c-d8cf-515f-6f6b-748e5f0b9fc1&t=53291ad3-d296-5424-da40-80462f49966c&l=37234

carenas:
and has been tested and confirmed golden!

dscho:
@carenas thank you so much for noticing, pointing out, and verifying the fix.

dscho:
Having said that, I think we have another problem on our hands:

```console
$ git ls-remote <ssh-host>
Warning: Permanently added the RSA host key for IP address '40.74.28.12' to the list of known hosts.
CreateProcessW failed error:193
ssh_askpass: posix_spawnp: Unknown error
Permission denied, please try again.
CreateProcessW failed error:193
ssh_askpass: posix_spawnp: Unknown error
Permission denied, please try again.
CreateProcessW failed error:193
ssh_askpass: posix_spawnp: Unknown error
git@ssh.dev.azure.com: Permission denied (password,publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

This happens when I try to access an SSH host without having a private key. And I _think_ the reason is this:

```console
$ env | grep ASKPASS
SSH_ASKPASS=/mingw64/libexec/git-core/git-gui--askpass
```

The `git-gui--askpass` file implements a Tcl/Tk script, with a `#!` line that Windows' OpenSSH obviously doesn't know how to handle.

I _think_ what we can do is to use GCM Core's `askpass` helper (which is a `.exe` that Windows' OpenSSH can spawn without problems). Like this:

```console
$ SSH_ASKPASS=/mingw64/libexec/git-core/git-askpass.exe
```

After that, I am greeted with this dialog when I try to connect:



The `SSH_ASKPASS` variable gets defined in `/etc/profile.d/env.sh` (a file which is tracked here: https://github.com/git-for-windows/build-extra/blob/HEAD/git-extra/env.sh).

I guess we could improve things by testing whether `/mingw64/libexec/git-core/git-askpass.exe` exists here and prefer it over the Tcl/Tk script.

rimrul:
Avoiding a `Tcl/Tk` based `askpass` might be a good idea in general, when I think back to #234.

rimrul:
Looking at the GCM core repo, they don't seem to ship a `git-askpass.exe`, yet. I think the one you found is from the classic GCM for Windows.

Okeanos:
@dscho thank you so much for spending the time and fixing this on so short notice and everything ❤️ . If there is anything I can do to try and help with the `git-askpass.exe` (which I obviously failed to notice during my tests as well) problem please say so.

dscho:
> 
> 
> Looking at the GCM core repo, they don't seem to ship a `git-askpass.exe`, yet. I think the one you found is from the classic GCM for Windows.

Oh, I think you're right... 🤦 

> Avoiding a `Tcl/Tk` based `askpass` might be a good idea in general, when I think back to #234.

That's a good point. We should reimplement a pure Win32 `askpass.exe`. For obvious reasons, we cannot call it `git-askpass.exe` (because it would clash with GCM's executable).

dscho:
> we cannot call it `git-askpass.exe` (because it would clash with GCM's executable).

Oh, now that I said it: I think we actually _can_ call it `git-askpass.exe`, because we will install it into `bin`, not into `libexec/git-core`...

rimrul:
I've created a pretty barebones Win32 `askpass` in #371.

installer/install.iss
```diff
@@ -3327,6 +3337,18 @@ begin
         InstallWindowsTerminalFragment();
     end;
 
+    {
+        Optionally "skip" installing bundled SSH binaries conflicting with external OpenSSH:
+    }
+
+#ifdef DELETE_OPENSSH_FILES
+    if RdbSSH[GS_ExternalOpenSSH].Checked then begin
```
PetSerAl, line 3345: Should also check for `SSHChoicePage<>NIL`.

carenas: Isn't it implied?, how could you get this Checked if there was no form to select it?

PetSerAl: There is third option `RdbSSH[GS_ExternalOpenSSH]` was never initialized, thus referencing to `Checked` property is not allowed. https://github.com/git-for-windows/git/issues/3368

Okeanos: Sorry about that; I didn't have a Windows 7 installation (or any Windows installation prior to the Win32 SSH rollout) around to test this with and it didn't occur to me that I could've used the [IE Developer VMs](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/) to test this, until now. Interestingly, the Win7 IE 11 VM I downloaded comes with [OpenSSH preinstalled](https://www.mls-software.com/opensshd.html).

PR #368, opened 2021-07-27
GCM Core: don't strip signatures

rimrul:
Stripping assembly signatures out of PE files without adjusting the header information leads to corrupted binaries. That causes issues with our effort to cresate an MSIX package.


dscho (approved):

dscho:
Good find! I just verified that the original `WebView2Loader.dll` can be signed without any problem.

PR #370, opened 2021-08-12
Some hot fixes for the installer, in preparation for v2.33.0-rc2

dscho:
This fixes three problems I just saw while testing a tentative version of the -rc2 installer:

- The running instances of the FSMonitor daemon did not shut down properly because the stop command was invoked from the gitdirs instead of the worktree directories, and FSMonitor now mistakes the former scenario for being called in a bare repository.
- The components page was shown still, to show the Windows Terminal Profile option (which was new in v2.32.0), even if upgrading _from v2.32.0_.
- The "external SSH" choice was not shown when running the installer in the "Only show unseen options" mode, even if it could not possibly have been seen before v2.33.0.

[EDIT] While at it, I also added the missing entry in the release notes where I want to mention that the FSMonitor patches were updated.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -610,7 +610,7 @@ begin
                 BuiltinFSMonitorStopOption:='(huh?)';
                 if not ExecWithCapture('"'+AppDir+'\cmd\git.exe" fsmonitor--daemon -h',Str,Str,ExitCode) or (ExitCode<>129) then begin
                     if (i<>1) and (i<>127) then // Suppress message if `git.exe` was not found, or if it does not know about the built-in FSMonitor
-                        LogError('Could not get FSMonitor help (exit code'++IntToStr(ExitCode)+'):'+#13+Str);
+                        LogError('Could not get FSMonitor help (exit code '+IntToStr(ExitCode)+'):'+#13+Str);
                     Exit;
                 end else begin
                     i:=Pos('stop'+#10,Str);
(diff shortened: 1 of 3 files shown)


installer/install.iss
```diff
@@ -989,7 +997,10 @@ begin
 #ifdef GIT_VERSION
     if Result or (CountDots(CurrentVersion)>3) or (CountDots(PreviousGitForWindowsVersion)>3) then begin
         // maybe the previous version was a prerelease (prereleases have five numbers: v2.24.0-rc1.windows.1 reduces to '2.24.0.1.1')?
-        CurrentVersion:='{#GIT_VERSION}';
+        if CurrentVersion=ExpandConstant('{#APP_VERSION}') then
+            CurrentVersion:='{#GIT_VERSION}'
+        else
+            CurrentVersion:='git version '+CurrentVersion+'.windows.1';
```
vdye, line 1003: I'm probably misreading this, but wouldn't the setting `CurrentVersion` to `GIT_VERSION` yield something like "v2.33.0-rc2.windows.1", but setting it to that custom string would get something like `git version v2.33.0-rc2.windows.1`? Why is the `git version` prefix needed?

derrickstolee (approved):

vdye (approved):
The changes make sense with the issues you listed - just had one question about the downgrade version checks.

jeffhostetler (approved):

installer/install.iss
dscho: `GIT_VERSION` is set here: https://github.com/git-for-windows/build-extra/blob/37805f958614c014fb86bbae91f08d11f9f90821/installer/release.sh#L165-L166

As you can see, it is the output of `git version`, unprocessed. The `git version ` prefix is part of that output.

PR #373, opened 2021-08-19
make-file-list: include git-gui and gitk l10n in WITH_L10N

rimrul:
When asked to build an installer or a portable with localization,
we should also add the localization for git-gui and gitk.


dscho (approved):

dscho:
Thank you!

I'm curious: do you have a need for localized versions of Git for Windows?

rimrul:
No I don't. I've just seen a mail on the mailing list about our `gitk` (from the regular installers) not including l10n. While replying I noticed that even when building installers with l10n they don't cover `gitk` and `git gui` and that annoyed me.

PR #377, opened 2021-09-03
Drop Git Credential Manager for Windows

dscho:
This is _not_ a change to drop Git Credential Manager!

Git for Windows ships with two different versions of Git Credential Manager:

- **Git Credential Manager _for Windows_**
  This has been the original (and in the beginning, the only) version of Git Credential Manager.

  It was borne out of the need for something more stable, more robust, more actively-developed than `git-credential-wincred`, to allow e.g. for Two-Factor-Authentication support. It is based on .NET Framework and hence limited to supporting Windows, and Windows only.

  It is no longer actively maintained.
- **Git Credential Manager _Core_**
  This version supersedes the first, tries to accomplish the same on Windows, but targets .NET Core, i.e. supports multiple platforms. In particular, it supports macOS and Linux.

  It is maintained actively.

Since the first version is no longer maintained, GitHub stopped accepting password-based authentication. In other words, Git Credential Manager for Windows is now obsolete, it does not even manage to allow logging into GitHub.

Therefore, it is high time to just drop the first version, and henceforth only support and include Git Credential Manager Core.


installer/install.iss
```diff
@@ -383,8 +383,7 @@ const
 
     // Git Credential Manager settings.
     GCM_None          = 1;
-    GCM_Classic       = 2;
-    GCM_Core          = 3;
```
mjcheetham, line 387: Just to check, these values are local only to the installer, and not stored or persisted to the registry/install-options?

mjcheetham (approved):

installer/install.iss
rimrul: Yes, these are just an internal index for the array of radio buttons. The persisted values are human readable strings, not numeric values.

installer/install.iss
```diff
@@ -2265,32 +2264,22 @@ begin
...
     if DetectNetFxVersion()<378675 then begin
-        RdbGitCredentialManager[GCM_Classic].Checked:=False;
-        RdbGitCredentialManager[GCM_Classic].Enabled:=False;
         RdbGitCredentialManager[GCM_Core].Checked:=False;
         RdbGitCredentialManager[GCM_Core].Enabled:=False;
     end else begin
         case ReplayChoice('Use Credential Manager','Core') of
             'Disabled': RdbGitCredentialManager[GCM_None].Checked:=True;
-            'Enabled': RdbGitCredentialManager[GCM_Classic].Checked:=True;
+            'Enabled': RdbGitCredentialManager[GCM_Core].Checked:=True;
```
rimrul, line 2278: @mjcheetham the Enabled/Disabled/Core here is what's persisted.

PR #378, opened 2021-09-05
git-extra/wordpad: teach wordpad how to deal with spaces in path

carenas:
When wordpad path has spaces (ex: C:\Program Files\Windows NT\),
invoking the binary will fail.

Add quotes around the $WORDPAD variable to avoid that.

Signed-off-by: Carlo Marcelo Arenas Belón <carenas@gmail.com>

### DIFF
--- a/git-extra/wordpad
+++ b/git-extra/wordpad
@@ -17,10 +17,10 @@ esac
 
 case "$1" in
 */.git/*) ;; # needs LF line endings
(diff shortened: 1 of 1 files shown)


dscho:
Thank you!

dscho:
The package [is being built and uploaded](https://dev.azure.com/Git-for-Windows/git/_build/results?buildId=84363&view=results).

PR #381, opened 2021-09-22
git-askpass: don't expect a prompt

rimrul:
OpenSSH considers the prompt [an optional parameter
to their ssh-askpass](https://man.openbsd.org/ssh-askpass.1) and calls git-askpass without
a prompt as well. Provide a default prompt that instead
of erroring out.

This fixes https://github.com/git-for-windows/git/issues/3434


dscho:
Thank you!

PR #385, opened 2021-10-15
Hopefully fix git-for-windows/git#3471

segevfiner:
They also include a `bin\code` bash script which seems to work, executing a `.cmd` doesn't seem to work for some reason.

This fixes https://github.com/git-for-windows/git/issues/3471

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1850,15 +1850,15 @@ begin
 
     // Remove `" %1"` from end and unqote the string.
     if (EditorAvailable[GE_VisualStudioCode]) then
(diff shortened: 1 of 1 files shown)


dscho:
@segevfiner could I trouble you to work a bit on the commit message? It would be good to
- prefix the commit message with "installer"
- mention in the first line that this is about a quoting bug, and about VS Code
- it would be good to mention that the typical trigger is a space in VS Code's path
- clarify who is meant by "They" (= VS Code)
- include the information I provided in https://github.com/git-for-windows/git/issues/3471#issuecomment-944223210

Thank you!

PR #387, opened 2021-10-30
Change path of "recently seen" config from HOME to APPDATA

sryze:
Related to https://github.com/git-for-windows/git/issues/3503

Tested this locally by updating `C:\Program Files\Git\mingw64\bin\git-update-git-for-windows`

### DIFF
--- a/git-extra/git-update-git-for-windows
+++ b/git-extra/git-update-git-for-windows
@@ -134,7 +134,9 @@ http_get () {
 }
 
 get_recently_seen() {
-    if [ -f "$HOME"/.git-for-windows-updater ]; then
+    if [ -f "$APPDATA/Git/updater.ini" ]; then
+        git config -f "$APPDATA/Git/updater.ini" winUpdater.recentlySeenVersion
+    elif [ -f "$HOME"/.git-for-windows-updater ]; then
         git config -f "$HOME"/.git-for-windows-updater winUpdater.recentlySeenVersion
     else
         git config --global winUpdater.recentlySeenVersion
@@ -142,7 +144,8 @@ get_recently_seen() {
 }
 
 set_recently_seen() {
-    git config -f "$HOME"/.git-for-windows-updater winUpdater.recentlySeenVersion "$1"
+	mkdir -p "$APPDATA/Git" &&
+	git config -f "$APPDATA/Git/updater.ini" winUpdater.recentlySeenVersion "$1"
 }
 
 # The main function of this script


dscho (commented):
The change looks sound, but I fear that the commit message does not do anything to convince the casual (and also the not-so-casual) reader that this change should be integrated.

Why, exactly, should we not write into the home directory? We write `.gitconfig` in the home directory anyway. And this change would not remove any existing files, anyway, so maybe it agrees that those files in the home directory are not all that bad to begin with?

sryze:
Good point... I guess it does not make a lot of sense for git

I just feel like more and more "cross-platform" programs/tools tend to pollute the user's home directory with their internal config files on Windows (because the authors are too lazy or whatever), even Microsoft's own tools lol, and I really don't like this trend. I like to keep my user folder clean.

dscho:
> because the authors are too lazy or whatever

If you want to assume malice, that's a sensible take on it. In the case of Git, it is much simpler, though: consistency.

colemickens:
I mean, could it *at least be marked as a hidden file*?

dscho:
> I mean, could it _at least be marked as a hidden file_?

Yes, it could be. Please provide a PR to do so. You will want to append the appropriate command to [this line](https://github.com/git-for-windows/build-extra/blob/4756223ce8c2a1c7afadef2781afb733e16b7d81/git-extra/git-update-git-for-windows#L147).

PhilipOakley:
>     I mean, could it at least be marked as a hidden file?
>
> Yes, it could be. Please provide a PR to do so. You will want to
> append the appropriate command to this line
> <https://github.com/git-for-windows/build-extra/blob/4756223ce8c2a1c7afadef2781afb733e16b7d81/git-extra/git-update-git-for-windows#L147>.
>
>
While at it, you could consider cleaning up any other
`recentlySeenVersion` entries that may have been historically set by
`get_recently_seen()`
https://github.com/git-for-windows/build-extra/blob/4756223ce8c2a1c7afadef2781afb733e16b7d81/git-extra/git-update-git-for-windows#L138-L144

e.g. I have in my `--global` (C:\Users\phili\.gitconfig) an entry:
```
[winUpdater]
    recentlySeenVersion = 2.25.0.windows.1
```

while in C:\Users\phili\.git-for-windows-updater it has
```
[winUpdater]
    recentlySeenVersion = 2.33.1.windows.1
```

I'm now on
$ git version
git version 2.34.0.windows.1

$ echo $HOME
/c/Users/phili

$ echo $APPDATA
C:\Users\phili\AppData\Roaming

for info, There's nothing git related in my
C:\Users\phili\AppData\Roaming at present.

Hope That Helps (HTH)
--
Philip

PR #388, opened 2021-10-31
Fix GIT_SDK_SH_PATH in Zsh

BasixKOR:
Fixes `realpath: '': No such file or directory` error when using zsh as a login shell. Also corrects `$GIT_SDK_SH_PATH`.


dscho (changes requested):
Please do not send readers of the commit message chasing down information that should be included in the commit message.

It is okay, of course, to summarize the issue and the solution and then link to an external source.

Just linking, without any further information, is insufficient.

dscho:
@BasixKOR are you interested in getting this merged? If so, please address the concerns I offered.

BasixKOR:
Sorry, I was working on other things. Let me quickly make some changes on the commit message.

dscho (changes requested):
Great, thanks for improving the commit message. Just two things:

- The sign-off should have your real name.
- The PR build failed because you need to update the checksum for the `git-sdk.sh` file. The easiest way to do that is to run `updpkgsums` in `git-extra/`, then commit the updated `PKGBUILD` (I would suggest to amend the commit, not to make a separate one).

dscho:
@BasixKOR unfortunately, the `main` branch was modified in the meantime (quite a bit, actually) and now conflicts. I resolved this locally and will force-push in a moment.

dscho (approved):

PR #390, opened 2021-11-15
install.iss: use --replace-all when setting config values

derrickstolee:
When testing the 2.34.0 installer, I saw this confusing page of the wizard:

image (1)

Changing the selection to GCM Core got me this error after the rest of everything was installed:

image (2)

By using `--replace-all`, we get rid of this error.

The Wizard page's default setting might still be incorrect. My system config had multiple values because GCM configured it for the system, but used an exact path to the executable instead of just `manager-core`.

```
$ git config --show-origin --get-all credential.helper
file:C:/Program Files/Git/etc/gitconfig manager-core
file:C:/Program Files/Git/etc/gitconfig
file:C:/Program Files/Git/etc/gitconfig C:/Program\ Files\ \(x86\)/Git\ Credential\ Manager\ Core/git-credential-manager-core
file:C:/Users/dstolee/.gitconfig
file:C:/Users/dstolee/.gitconfig        C:/Program\ Files\ \(x86\)/Git\ Credential\ Manager\ Core/git-credential-manager-core
```

**Update:** I added a second commit that attempts to correct for this case. The `--replace-all` would still modify the system config incorrectly (it would only have one value, `manager-core`), but at least it wouldn't clear the system config entirely. Hopefully GCM Core configures itself correctly despite these things the Git installer is doing?

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -758,14 +758,14 @@ var
     StdOut,StdErr:String;
 begin
     if (Value=#0) then begin
-        if ExecWithCapture('"'+AppDir+'\{#MINGW_BITNESS}\bin\git.exe" config --system --unset '+Key,StdOut,StdErr,ExitCode) And ((ExitCode=0) Or (ExitCode=5)) then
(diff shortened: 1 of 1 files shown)


dscho (approved):
Nice!

PR #391, opened 2021-11-16
install.iss: fix if/else syntax

derrickstolee:
This previously had this compile error:

  Error on line 858 in C:\(...)\install.iss: Column 33:
  Identifier expected

The reason we didn't catch this earlier was because we had inserted some
logging that changed the nature of this if/else to have begin/end
blocks.

Fixes the bug introduced by #390.


dscho (approved):
D'oh! I should have compile-tested _without_ the debug statements I added ;-)

PR #393, opened 2021-11-17
Revert "installer: allow configuring the default behavior of `git pull`"

alexhenrie:
This reverts commit 6ab14d9a169d7aeb3fdcf861148ec4d1a0de91cf.

The reason why `git pull` now aborts by default when a fast-forward is
not possible is to prevent accidental creation of merge commits when
rebasing would be more appropriate (or vice-versa). Encouraging users to
always set a different default globally (which they will do without
thinking) defeats the safeguard against such a mistake.

Signed-off-by: Alex Henrie <alexhenrie24@gmail.com>

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -376,11 +376,6 @@ const
     GB_MinTTY         = 1;
     GB_ConHost        = 2;
 
-    // `git pull` behavior settings.
-    GP_GitPullMerge   = 1;
-    GP_GitPullRebase  = 2;
-    GP_GitPullFFOnly  = 3;
-
     // Git Credential Manager settings.
     GCM_None          = 1;
     GCM_Core          = 2;
@@ -485,10 +480,6 @@ var
     // Wizard page and variables for the terminal emulator settings.
     BashTerminalPage:TWizardPage;
     RdbBashTerminal:array[GB_MinTTY..GB_ConHost] of TRadioButton;
-
-    // Wizard page and variables for the `git pull` options.
-    GitPullBehaviorPage:TWizardPage;
-    RdbGitPullBehavior:array[GP_GitPullMerge..GP_GitPullFFOnly] of TRadioButton;
 
     // Wizard page and variables for the credential manager options.
     GitCredentialManagerPage:TWizardPage;
@@ -2238,30 +2229,6 @@ begin
     end;
 
     (*
-     * Create a custom page for the default behavior of `git pull`.
-     *)
-
-    GitPullBehaviorPage:=CreatePage(PrevPageID,'Choose the default behavior of `git pull`','What should `git pull` do by default?',TabOrder,Top,Left);
-
-    // 1st choice
-    RdbGitPullBehavior[GP_GitPullMerge]:=CreateRadioButton(GitPullBehaviorPage,'Default (fast-forward or merge)','This is the standard behavior of `git pull`: fast-forward the current branch to'+#13+'the fetched branch when possible, otherwise create a merge commit.',TabOrder,Top,Left);
-
-    // 2nd choice
-    RdbGitPullBehavior[GP_GitPullRebase]:=CreateRadioButton(GitPullBehaviorPage,'Rebase','Rebase the current branch onto the fetched branch. If there are no local'+#13+'commits to rebase, this is equivalent to a fast-forward.',TabOrder,Top,Left);
-
-    // 3rd choice
-    RdbGitPullBehavior[GP_GitPullFFOnly]:=CreateRadioButton(GitPullBehaviorPage,'Only ever fast-forward','Fast-forward to the fetched branch. Fail if that is not possible.',TabOrder,Top,Left);
-
-    // Restore the setting chosen during a previous install.
-    case ReplayChoice('Git Pull Behavior Option','Merge') of
-        'Merge': RdbGitPullBehavior[GP_GitPullMerge].Checked:=True;
-        'Rebase': RdbGitPullBehavior[GP_GitPullRebase].Checked:=True;
-        'FFOnly': RdbGitPullBehavior[GP_GitPullFFOnly].Checked:=True;
-    else
-        RdbGitPullBehavior[GP_GitPullMerge].Checked:=True;
-    end;
-
-    (*
      * Create a custom page for the choice of Git Credential Manager.
      *)
 
@@ -3086,19 +3053,6 @@ begin
     end;
 
     {
-        Configure the default `git pull` behavior
-    }
-
-    WizardForm.StatusLabel.Caption:='Configuring default `git pull` behavior';
-    if RdbGitPullBehavior[GP_GitPullMerge].Checked then begin
-        GitSystemConfigSet('pull.rebase','false')
-    end else if RdbGitPullBehavior[GP_GitPullRebase].Checked then begin
-        GitSystemConfigSet('pull.rebase','true')
-    end else if RdbGitPullBehavior[GP_GitPullFFOnly].Checked then begin
-        GitSystemConfigSet('pull.ff','only')
-    end;
-
-    {
         Configure credential helper
     }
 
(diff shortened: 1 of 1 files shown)


dscho:
I found the message about `git pull` _requiring_ that config setting to be configured highly disruptive to my work. And users who are even less knowledgeable in Git would probably have been disrupted even more. That's why I introduced that option.

So I am quite doubtful that reverting it is something we want to do.

alexhenrie:
I realize that when the message was first introduced, it was very naggy, but in recent versions of Git it only appears if a fast-forward is not possible. There's also a short option `git pull -r` which you might find easier in your personal workflow than configuring the `pull.rebase` option.

The consensus on the Git mailing list was essentially that it's much more disruptive to accidentally create a merge commit and have to undo it than to ask for confirmation first. I know from experience that Git novices run `git pull` all the time without thinking and then project maintainers complain when the novices' upstream pull requests contain unnecessary merge commits.

dscho:
The problem is that we already have offered this option for ages. If we remove it, we force everybody to answer the question _again_.

I don't doubt that your intentions are good. But the cost of what you propose is _really_ high.

Besides, I suspect that most users would (after a lot of muttering) choose the `merge` option anyway. Meaning: your project maintainers will have to continue to deal with merge commits _despite_ the disruption caused by this change that was supposed to relieve them! Meaning: not only is the cost _really_ high, the return of investment is likely to be pretty low in comparison, either.

alexhenrie:
As a compromise, would you be willing to at least add an option on that page of the setup wizard to preserve Git's default behavior (i.e. not configure `pull.rebase` or `pull.ff`)?

alexhenrie:
By the way, why do you think that most users would choose the merge option anyway? The choice between merge and rebase is entirely dependent on the project's workflow, and from what I've seen, most projects prefer cleanly formatted pull requests without merge commits.

dscho:
> As a compromise, would you be willing to at least add an option on that page of the setup wizard to preserve Git's default behavior (i.e. not configure `pull.rebase` or `pull.ff`)?

Sure. This cannot be the default, of course.

> why do you think that most users would choose the merge option anyway?

The current wording of the advice. It says:

```
You have divergent branches and need to specify how to reconcile them.
You can do so by running one of the following commands sometime before
your next pull:

  git config pull.rebase false  # merge (the default strategy)
  git config pull.rebase true   # rebase
  git config pull.ff only       # fast-forward only

You can replace \"git config\" with \"git config --global\" to set a default
preference for all repositories. You can also pass --rebase, --no-rebase,
or --ff-only on the command line to override the configured default per
invocation.
```

The very first suggested command is: `git config pull.rebase false  # merge (the default strategy)`. The well-studied [anchoring bias](https://en.wikipedia.org/wiki/Anchoring_%28cognitive_bias%29) naturally leads to this option being favored by the majority. This is exacerbated by the [default effect](https://en.wikipedia.org/wiki/Default_effect) where people tend to use the default option (and the advice prominently mentions that this is _the default strategy_) without giving the other options any thought.

_Especially_ when the users are annoyed by what they perceive as an unnecessary disruption to their work, or when users do not actually care which option Git should use, or both (which I highly suspect to be the case from my experience with users), it is highly likely that they just pick the first command and move on to what they _actually_ want to do.

alexhenrie:
Crud. That phrase, "the default strategy", was supposed to be dropped from the advisory message in Git 2.33.1. The fact that it is still there is an accident that needs to be fixed. Git intentionally does not recommend one strategy over another anymore.

dscho:
> Git intentionally does not recommend one strategy over another anymore.

Oh, I think that ship has sailed a long time ago. Everybody who started using Git more than a few months ago knows the default is "merge". And anybody learning from people who started using Git more than a few months ago will learn it from them.

I don't think it is honest to say that Git does not recommend one strategy over another anymore. That's just not reality.

dscho:
After thinking about this for a while, I don't think Git for Windows should be (ab-)used as a vehicle to push users away from the default they have gotten used to for such a long time.

alexhenrie:
Are you still OK with adding an _option_ to make Git for Windows behave the same as Git for Linux?

alexhenrie:
I wouldn't call it "pestering" when the warning gives the user situational awareness that fast-forwarding is impossible, and it makes more sense to set the behavior on a per-repository basis anyway (rather than set it globally) because different projects have different recommended workflows. But if you don't think that Git's default should be "fast-forward if possible, warn if not possible", please take it up on the mailing list rather than deciding for everyone that `git pull` should behave one way on Windows and a different way on Mac and Linux.

Another way to harmonize Git for Windows with Git on other platforms would be to set `pull.ff=only` by default during Git for Windows setup (which is already offered as an option), but then users wouldn't get any automatic help on how to choose between merge and rebase when fast-forwarding doesn't work.

dscho:
The problem I see is that Git itself, and all the Git workshops, and all the Git textbooks, for over a decade, said "a pull is simply a fetch followed by a merge". And then this was changed, and it was disruptive (and still is!), and the Git for Windows installer tried to bring back _some_ of the convenience by avoiding to "raise situational awareness" that is most often just perceived as a nuisance that keeps engineers from doing their job.

alexhenrie:
> The problem I see is that Git itself, and all the Git workshops, and all the Git textbooks, for over a decade, said "a pull is simply a fetch followed by a merge". And then this was changed, and it was disruptive (and still is!)

It's regrettable that Git had a bad default for so long, and you're right that it will take time for people to adjust. I was really hoping that Git for Windows would help send a clear message rather than continuing to muddy the water about what `git pull` does.

> and the Git for Windows installer tried to bring back _some_ of the convenience by avoiding to "raise situational awareness" that is most often just perceived as a nuisance that keeps engineers from doing their job.

Where is the evidence that it's "most often just perceived as a nuisance"? There were several long discussions on the mailing list before making the change, and the overwhelming consensus was that it is more annoying to accidentally merge than to be informed that fast-forwarding is impossible. See [this message](https://lore.kernel.org/git/20210705123209.1808663-1-felipe.contreras@gmail.com/) for a summary:

> Junio has suggested there's no consensus over fast-forward by default,
> but that's not true. Here is a non-exhaustive list of people that have
> agreed the default should be fast-forward only.
> 
>  * Linus Torvalds
>  * Junio C Hamano
>  * Jeff King
>  * Jonathan Nieder
>  * Richard Hansen
>  * Philip Oakley
>  * Elijah Newren
>  * John Keeping
>  * Ramkumar Ramachandra
>  * Alex Henrie
>  * W. Trevor King
>  * Greg Troxel
>  * Peter Kjellerstedt
>  * Konstantin Tokarev
>  * Robert Dailey
>  * Vít Ondruch
>  * Raymond E. Pasco
>  * Jacob Keller
>  * Felipe Contreras
> 
> There's only one person against the change: Matthieu Moy. His argument
> back then [13] was: why ask users to merge or rebase if newcomers will
> not pick a rebase? To that my response is: for the same reason we ask
> them to do `git commit --all`: so that they eventually learn about the
> staging area. Just like they can do `git commit --all`, they can do
> `git pull --merge`. They can figure out what that means later.

Obviously `pull.ff=only` is not the best behavior for everyone; I actually set `pull.rebase=true` in my global config file and I'm sure you will continue to set `pull.rebase=false`. But what each of us individually prefers is irrelevant because Git for Windows should not be used as a vehicle to undermine the community consensus.

Anyway, I doubt that we're going to resolve anything by going back and forth here. I suggest that you send a message to the Git mailing list to hear what the rest of the community has to say.

dscho:
> it will take time for people to adjust.

If my experience with users (thousands of them in person, even, over the course of the past 16 years) is any indication, you are not doing yourself any favors by believing that your goal is attainable. People won't adjust, unless they _have_ to. And in this instance, the costs outweigh the benefits, in my opinion.

Sure, people familiar with Git's internals (including the Git maintainer) might get talked into agreeing that "merge" was an unfortunate default for "pull". However, it is eminently easier to explain, and the Law of Least Surprises is violated by "rebase" making formerly reachable commits unreachable. So, from a usability point of view, it looks as if this entire mission to turn a formerly clear concept into a choice achieves only one thing: to make Git even harder to use than before.

Having said that, if I see proper evidence (with A/B testing on real users, not with the few people who contribute patches to Git itself who feel strongly enough to participate in a discussion about this) that I am wrong, I will change my stance on this. It won't be easy, though, and it will cost real money instead of just words and opinions from a few Git experts who might have forgotten how hard it was to learn how to use Git in the first place.

PR #394, opened 2021-11-18
Prepare for Git Credential Manager Core losing the "Core" and having a new home

dscho:
The Git Credential Manager project moved into [its own org](https://github.com/GitCredentialManager/), and while at it, dropped the `Core` suffix.

Let's prepare for that with this Pull Request.

Note: we can only merge this (and test it thoroughly) once a new release is out because the assets of all existing releases are still called `gcmcore-*`, but future releases will call them `gcm-*`.


ldennington:
Oof. I may have been a bit premature with the commit to update package/setup exe names. That's going to be a broader change that I think @mjcheetham wants to happen at a later time. I will go ahead and remove that commit from my series.

ldennington:
> Oof. I may have been a bit premature with the [commit to update package/setup exe names](https://github.com/GitCredentialManager/git-credential-manager/pull/541/commits/db4c9eddfb5c90ad840b93062ebc466ac4d89625). That's going to be a broader change that I think @mjcheetham wants to happen at a later time. I will go ahead and remove that commit from my series.

Note: This doesn't mean you should close this PR! Just that I was getting a bit ahead of myself 😬 .

mjcheetham:
Yeah this is probably premature. We've dropped the `Core` suffix from the naming, but the executables are still the same.
That change will require more thought and coordination between our two projects 😀

rimrul:
The first commit might cause issues when updatin ancient git sdk installations via `pacman`, but that's not a huge issue.

dscho:
>  That change will require more thought and coordination between our two projects 😀

I hoped that https://github.com/git-for-windows/build-extra/pull/394/commits/7f6509614e292208bef2be64d82810864bcd4e1c would provide all we need here: a shell script that ensures that obsolete configs will still work, but with a warning that suggests to adjust the obsolete config.

dscho:
> 
> 
> Yeah this is probably premature. We've dropped the `Core` suffix from the naming, but the executables are still the same. That change will require more thought and coordination between our two projects 😀
> 
> The only change to respond to right now is probably in the installer UI, and the repo URL.

After giving this much thought, I am not really a fan of separating the project rename from the executable rename and from the archive rename (i.e. the name of the `gcmcore-<platform>-*.zip` files that are attached to Git Credential Manager's releases).

Such a rename is pretty disruptive, so if we already disrupt users, I'd prefer to do it only once.

What do you think, @mjcheetham?

dscho:
I split out https://github.com/git-for-windows/build-extra/pull/396 and https://github.com/git-for-windows/build-extra/pull/397 from this PR. Only the parts remain here that eagerly anticipate the change of the executable name (to `git-credential-manager.exe`) and of the archive name (to `gcm-win-x86-<version>.zip`).

mjcheetham (approved):

PR #398, opened 2021-11-22
installer: PrivilegesRequired set to admin

dbaileyut:
* "none" is undocumented:
  https://jrsoftware.org/ishelp/index.php?topic=setup_privilegesrequired
* "admin" implements the desired behavior of requiring a machine-wide
  install. See https://github.com/git-for-windows/git/issues/3550

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -82,7 +82,7 @@ InfoBeforeFile={#SourcePath}\..\gpl-2.0.rtf
 #ifdef OUTPUT_TO_TEMP
 PrivilegesRequired=lowest
 #else
-PrivilegesRequired=none
+PrivilegesRequired=admin
 #endif
 UninstallDisplayName={#APP_NAME}
 UninstallDisplayIcon={app}\{#MINGW_BITNESS}\share\git\git-for-windows.ico


dscho:
> Signed-off-by: Doug Bailey <7095040+dbaileyut@users.noreply.github.com>

Would you mind signing off with your real email address? If there is any issue with the patch, we will need to have a way to contact you, after all, and the "noreply" email won't allow that.

installer/install.iss
```diff
@@ -82,7 +82,7 @@ InfoBeforeFile={#SourcePath}\..\gpl-2.0.rtf
 #ifdef OUTPUT_TO_TEMP
 PrivilegesRequired=lowest
 #else
-PrivilegesRequired=none
+PrivilegesRequired=admin
```
dscho, line 85: The commit message gives enough context to understand what this does, but I am missing an explanation as to _why_ we want to do that. Like, why not leave it with the undocumented behavior?

Looking through the commit history, I see that 1f3e1f7e360ca997f877cece0c0eb5cb1927a198 followed the advice to replace `none` with `lowest`. And 23672af723da18e5bc3c679e52106de3c2dec55a reverted that choice. In particular the latter's commit message makes me dubious whether we considered all the consequences of changing to `admin`.

dscho: Also: do we want https://jrsoftware.org/ishelp/topic_setup_usepreviousprivileges.htm? And what about https://jrsoftware.org/ishelp/topic_setup_privilegesrequiredoverridesallowed.htm?

Again, I am not opposed to merge this PR, I just need to make sure that the change is sound, i.e. well-researched. We cannot unleash a broken installer onto potentially millions of users.

dbaileyut:
This does not implement the desired behavior per https://github.com/git-for-windows/git/issues/3550#issuecomment-975933677

dscho:
> This does not implement the desired behavior per [git-for-windows/git#3550 (comment)](https://github.com/git-for-windows/git/issues/3550#issuecomment-975933677)

Well, can't we figure out a way to implement it, using all those new options that are now available (and weren't before)?

dbaileyut:
We can - I just thought that would be a new PR. Happy to test that out and reopen if that's the best course of action

dscho:
> We can - I just thought that would be a new PR. Happy to test that out and reopen if that's the best course of action

I would not mind at all reopening this PR with a force-pushed update, and an updated title and description.

dscho:
> [dbaileyut](/dbaileyut) closed this [5 minutes ago](#event-5656824645)

Oh, feel free to reopen it even before force-pushing an update ;-)

dbaileyut:
Reopening with changes

dbaileyut:
Force pushed changes to my branch. Reopen and comment is greyed out for me so I assume a maintainer must re-open it now.

FYI, the choice for system-wide vs user-wide looks like this:
image

dscho:
> Force pushed changes to my branch. Reopen and comment is greyed out for me so I assume a maintainer must re-open it now.

Uh oh. I fear not even a maintainer can re-open now...

image


> FYI, the choice for system-wide vs user-wide looks like this: [...]

That looks good, if a bit annoying. I would have hoped that we could have a mode where admins would always install for all, just like before. You have no idea how many complaints I get about the many, many, _many!!!_ options the installer shows, and adding another choice before even starting the installer is going to make the situation even worse...

dbaileyut:
Here are the options I see beyond the status quo:

1. `PrivilegesRequired=none` and `PrivilegesRequiredOverridesAllowed=commandline`. This preserves the existing behavior when launched normally (along with the upgrade error I reported). However, it does add the /ALLUSERS and /CURRENTUSER command line options to explicitly choose. There doesn't appear to be any downsides compared to the status quo
2. `PrivilegesRequired=admin` and `PrivilegesRequiredOverridesAllowed=dialog`. Seems to do the right thing in all situations but adds an extra, slightly ugly, dialog up front.
3. `PrivilegesRequired=admin` and`PrivilegesRequiredOverridesAllowed=commandline` - this would invoke UAC prompt and system-wide install no matter what with GUI installs but there would be a command-line switch for user-wide installation
4. Generate separate machine-wide and user-wide installer packages like Visual Studio Code does
  image


Doing at least option 1 seems like a no brainer. I feel like 2 is the best option but maybe 4 is worth considering. I think 3 is too disruptive to folks who want to install user-wide.

Thoughts before I open a new PR?

dscho:
> Thoughts before I open a new PR?

Several.

1. Thank you for this excellent summary!
2. My preference would be for 3 if it wasn't for that disruption, as you say.
3. If we can come up with a behavior that we would like to have but that InnoSetup does not offer, we could consider patching it, [it is Open Source](https://github.com/jrsoftware/issrc).

PR #403, opened 2022-02-01
Astextplain updates

rimrul:
git-for-windows/git#3673 has brought some issues in `astextplain` to my attetntion.

This PR fixes most of them, except for the fairly pointless `RTF` case arm, that's been a `TODO` since 2010. We could add some actual text conversion to that with `unrtf`, like the original author intended, but that would be a whole new package. Alternatively, we could drop the case arm and the associated gitattributes.


dscho (approved):
Thank you! The changes all look good.

I called `git rebase -ix "cd git-extra && updpkgsums && git commit --amend --no-edit PKGBUILD" main` and force-pushed, to fix the PR build.

PR #404, opened 2022-02-01
ci: avoid a bogus error when no packages/artifacts need to be built

dscho:
Avoid [a bogus error](https://github.com/git-for-windows/build-extra/actions/runs/1780156306):

> Error when evaluating 'strategy' for job 'build-artifacts'. .github/workflows/main.yml (Line: 63, Col: 20): Matrix vector 'directory' does not contain any values

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
             comm -12 "$2" "$3" >"$4" &&
             cat "$4" &&
             sed -e 's/\/$//' -e 's/.*/"&", /' -e '$s/, $//' <"$4" >list.txt &&
-            echo "::set-output name=$1::[$(tr -d '\n' <list.txt)]"
+            echo "::set-output name=$1::[$(test ! -s list.txt && echo '""' || tr -d '\n' <list.txt)]"
           }
 
           git ls-files \*/PKGBUILD | sed 's|[^/]*$||' | sort >directories.txt &&
@@ -32,6 +32,7 @@ jobs:
   build-packages:
     needs: determine-packages
     runs-on: windows-latest
+    if: needs.determine-packages.outputs.matrix != '[""]'
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


dscho:
@rimrul since you also worked in the space, could I ask you to review this PR?

.github/workflows/main.yml
```diff
@@ -13,24 +13,26 @@ jobs:
...
-          sed -e 's/\/$//' -e 's/.*/"&", /' -e '$s/, $//' <artifacts.txt >list.txt &&
-          echo "::set-output name=artifacts::[$(tr -d '\n' <list.txt)]"
+          define_matrix artifacts releaseable.txt touched.txt artifacts.txt
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
       artifacts: ${{ steps.set-matrix.outputs.artifacts }}
   build-packages:
     needs: determine-packages
     runs-on: windows-latest
+    if: needs.determine-packages.outputs.matrix != '["empty"]'
```
rimrul, line 35: Do we need a string in our array to explicitly mark it as empty or could we just check for an empty array (`[]`)?

dscho: If we use the empty array, then the matrix jobs will _still_ fail to even execute. I changed it to the empty string, and clarified in the commit message:

> Note: we need to add a bogus string because otherwise the parser that wants to construct the matrix jobs will _still_ fail.

dscho: See here for the error: https://github.com/git-for-windows/build-extra/actions/runs/1780285695, and note that it uses the empty array in the workflow: https://github.com/git-for-windows/build-extra/actions/runs/1780285695/workflow

dscho: Since the logs will be deleted in a couple months, here is a [link to the `main.yml` revision](https://github.com/git-for-windows/build-extra/blob/db2e236e66dbe84c9ba0faf66075119faf3fe4e2/.github/workflows/main.yml) and a screenshot for posterity:

image

dscho: @rimrul does that address your concern?

rimrul: Yes.

PR #411, opened 2022-04-06
Warn about end of Windows Vista support

dscho:
This moves up the Bash v5.x warning and starts warning about the imminent end of support for Windows Vista, as well as about the end of Windows 7/8 support early next year.


dscho:
I know, I know, we planned on upgrading to Bash v5.x already, but this late in the release cycle I need to feel less than confident about sneaking in the upgrade. Let's do it right after v2.36.0.

rimrul (approved):
We don't need to rush the bash update.

PR #413, opened 2022-04-29
fix `wprintf` format string for `w_char*`

miimou:
In ISO C, specifier for wide character strings is `%ls`.
As `proxy-lookup.c` is compiled with GCC of mingw-w64 which complies the standard,
this change uses `%ls` instead of `%s`.
See https://github.com/git-for-windows/git/issues/3818


miimou:
Thank you for your correction.
bd4962e worked in my environment.

PR #414, opened 2022-05-02
aliases.sh: allow pipe to work

temsa:
After investigating for a while why [this issue, by which I have been affected recently, happened](https://github.com/pinojs/pino-pretty/issues/305), I found out that we could avoid this `pipe` problem.

this is how typically it materializes :
```
$ node -e 'console.log(666)'
666

$ node -e 'console.log(666)' | pino-pretty
stdout is not a tty

$ node.exe -e 'console.log(666)' | pino-pretty
666
```

We just need to use the [undocumented](https://github.com/rprichard/winpty/blob/7e59fe2d09adf0fa2aa606492e7ca98efbc5184e/src/unix-adapter/main.cc#L502) switch from `winpty` to allow for the pipe to another command to actually work without getting into `stdout is not a tty`. It also still provides the value of `winpty`, e.g. for running an interactive shell

This PR just adds the switch, which should allow both of those commands to work in the intended manner:
```
$ node.exe -e 'console.log(666)' | pino-pretty
666

$ node -e 'console.log(666)' | pino-pretty
666
```

### DIFF
--- a/git-extra/aliases.sh
+++ b/git-extra/aliases.sh
@@ -15,7 +15,9 @@ xterm*)
 		case "$(type -p "$name".exe 2>/dev/null)" in
 		''|/usr/bin/*) continue;;
 		esac
-		alias $name="winpty $name.exe"
+		# The -Xallow-non-tty undocumented switch allows
+		# for pipe '|' to work as expected with those commands
+		alias $name="winpty -Xallow-non-tty $name.exe"
 	done
 	;;
 esac


dscho:
Interesting! This switch was introduced in https://github.com/rprichard/winpty/commit/222ecb9f4404cce3cdbafa0a97c7c3da4ce2b3c2. The commit message suggests that those options might be pulled at any point, but then, they have been in `winpty` since 2016...

dscho:
Seems that the `-X` options are still under consideration: https://github.com/rprichard/winpty/issues/103. Might be worth reviving that discussion, as well as https://github.com/rprichard/winpty/issues/101.

temsa:
Moving instead to a wrapper script running `winpty` only when relevant could be an option too, like https://github.com/rprichard/winpty/issues/101#issuecomment-279883299, but I've also asked if [this option is here to stay](https://github.com/rprichard/winpty/issues/103#issuecomment-1115835613)

rimrul:
I don't think you'll get a response, active `winpty` development has seemingly stopped a few years ago.
A wrapper script is probably still the better solution, since it prevents running into the issue that is outlined in https://github.com/rprichard/winpty/issues/103#issuecomment-285987050, i.e. programs that don't produce coloured output when piped to a file producing colour sequences when piped to a file in Git Bash.

dscho:
> A wrapper script is probably still the better solution

Since we're talking about a Bash alias, we could do the same in the alias itself, too, right?

dscho:
> Since we're talking about a Bash alias, we could do the same in the alias itself, too, right?

Seems that I was wrong about aliases: they cannot contain if-then-else blocks. But shell _functions_ can:

```console
$ unalias node

$ function node () { if test -t 0 -a -t 1; then winpty.exe node.exe "$@"; else node.exe "$@"; fi; }

$ node
Welcome to Node.js v16.15.0.
Type ".help" for more information.
> console.log("123")
123
undefined
> process.exit(0)

$ node -e 'console.log("123")' | cat
123
```

dscho:
@temsa how about trying your hand at patching [`aliases.sh`](https://github.com/git-for-windows/build-extra/blob/HEAD/git-extra/aliases.sh) accordingly?

dscho:
Let's just close this PR, it's multi-year stale.

PR #415, opened 2022-05-03
winpty: auto-detect redirects of stdin/stdout

rimrul:
Running commands through `winpty` with a redirected `stdin` or
`stdout` only leads to error messages, that are at best not very
helpful to users and at worst outright confusing users.
We can do better by adding a thin wrapper script that detects
when it is and isn't appropriate to run a command through `winpty`
and acts accordingly.

This is an alternative to #414 that doesn't use undocumented options and shouldn't cause commands to think they're outputting to a console when they aren't.

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -55,7 +55,8 @@ source=('inputrc'
         'git-credential-helper-selector.rc'
         'git-askpass.c'
         'git-askpass.h'
-        'git-askpass.rc')
+        'git-askpass.rc'
+        'winpty')
 sha256sums=('8ed76d1cb069ac8568f21c431f5e23caebea502d932ab4cdff71396f4f0d5b72'
             'e36a3b93b8a33b0a74619f2449ed6d56ed54e4e2938b97070bce4926f1f44054'
             '640d04d2a2da709419188a986d5e5550ad30df23c7ea9be1a389c37a6b917994'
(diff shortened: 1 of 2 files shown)


dscho:
That's an interesting idea!

However, I would like to see this in `winpty.exe` directly, rather than requiring a shell script wrapper (these things really bring a performance hit with them, on Windows).

I'd be totally willing to carry a patch in `MSYS2-packages/winpty/` and build our own versions of the `winpty` package, but I could also imagine that the MSYS2 project would be quite keen on that.

What I have in mind is a new option that _opts out_ of this new handling, so that by default we're checking `isatty(0) && isatty(1)` and if either one is redirected, simply spawn the child process with inherited handles (so that we do not have to pipe `stdin`/`stdout`/`stderr` back and forth) and skip the rest of `winpty`.

What do you think @rimrul?

rimrul:
> these things really bring a performance hit with them, on Windows

I've assumed the performance impact to be negligible when executing a shell script in bash, but you're probably right, this is probably a whole new bash proccess, just for this.

> I could also imagine that the MSYS2 project would be quite keen on that.

That  definitely sounds plausible.

>What do you think @rimrul?

I'll see what I can do.

dscho:
> > What do you think @rimrul?
> 
> I'll see what I can do.

Thank you!

PR #416, opened 2022-05-12
Preserve /etc/ssh/ssh_config between installations

Quenty:
Currently any user configuration of /etc/ssh/ssh_config is lost whenever the user installs a new copy of Git. This is undesirable.

What occurs is the following:

1. ssh_config is included in the [Files] list for install.iss
2. The uninstall executable for the previuos version of Git is invoked before install.
3. The uninstall executable removes all files in the [Files] section, including /etc/ssh_config

We can follow the same pattern that gitconfig follows, which is to back up the current config to the tmp file system, and then
restore it upon install. This code implements this behavior.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -2679,6 +2679,15 @@ begin
         end else if FileExists(UninstallAppPath+'\etc\gitconfig') and
             (not FileCopy(UninstallAppPath+'\etc\gitconfig',ExpandConstant('{tmp}\gitconfig.system'),True)) then
             LogError('Could not save system config; continuing anyway');
+
+        // Save a copy of the system ssh config so that we can copy it back later
+        if FileExists(UninstallAppPath+'\{#MINGW_BITNESS}\etc\ssh\ssh_config') then begin
+            if (not FileCopy(UninstallAppPath+'\{#MINGW_BITNESS}\etc\ssh\ssh_config',ExpandConstant('{tmp}\ssh_config.system'),True)) then
+                LogError('Could not save system ssh config; continuing anyway');
+        // Save a copy of the system ssh config so that we can copy it back later
+        end else if FileExists(UninstallAppPath+'\etc\ssh\ssh_config') and
+            (not FileCopy(UninstallAppPath+'\etc\ssh\ssh_config',ExpandConstant('{tmp}\ssh_config.system'),True)) then
+            LogError('Could not save system ssh config; continuing anyway');
     end;
 
     if UninstallString<>'' then begin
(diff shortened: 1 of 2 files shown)


installer/install.iss
```diff
@@ -2679,6 +2679,15 @@ begin
         end else if FileExists(UninstallAppPath+'\etc\gitconfig') and
             (not FileCopy(UninstallAppPath+'\etc\gitconfig',ExpandConstant('{tmp}\gitconfig.system'),True)) then
             LogError('Could not save system config; continuing anyway');
+
+        // Save a copy of the system ssh config so that we can copy it back later
+        if FileExists(UninstallAppPath+'\{#MINGW_BITNESS}\etc\ssh\ssh_config') then begin
```
rimrul, line 2684: Why would there be a MingW OpenSSH?

installer/release.sh
```diff
@@ -326,12 +326,14 @@ test -z "$arm64_artifacts_directory" || {
 die "Could not include ARM64 artifacts"
 
 etc_gitconfig_dir="${etc_gitconfig%/gitconfig}"
+etc_sshconfig_dir="${${etc_gitconfig%/ssh%/ssh_config}"
```
rimrul, line 329: I'm pretty sure this doesn't work. The double `${` with a single `}` cant be correct. I also don't think you can just concatenate these `%` patterns. And even if both of that would work, what are you trying to do? Strip a path that ends in `/etc/gitconfig` of a trailing `/ssh/ssh_config`?

rimrul (commented):
Wy would there be user configuration in /etc/ssh/ssh_config? Shouldn't that go into `~/.ssh/config`?

dscho:
There is a fundamental problem with this approach: what if Git for Windows _needs_ to change that `ssh_config`, e.g. to (temporarily) override choices made by the OpenSSH project?

For example, at some stage there was a problem with `ssh` where Azure Repos would not support _any_ of the new key formats, and therefore we had to re-enable a newly-considered-unsafe format.

With this patch, we would have no way to help our users back on track because they would be stuck with their version of `ssh_config` that simply does not contain the necessary workaround.

I do strongly agree with @rimrul that `~/.ssh/config` is the way to configure your local SSH in a way that persists across Git for Windows' updates.

dscho:
I haven't heard any convincing argument that would address the fundamental issue I raised: the presented approach would break upgrades that _need_ to modify the SSH config. And we need this not to be broken.

So I'll close this ticket. We can always revisit this should a splendid idea arise that _does_ address said issue.

PR #421, opened 2022-06-01
git-extra: disable CBC ciphers in ssh_config again

dscho:
We had to re-enable them in b46fba6f4 (git-extra: re-enable some SSHv1 protocols, 2017-10-12) after an OpenSSH upgrade disabled them.

The reason why we had to re-enable those ciphers is gone now, though: Azure Repos understand newer (safer) ciphers.


dscho:
The [build](https://dev.azure.com/Git-for-Windows/git/_build/results?buildId=100336&view=results) is running.

PR #425, opened 2022-06-15
Link PCRE2 changelog in release notes

rimrul:
The release notes for v2.37.0-rc0 link to [api.github.com](https://api.github.com/repos/PhilipHazel/pcre2/releases/latest), that's not designed for human consumption. We can do better.


dscho (approved):
Thank you for not only noticing but also fixing the issue! Great work!

PR #426, opened 2022-06-27
installer/portable: accommodate for Tig now linking to PCRE2

rimrul:
As discussed in https://github.com/git-for-windows/MSYS2-packages/pull/55#issuecomment-1156092406 we experimented with a Tig build that uses PCRE2 instead of PCRE. Since this experiment came out with a mere 0.1MB increase of the installer, we're going forward with that switch.

### DIFF
--- a/make-file-list.sh
+++ b/make-file-list.sh
@@ -125,7 +125,7 @@ for req in mingw-w64-$ARCH-git-credential-manager $SH_FOR_REBASE \
 		mingw-w64-$ARCH-connect git-flow unzip docx2txt \
 		mingw-w64-$ARCH-antiword mingw-w64-$ARCH-odt2txt \
 		mingw-w64-$ARCH-xpdf-tools ssh-pageant mingw-w64-$ARCH-git-lfs \
-		tig nano perl-JSON $GIT_UPDATE_EXTRA_PACKAGES)
+		tig nano perl-JSON libpcre2_8 libpcre2posix $GIT_UPDATE_EXTRA_PACKAGES)
 do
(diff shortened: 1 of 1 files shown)


dscho:
@rimrul this actually does not look too bad:

image

About this PR: I wonder whether there is anything we can do to transmogrify it into a `workflow_dispatch`-based workflow that would help similar scenarios in the future?

dscho:
> @rimrul [this](https://github.com/git-for-windows/build-extra/actions/runs/2577795266#artifacts) actually does not look too bad:

Note: I have _*not*_ tested `tig` in the resulting installer (or portable Git). It might simply not work, due to missing DLLs...

rimrul:
> Note: I have __not__ tested `tig` in the resulting installer (or portable Git). It might simply not work, due to missing DLLs...

I've given it a quick test, it seems to be working. It does start, the UI doesn't show any signs of breakage, regex search works.

dscho:
@rimrul since I merged https://github.com/git-for-windows/MSYS2-packages/pull/55, the `check-for-missing-dlls` builds are now failing: [`git-sdk-32` run](https://github.com/git-for-windows/git-sdk-32/runs/7280596797?check_suite_focus=true) & [`git-sdk-64` run](https://github.com/git-for-windows/git-sdk-64/runs/7280618334?check_suite_focus=true).

Therefore, we must adjust `make-file-list.sh` accordingly.

Since you already did this as part of this here PR, I repurposed it to no longer experiment with including the PCRE2 version of Tig, but to accommodate for the already-included PCRE2 version of Tig instead.

rimrul:
Looks good. I would've opened a second PR, but this works.

PR #429, opened 2022-07-10
git-extra: link executables as terminal server aware

rimrul:
With Windows 2000, Microsoft introduced a flag to the PE header to mark executables as
"terminal server aware". Windows terminal servers provide a redirected Windows directory and
redirected registry hives when launching legacy applications without this flag set. Since we
do not use any INI files in the Windows directory and don't write to the registry, we don't
need  this additional preparation. Telling the OS that we don't need this should provide
slightly improved startup times in terminal server environments.

This partially addresses https://github.com/git-for-windows/git/issues/3935


dscho:
Hmm. [Seems as if there is a new problem](https://github.com/git-for-windows/build-extra/runs/7271652328?check_suite_focus=true#step:4:138): apparently `WhoUses.o` is not being built, but it is needed for `WhoUses.exe`. @rimrul would you happen to have time to investigate? It would help me a lot.

dscho (commented):
Apart from the build failure, the change looks good to me, but I haven't really completely wrapped my head around the implications of that `tsaware` flag yet.

rimrul:
> would you happen to have time to investigate? It would help me a lot.

I'll take a look.

rimrul:
Ok, I've got it to build again. I had messed up some of the makefile syntax.

dscho:
The updated `git-extra` package [is being built](https://dev.azure.com/Git-for-Windows/git/_build/results?buildId=103351&view=results) as we speak.

dscho:
> The updated `git-extra` package [is being built](https://dev.azure.com/Git-for-Windows/git/_build/results?buildId=103351&view=results) as we speak.

The build was successful. Before I synchronize `git-sdk-32`/`git-sdk-64`, I want to finalize [the Perl upgrade to v5.36](https://github.com/git-for-windows/git/issues/3874#issuecomment-1196601966).

dscho:
The newest [snapshot](https://wingit.blob.core.windows.net/files/index.html) has these changes.

PR #431, opened 2022-07-25
Release Notes: copy-edit the Bash/Vista/Windows 7/Windows 8 bits

dscho:
For a while now, we announced that Bash will be upgraded. I just did that, the next snapshot will have that change.

Let's mention that in the release notes.

While at it, also drop Windows Vista support as announced earlier, and mention that Git for Windows will drop Windows 7/Windows 8 support soon as well.


jeffhostetler (approved):
I can't comment on the MinVersion change, but the rest looks good.

dscho:
> I can't comment on the MinVersion change, but the rest looks good.

@jeffhostetler That's all right, I am very confident about the `MinVersion` change ;-)

Thank you for the quick review!

PR #432, opened 2022-08-15
Only use 'antiword' in 'astextplain' if 'file' identified a Word document

Vampire:
Without this change you always get output that a given file is not a Word document, for example during pickaxeing.


git-extra/astextplain
```diff
@@ -12,7 +12,14 @@ case "$1" in
 		odt2txt "$1" || cat "$1"
 		;;
 	*.doc | *.DOC | *.dot | *.DOT)
-		out=$(antiword -m UTF-8 "$1") && sed "s/\^M$//" <<<$out || cat "$1"
+		case "$(file --brief --mime-type "$1")" in
```
dscho, line 15: My first reaction was to be concerned about the time spent on spawning `file`, but [`file` has hard enough of a time to identify all the Word Document flavors](https://github.com/search?q=repo%3Afile%2Ffile%20msword&type=code) that the alternative (namely, to hand-roll a script to identify Word Documents) seems prohibitive.

My next reaction was to be concerned about MinGit, but then I looked and found out that MinGit does not ship with `astextplain` (nor with `file`, for that matter).

dscho (approved):

dscho:
@Vampire thank you for your contribution!

S3od11:
'

S3od11 (commented):
'

S3od11:
'

git-extra/astextplain
S3od11: '

S3od11:
'

MayraPeralesContrera:
Sensores

PR #435, opened 2022-09-17
mingw-w64-git-cv2pdb: add ARM64 support

dennisameling:
Needs https://github.com/rainers/cv2pdb/pull/81 to add ARM64 support (merged ✅)

Tested with `MINGW_ARCH=clangarm64 makepkg-mingw -i` and things work as expected:

```
$ cv2pdb --help
Convert DMD CodeView/DWARF debug information to PDB files, Version 0.52
Copyright (c) 2009-2012 by Rainer Schuetze, All Rights Reserved

License for redistribution is given by the Artistic License 2.0
see file LICENSE for further details

usage: --help [-D<version>|-C|-n|-e|-s<C>|-p<embedded-pdb>] <exe-file> [new-exe-file] [pdb-file]
```

### DIFF
--- a/mingw-w64-cv2pdb/PKGBUILD
+++ b/mingw-w64-cv2pdb/PKGBUILD
@@ -3,7 +3,7 @@
 _realname=cv2pdb
 pkgbase="mingw-w64-${_realname}"
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=0.50
+pkgver=0.51.8.g9286b93
 pkgrel=1
 pkgdesc="Converter of DWARF debug information to PDB files (mingw-w64)"
 arch=('any')
@@ -24,6 +24,7 @@ die () {
 case "$CARCH" in
 i686) WARCH=Win32; OUTDIR=bin/Release;;
 x86_64) WARCH=x64; OUTDIR=bin/Release_x64;;
+aarch64) WARCH=ARM64; OUTDIR=bin/Release_ARM64;;
 *) die "Unsupported architecture: $CARCH"
 esac
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,11 +56,22 @@ jobs:
       fail-fast: false
       matrix:
         directory: ${{ fromJSON(needs.determine-packages.outputs.matrix) }}
+    env:
+      HOME: "${{github.workspace}}\\home"
     steps:
       - uses: actions/checkout@v3
       - uses: git-for-windows/setup-git-for-windows-sdk@v1
         with:
           flavor: full
+      - name: configure build
+        shell: bash
+        run: |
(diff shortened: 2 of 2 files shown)


dennisameling:
@dscho my [upstream PR](https://github.com/rainers/cv2pdb/pull/81) has been merged. Note that this PR bumps the version to the [latest commit](https://github.com/rainers/cv2pdb/commit/9286b93a6535a7ac3228922b547a7ae08379cedd) of `cv2pdb`, so you might want to do some checks if the `x64` version also keeps working as it should.

mingw-w64-cv2pdb/PKGBUILD
```diff
@@ -3,7 +3,7 @@
 _realname=cv2pdb
 pkgbase="mingw-w64-${_realname}"
 pkgname=("${MINGW_PACKAGE_PREFIX}-${_realname}")
-pkgver=0.50
+pkgver=0.51.8.g9286b93
```
dscho, line 6: I would prefer to wait for a full release than to upgrade to an interim version.

dscho (commented):
I am not _totally_ opposed to this PR, but I wonder why we bother with `cv2pdb` on ARM64 because we use clang to compile for ARM64, right? And clang [optionally generates `.pdb` itself](http://blog.llvm.org/2017/08/llvm-on-windows-now-supports-pdb-debug.html)... So all we would need to do is to ask clang to generate it, instead of building without `.pdb`s and then generating `.pdb` files using cv2pdb after the fact...

dscho:
> PR build / build-packages (mingw-w64-cv2pdb) (pull_request) Failing after 8m 

Hmm. [This](https://github.com/git-for-windows/build-extra/actions/runs/3081494109/jobs/4980101951#step:4:35) looks [as if the build wants to call `git tag`](https://github.com/git-for-windows/build-extra/blob/367b471b47d83e36c27eac99124a1a553894195e/mingw-w64-cv2pdb/PKGBUILD#L43) and is missing a configured user.

@dennisameling would you mind cherry-picking 1c34c0d55b3949a81ea2d6e582d0b4de43a915ce to see whether that fixes the build?

dennisameling:
> @dennisameling would you mind cherry-picking https://github.com/git-for-windows/build-extra/commit/1c34c0d55b3949a81ea2d6e582d0b4de43a915ce to see whether that fixes the build?

Thanks! That resulted in:

```
D:\a\_temp\817b228c-038e-4e8b-930c-a2540b812488.sh: line 7: syntax error: unexpected end of file
Error: Process completed with exit code 2.
```

You might want to remove the `&&` at the end of the Bash script 😊

dennisameling:
This is no longer needed as I'm now using Clang's built-in functionality to generate PDBs in https://github.com/git-for-windows/MINGW-packages/pull/56. Thanks for the suggestion!

PR #436, opened 2022-09-18
mingw-w64-git-lfs: add ARM64 support

dennisameling:
I added ARM64 support to `git-lfs` [a few months ago](https://github.com/git-lfs/git-lfs/pulls?q=is%3Apr+author%3Adennisameling), so let's use it as part of Git for Windows' journey to ARM64 support.

Tested with `MINGW_ARCH=clangarm64 makepkg-mingw -i` and things work as expected:

```
$ git-lfs --version
git-lfs/3.2.0 (GitHub; windows arm64; go 1.18.2)
```


dscho (approved):
Awesome work!

This will most likely require quite a bit of follow-up work, as Git for Windows does not have any `mingw/aarch64` Pacman repository yet. Obviously, this PR is a hard prerequisite for that follow-up work.

PR #437, opened 2022-09-24
portable + installer: add CLANGARM64 support

dennisameling:
~Needs https://github.com/git-for-windows/MINGW-packages/pull/56~ ✅ merged
~Needs https://github.com/git-for-windows/MINGW-packages/pull/59~ ✅ merged
~Needs https://github.com/git-for-windows/MINGW-packages/pull/58~ ✅ merged
~Needs https://github.com/git-for-windows/setup-git-for-windows-sdk/pull/397~ ✅ merged
~Needs https://github.com/git-for-windows/git/pull/3916~ ✅ merged
~Needs https://github.com/msys2/MINGW-packages/pull/13787~ ✅ merged
~Needs https://github.com/git-for-windows/MINGW-packages/pull/57~ ✅ merged
~Needs https://github.com/git-for-windows/build-extra/pull/440~ ✅ merged

## How to test

On an ARM64 device, do the following:

- Install the `git-sdk-64`
- Make sure you create the `clangarm64` folder and start the SDK with `MSYSTEM=clangarm64`, as well as install tools like the Clang compiler. Note that the Git SDK GitHub Action will do that for you soon - see https://github.com/git-for-windows/setup-git-for-windows-sdk/pull/397/files for all details, including a list of the dependencies you need.
- Download this ZIP file below with the Git artifacts: [git-artifacts-zst.zip](https://github.com/git-for-windows/build-extra/files/9896613/git-artifacts-zst.zip) (created with https://github.com/git-for-windows/MINGW-packages/pull/56)
- Download this ZIP file with the `curl`/`openssl`/`xpdf-tools`/`wintoast` dependencies for `clangarm64`: [git-build-extra-artifacts-zst.zip](https://github.com/git-for-windows/build-extra/files/9896791/git-build-extra-artifacts-zst.zip)

Then run the following command for **portable**:

```
/usr/src/build-extra/please.sh make_installers_from_mingw_w64_git --version=2.39.0 --portable -o artifacts \
--pkg=/mingw-w64-clang-aarch64-git-2.39.0.1.de9501c66f-1-any.pkg.tar.zst \
--pkg=/mingw-w64-clang-aarch64-git-doc-html-2.39.0.1.de9501c66f-1-any.pkg.tar.zst \
--pkg=/mingw-w64-clang-aarch64-xpdf-tools-4.00-1-any.pkg.tar.zst \
--pkg=/mingw-w64-clang-aarch64-curl-7.86.0-1-any.pkg.tar.zst \
--pkg=/mingw-w64-clang-aarch64-openssl-1.1.1.s-1-any.pkg.tar.zst \
--pkg=/mingw-w64-clang-aarch64-wintoast-1.0.0.181.9b0663d-1-any.pkg.tar.zst
```

... or this command for **installer**:

```
/usr/src/build-extra/please.sh make_installers_from_mingw_w64_git --version=2.39.0 --installer -o artifacts \
--pkg=/mingw-w64-clang-aarch64-git-2.39.0.1.de9501c66f-1-any.pkg.tar.zst \
--pkg=/mingw-w64-clang-aarch64-git-doc-html-2.39.0.1.de9501c66f-1-any.pkg.tar.zst \
--pkg=/mingw-w64-clang-aarch64-xpdf-tools-4.00-1-any.pkg.tar.zst \
--pkg=/mingw-w64-clang-aarch64-curl-7.86.0-1-any.pkg.tar.zst \
--pkg=/mingw-w64-clang-aarch64-openssl-1.1.1.s-1-any.pkg.tar.zst \
--pkg=/mingw-w64-clang-aarch64-wintoast-1.0.0.181.9b0663d-1-any.pkg.tar.zst
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,7 +112,8 @@ jobs:
             build-installers &&
           cygpath -aw "$PWD/sdk-artifact/usr/bin/core_perl" >>$GITHUB_PATH &&
           cygpath -aw "$PWD/sdk-artifact/usr/bin" >>$GITHUB_PATH &&
-          cygpath -aw "$PWD/sdk-artifact/mingw64/bin" >>$GITHUB_PATH
+          cygpath -aw "$PWD/sdk-artifact/mingw64/bin" >>$GITHUB_PATH &&
+          echo "MSYSTEM=MINGW64" >>$GITHUB_ENV
       - name: build ${{ matrix.directory }}/
         shell: bash
         run: |
@@ -166,7 +167,8 @@ jobs:
           esac &&
           cygpath -aw "$PWD/sdk-artifact/usr/bin/core_perl" >>$GITHUB_PATH &&
           cygpath -aw "$PWD/sdk-artifact/usr/bin" >>$GITHUB_PATH &&
-          cygpath -aw "$PWD/sdk-artifact/mingw${{ matrix.bitness }}/bin" >>$GITHUB_PATH
+          cygpath -aw "$PWD/sdk-artifact/mingw${{ matrix.bitness }}/bin" >>$GITHUB_PATH &&
+          echo "MSYSTEM=MINGW${{ matrix.bitness }}" >>$GITHUB_ENV
       - name: build installer
         if: matrix.artifact == 'build-installers'
         shell: bash
@@ -202,7 +204,8 @@ jobs:
             build-installers &&
           cygpath -aw "$PWD/sdk-artifact/usr/bin/core_perl" >>$GITHUB_PATH &&
           cygpath -aw "$PWD/sdk-artifact/usr/bin" >>$GITHUB_PATH &&
-          cygpath -aw "$PWD/sdk-artifact/mingw${{ matrix.bitness }}/bin" >>$GITHUB_PATH
+          cygpath -aw "$PWD/sdk-artifact/mingw${{ matrix.bitness }}/bin" >>$GITHUB_PATH &&
+          echo "MSYSTEM=MINGW${{ matrix.bitness }}" >>$GITHUB_ENV
       - name: check for missing DLLs
         shell: bash
         run: ./check-for-missing-dlls.sh
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -4,8 +4,8 @@
 
 #include "config.iss"
 
-#if !defined(APP_VERSION) || !defined(BITNESS)
-#error "config.iss should define APP_VERSION and BITNESS"
+#if !defined(APP_VERSION) || !defined(BITNESS) || !defined(MINGW_BITNESS)
+#error "config.iss should define APP_VERSION, BITNESS and MINGW_BITNESS"
 #endif
 
 #define APP_NAME      'Git'
@@ -13,7 +13,6 @@
 #undef APP_VERSION
 #define APP_VERSION   'Snapshot'
 #endif
-#define MINGW_BITNESS 'mingw'+BITNESS
 #define APP_CONTACT_URL 'https://github.com/git-for-windows/git/wiki/Contact'
 #define APP_URL       'https://gitforwindows.org/'
 #define APP_BUILTINS  'share\git\builtins.txt'
--- a/installer/release.sh
+++ b/installer/release.sh
@@ -111,19 +111,20 @@ case "$displayver" in
 esac
 
 # Evaluate architecture
-ARCH="$(uname -m)"
-
-case "$ARCH" in
-i686)
+case "$MSYSTEM" in
+MINGW32)
 	BITNESS=32
+	ARCH=i686
 	;;
-x86_64)
+MINGW64)
 	BITNESS=64
+	ARCH=x86_64
 	;;
 *)
-	die "Unhandled architecture: $ARCH"
+	die "Unhandled MSYSTEM: $MSYSTEM"
 	;;
 esac
+MSYSTEM_LOWER=${MSYSTEM,,}
 
 echo "Generating release notes to be included in the installer ..."
 ../render-release-notes.sh --css usr/share/git/ ||
@@ -152,7 +153,7 @@ then
 	LIST=
 else
 	echo "Generating file list to be included in the installer ..."
-	LIST="$(ARCH=$ARCH BITNESS=$BITNESS \
+	LIST="$(ARCH=$ARCH \
 		ETC_GITCONFIG="$etc_gitconfig" \
 		PACKAGE_VERSIONS_FILE=package-versions.txt \
 		INCLUDE_GIT_UPDATE=1 \
@@ -168,12 +169,12 @@ test -z "$cmd_git" || {
 die "Could not execute 'git version'"
 
 printf '; List of files\n%s\n%s\n%s\n%s\n%s\n%s\n' \
-	"Source: \"mingw$BITNESS\\bin\\blocked-file-util.exe\"; Flags: dontcopy" \
+	"Source: \"$MSYSTEM_LOWER\\bin\\blocked-file-util.exe\"; Flags: dontcopy" \
 	"Source: \"{#SourcePath}\\package-versions.txt\"; DestDir: {app}\\etc; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore" \
 	"Source: \"{#SourcePath}\\..\\ReleaseNotes.css\"; DestDir: {app}\\usr\\share\\git; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore" \
 	"Source: \"cmd\\git.exe\"; DestDir: {app}\\bin; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore" \
-	"Source: \"mingw$BITNESS\\share\\git\\compat-bash.exe\"; DestName: bash.exe; DestDir: {app}\\bin; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore" \
-	"Source: \"mingw$BITNESS\\share\\git\\compat-bash.exe\"; DestName: sh.exe; DestDir: {app}\\bin; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore" \
+	"Source: \"$MSYSTEM_LOWER\\share\\git\\compat-bash.exe\"; DestName: bash.exe; DestDir: {app}\\bin; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore" \
+	"Source: \"$MSYSTEM_LOWER\\share\\git\\compat-bash.exe\"; DestName: sh.exe; DestDir: {app}\\bin; Flags: replacesameversion restartreplace; AfterInstall: DeleteFromVirtualStore" \
 	"Source: \"{#SourcePath}\\..\\post-install.bat\"; DestName: post-install.bat; DestDir: {app}; Flags: replacesameversion restartreplace" \
 >file-list.iss ||
 die "Could not write to file-list.iss"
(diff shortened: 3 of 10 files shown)


rimrul:
> error: target not found: mingw-w64-clang-aarch64-antiword

This one will probably compile fine for clangarm64, msys2 seems to compile it already with clang and ubuntu/debian seem to compile it for arm64 with no arch specific patches.

> error: target not found: mingw-w64-clang-aarch64-odt2txt

This one will probably need some changes regarding [this patch](https://github.com/msys2/MINGW-packages/blob/c710adb682d0a22c4c51a462f977168511a52910/mingw-w64-odt2txt/0001-Compile-for-Msys2.patch#L35)
and  it's [counterpart in PKGBUILD](https://github.com/MSYS2/MINGW-packages/blob/master/mingw-w64-odt2txt/PKGBUILD#L28-L29).

> error: target not found: mingw-w64-clang-aarch64-xpdf-tools

xpdf-**tools**, that's from https://github.com/git-for-windows/MINGW-packages/blob/main/mingw-w64-xpdf, we'd need our own `CLANGARM64` repo for that.

dennisameling:
@rimrul I was able to get all the deps to work.
- `connect, antiword, odt2txt`: https://github.com/msys2/MINGW-packages/pull/13787
- `xpdf-tools`: the download URL of that package seems to be broken. After fixing it, I was able to build the package correctly locally: https://github.com/git-for-windows/MINGW-packages/pull/57

Will continue work on this PR soon. It's currently in a "very rough but sort of working" state 😅

dennisameling (commented):
@dscho @rimrul This is now ready for a first round of reviews. I was able to [generate working portable + installer artifacts](https://github.com/dennisameling/git/releases/tag/v2.38.1.windows.99) for 2.38.1 with this setup 🎉 

Curious to hear your thoughts on my general approach here. Especially `make-file-list.sh` was a bit tricky, but everything is looking good in my testing so far.

Thanks in advance!

portable/release.sh
```diff
@@ -40,23 +37,33 @@ die "Usage: $0 [--output=<directory>] <version> [optional components]"
...
+	ARCH=aarch64
+	ARTIFACT_SUFFIX=ARM64
 	MD_ARG=256M
 	;;
 *)
-	die "Unhandled architecture: $ARCH"
+	die "Unhandled MSYSTEM: $MSYSTEM"
 	;;
 esac
+MSYSTEM_LOWER=${MSYSTEM,,}
```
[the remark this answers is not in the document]
dennisameling: Added the environment variable in https://github.com/git-for-windows/build-extra/pull/437/commits/17a3a4a6d54979c669289517d49b3c1ee70c4359

make-file-list.sh
```diff
@@ -144,62 +161,62 @@ test -z "$required" || {
...
-	-e '^/mingw../share/doc/gettext/' \
-	-e '^/mingw../share/doc/lib' \
-	-e '^/mingw../share/doc/pcre2\?/' \
-	-e '^/mingw../share/doc/git-doc/.*\.txt$' \
-	-e '^/mingw../share/doc/zstd/' \
-	-e '^/mingw../lib/gettext/' -e '^/mingw../share/gettext/' \
-	-e '^/usr/include/' -e '^/mingw../include/' \
+	-e '^/(mingw|clang(arm)?)../bin/.*zstd\.exe$' \
+	-e '^/(mingw|clang(arm)?)../share/doc/openssl/' \
+	-e '^/(mingw|clang(arm)?)../share/doc/gettext/' \
```
dennisameling, line 187: @dscho after some testing, I found that this capturing group (or the non-capturing group alternative: `'^/(?:mingw|clang(arm)?)../share/doc/gettext/'`) only works when `-P` is specified in `grep`:

```
Regexp selection and interpretation:
  -E, --extended-regexp     PATTERN is an extended regular expression (ERE)
  -F, --fixed-strings       PATTERN is a set of newline-separated strings
  -G, --basic-regexp        PATTERN is a basic regular expression (BRE)
  -P, --perl-regexp         PATTERN is a Perl regular expression
  -e, --regexp=PATTERN      use PATTERN for matching
  -f, --file=FILE           obtain PATTERN from FILE
  -i, --ignore-case         ignore case distinctions
  -w, --word-regexp         force PATTERN to match only whole words
  -x, --line-regexp         force PATTERN to match only whole lines
  -z, --null-data           a data line ends in 0 byte, not newline
```

However, when I tested with `-P`, I ran into the following error:

```
grep: the -P option only supports a single pattern
```

🤦🏼‍♂️ this is slightly getting out of hand.

Would it also be ok for you to simply add the `MSYSTEM_LOWER` variable to the pattern, like this:

```
-e "^/$MSYSTEM_LOWER/share/doc/gettext/" \
```

... or do you really need `mingw..` so that it matches both `mingw32` and `mingw64`? Just checking!

dscho: I _guess_ that `$MSYSTEM_LOWER` would work. _Except_ in the case of `clangarm64` where we expect to also have to handle `mingw64`... But why not simply `-e "^/\(mingw\|clang\)[^/]*/share/doc/gettext/"`, i.e. match any top-level directory starting either with `mingw` or with `clang`?

dennisameling:
@dscho This is ready for review. The `build-installers 32` CI job is failing [due to a failed PGP check](https://github.com/git-for-windows/build-extra/actions/runs/3625271670/jobs/6113132825), but all the other ones have succeeded 👍🏼

installer/install.iss
```diff
@@ -34,6 +33,7 @@
 ; Compiler-related
 Compression=lzma2/ultra64
 LZMAUseSeparateProcess=yes
+LZMANumBlockThreads=2
```
rimrul, line 36: I'm not sure if this actually gets close to doubling the speed on GitHub hosted runners.

The docs say:

> Since LZMA2 (and LZMA) uses two threads for match-finding by default (see [CompressionThreads](https://jrsoftware.org/ishelp/topic_setup_compressionthreads.htm)), there ideally should be two processor cores available for each block thread. Thus, to see the maximum benefit from a value of 2, four cores are needed.

dennisameling: You're right. When building locally, this was super helpful as I could set it to `LZMANumBlockThreads=2` (my machine has 8 cores), but there likely won't be any performance benefit in CI. Will remove this commit.

portable/release.sh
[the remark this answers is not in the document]
dscho: The `MSYSTEM` variable is defined either in `/etc/profile` or in `git-bash.exe`, depending on the exact scenario, or in the case of CI runs [by the `setup-git-for-windows-sdk` Action](https://github.com/dscho/setup-git-for-windows-sdk/blob/0b1f2cf212dffda75a306481d91866af92278ac7/main.ts#L76). It should not be necessary to set it yourself unless it is set incorrectly.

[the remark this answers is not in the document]
dennisameling: Here's an example of when it failed without me explicitly setting the `MSYSTEM` in CI: https://github.com/git-for-windows/build-extra/actions/runs/3357168162/jobs/5562796433. The `build-artifacts` CI job doesn't use `setup-git-for-windows-sdk` so that explains why it failed. If you want, I can set `MSYSTEM: mingw64` only to the `build-artifacts` CI job.

installer/release.sh
```diff
@@ -111,19 +104,25 @@ case "$displayver" in
 esac
 
 # Evaluate architecture
-ARCH="$(uname -m)"
-
-case "$ARCH" in
-i686)
```
dscho, line 117: I would prefer the change from `uname -m` to `$MSYSTEM` to be done in a separate commit that still only handles i686 and x86_64, and add the aarch64 part in a separate commit.

Please let me know if you want assistance with that.

portable/release.sh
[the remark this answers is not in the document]
dscho: No, setting `MSYSTEM` should be scoped to the particular SDK instance. In other words, [here](https://github.com/git-for-windows/build-extra/blob/fe12b23f586d34cad55c855f7edce0fcbb40c37f/.github/workflows/main.yml#L113), [here](https://github.com/git-for-windows/build-extra/blob/fe12b23f586d34cad55c855f7edce0fcbb40c37f/.github/workflows/main.yml#L167) and [here](https://github.com/git-for-windows/build-extra/blob/fe12b23f586d34cad55c855f7edce0fcbb40c37f/.github/workflows/main.yml#L203), we need to insert a line like this (in a separate, preparatory commit):

```sh
          echo "MSYSTEM=MINGW${{matrix.bitness}}" >>$GITHUB_ENV &&
```

(The first instance is not inside a matrix, it needs to hard-code `MINGW64`.)

make-file-list.sh
```diff
@@ -372,11 +389,14 @@ etc/post-install/01-devices.post
 etc/post-install/03-mtab.post
 etc/post-install/06-windows-files.post
 usr/bin/start
-mingw$BITNESS/bin/pdftotext.exe
-mingw$BITNESS/bin/libstdc++-6.dll
+$MSYSTEM_LOWER/bin/pdftotext.exe
 usr/bin/column.exe
 EOF
 
+[[ -z "$MINIMAL_GIT" && $MSYSTEM_LOWER == mingw* ]] && cat <<EOF
```
dscho, line 396: The `[[ ... ]] &&` syntax is not used anywhere else in this file, therefore it threw me quite a bit. Let's not?

Also, maybe it's better to introduce a new variable like `PDFTOTEXT_FILES`, and set it in an architecture-specific way before using it here.

.github/workflows/main.yml
```diff
@@ -4,6 +4,7 @@ on: pull_request
 
 env:
   GIT_CONFIG_PARAMETERS: "'checkout.workers=56'"
+  MSYSTEM: MINGW64
```
dscho, line 7: Since we're not setting `MSYSTEM=MINGW32` explicitly in the 32-bit jobs, that's what causes [this problem](https://github.com/git-for-windows/build-extra/actions/runs/3625271670/jobs/6122096351#step:5:10).

make-file-list.sh
dennisameling: I used it so that I could do `mingw*` (wildcard). `test` can't do that. Do you have a better alternative to this: `if test -z "$MINIMAL_GIT" && (test "$MSYSTEM_LOWER" = "mingw64" || test "$MSYSTEM_LOWER" = "mingw32")`

I'm not sure whether I get the comment regarding `pdftotext`. That executable is supported on all platforms, so no need to introduce architecture-specific logic for it, right?

dscho: Right, the wildcard. The classical Unix method is to use `case`, like so:

```sh
case "$MINIMAL_GIT,$MSYSTEM_LOWER" in
,mingw*)
  ...
  ;;
esac
```

But that's a _lot_ more verbose, so I guess I'm fine with the `[[` construct.

> I'm not sure whether I get the comment regarding pdftotext.

What I mean is this: https://github.com/git-for-windows/build-extra/pull/437/files#diff-9b50f413340b32f6bc699741e8b09ab52b3e9c7698d787cc26c4c95e78580916L375-R401

```diff
-mingw$BITNESS/bin/pdftotext.exe
-mingw$BITNESS/bin/libstdc++-6.dll
+$MSYSTEM_LOWER/bin/pdftotext.exe
 usr/bin/column.exe
 EOF
 
+if test -z "$MINIMAL_GIT" && (test "$MSYSTEM_LOWER" = "mingw64" || test "$MSYSTEM_LOWER" = "mingw32"); then
+cat <<EOF
+$MSYSTEM_LOWER/bin/libstdc++-6.dll
+EOF
+fi
+
```

Separating the `pdftotext.exe` line from the `libstdc++-6.dll` line in the proposed way won't make it as obvious as the current version that `pdftotext.exe` is the only reason why that `.dll` is required in the first place.

What I would prefer is to have conditionals that add the appropriate lines for pdftotext depending on the platform, _unless_ we're enumerating the files for a MinGit.

So I would define these lines somewhat like this:

```
case $MSYSTEM_LOWER in
mingw*)
  PDFTOTEXT_FILES="$MSYSTEM_LOWER/bin/pdftotext.exe
$MSYSTEM_LOWER/bin/libstdc++-6.dll"
  ;;
*)
  # In the clang version, we do not need the libstdc++ DLL
  PDFTOTEXT_FILES="$MSYSTEM_LOWER/bin/pdftotext.exe
  ;;
esac
```

and then replace the current mentions of these files with `$PDFTOTEXT_FILES`.

dennisameling: Ah, I thought that `libstdc++-6.dll` was used for other things as well. Have just updated the commit and addressed the other feedback from this PR as well. Thanks!

dscho (approved):
Very nice, thank you _so much_!

PR #442, opened 2022-11-04
Set the config color.ui rather than individual configs

xim:
This brings color to the people without the complexities of the higher number of configs.

Relates to git-for-windows/git#4095

Signed-off-by: Morten Minde Neergaard <169057+xim@users.noreply.github.com>

### DIFF
--- a/git-extra/git-extra.install
+++ b/git-extra/git-extra.install
@@ -8,9 +8,7 @@ post_install () {
 	symlinks = false
 	autocrlf = true
 [color]
-	diff = auto
-	status = auto
-	branch = auto
+	ui = auto
 	interactive = true
 [pack]
 	packSizeLimit = 2g
--- a/git-extra/gitconfig
+++ b/git-extra/gitconfig
@@ -2,9 +2,7 @@
 	symlinks = false
 	autocrlf = true
 [color]
-	diff = auto
-	status = auto
-	branch = auto
+	ui = auto
 	interactive = true
 [pack]
 	packSizeLimit = 2g


dscho:
While I see that this does what it is supposed to do, I cannot help but wonder what else this change does.

For example, I seem to remember vaguely that `color.ui` was not respected in _all_ code paths where `color.diff` was respected, or some similar issue.

Also, this is technically a change in behavior that might negatively affect existing setups. What can we do to build confidence that this is not the case?

xim:
> While I see that this does what it is supposed to do, I cannot help but wonder what else this change does.
> 
> For example, I seem to remember vaguely that `color.ui` was not respected in _all_ code paths where `color.diff` was respected, or some similar issue.

There could always be bugs, of course, but from my reading of the source code it looks like `color.ui` should be respected quite universally.

> Also, this is technically a change in behavior that might negatively affect existing setups. What can we do to build confidence that this is not the case?

For one, the change appears to only affect PortableGit, which might make things easier to reason about. The cases I can think about where this might cause changes would be where the user has specified `color.ui = never` or `color.ui = always` in their config or on the command line, and that would still result in colored output in most cases before.

Of close, it's always possible there might be behaviour changes I'm not thinking of. Feel like this move goes in a positive direction however, as I it reduces complexity and possibilities for confusion.

dscho:
Okay, let's take it, then, and hope that any regressions are caught before [Git v2.39.0 is due](https://tinyurl.com/gitcal).

dscho:
The `git-extra` package [was built successfully](https://dev.azure.com/Git-for-Windows/git/_build/results?buildId=108799&view=results). Therefore, the next [snapshot](https://wingit.blob.core.windows.net/files/index.html) will have it.

xim:
Great! I'll give it a spin then!

xim:
The change doesn't appear to be present in the snapshots. Apologies for not spotting it sooner, still using my custom-built version.

dscho:
> The change doesn't appear to be present in the snapshots. Apologies for not spotting it sooner, still using my custom-built version.

@xim thank you for testing! And you're absolutely correct, the issue is not fixed in the snapshots at all.

The reason is that [this PR only addresses the issue if `/etc/gitconfig` did not exist yet](https://github.com/git-for-windows/build-extra/blob/38a184751fadd816b85a7f50b78fa2f008063ea5/git-extra/git-extra.install.in#L5-L6). But when git-for-windows/git-sdk-64@b6f79f4b0ac1835a123819d5f03ae9d357f4bfc5 upgraded `git-extra`, [the `/etc/gitconfig` file existed already](https://github.com/git-for-windows/git-sdk-64/blob/b6f79f4b0ac1835a123819d5f03ae9d357f4bfc5%5E/etc/gitconfig).

PR #444, opened 2022-11-17
pacman-helper quick_add: stop relying on a single fallback architecture

dscho:
I plan on using this to deploy both i686 and x86_64 flavors of the heimdal package over at https://github.com/git-for-windows/MSYS2-packages/pull/63.


dscho:
There is only the attempt left to build a `pacman` flavor, but that seems more complex than is worth the benefit. So I'll close the PR.

PR #446, opened 2022-11-25
Automate adding release notes

dscho:
Implementing this as a relatively easy-to-use workflow is step 1. At some stage, I want to end up with a solution where a member of the Git for Windows team can simply add a comment of the form `/relnote <type> <message>` to a PR and have this workflow triggered automatically.

[This](https://github.com/dscho/build-extra/actions/runs/3548059031) is the workflow that added the release note about MinTTY, and [this](https://github.com/dscho/build-extra/actions/runs/3548082900) is the one that talks about Bash v5.2 patchlevel 12.

### DIFF
--- a/.github/workflows/add-release-note.yml
+++ b/.github/workflows/add-release-note.yml
@@ -1,0 +1,47 @@
+name: Add Release Note
+run-name: 'Add Release Note (${{ inputs.type }}): ${{ inputs.message }}'
+
+on:
+  workflow_dispatch:
+    inputs:
+      type:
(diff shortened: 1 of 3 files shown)


dscho:
And [this run](https://github.com/dscho/build-extra/actions/runs/3548104710) updated the description [to use the better link again](https://github.com/git-for-windows/build-extra/pull/446/commits/bce678e6add2ce279822c6d3c770ee6934f14d01) that I had totally forgotten about.

PR #447, opened 2022-11-28
git-extra: set the config color.ui also in existing /etc/gitconfig

dscho:
This is a follow-up for #442 (which did not fix our problem completely).


dscho:
Superseded by https://github.com/git-for-windows/build-extra/pull/448.

PR #448, opened 2022-11-28
git-extra: set the config color.ui also in existing /etc/gitconfig

dscho:
This is a follow-up for https://github.com/git-for-windows/build-extra/pull/442 (which did not fix our problem completely).

It supersedes https://github.com/git-for-windows/build-extra/pull/447, using a branch in the same repository so that we can branch-deploy.


dscho:
image

Yes!!! This is the first successful branch-deployment using the new workflow in https://github.com/git-for-windows/git-for-windows-automation.

PR #452, opened 2022-12-17
`mingit` + `archive`: add CLANGARM64 support

dennisameling:
Some final bits before I can set up a full pipeline for creating ARM64 artifacts! 🎉


please.sh
```diff
@@ -4775,20 +4775,24 @@ find_mspdb_dll () { #
...
 
-build_mingw_w64_git () { # [--only-32-bit] [--only-64-bit] [--skip-test-artifacts] [--skip-doc-man] [--skip-doc-html] [--force] [<revision>]
+build_mingw_w64_git () { # [--only-i686] [--only-x86_64] [--only-aarch64] [--skip-test-artifacts] [--skip-doc-man] [--skip-doc-html] [--force] [<revision>]
 	output_path=
 	sed_makepkg_e=
 	force=
 	src_pkg=
 	while case "$1" in
-	--only-32-bit)
+	--only-i686|--only-32-bit)
```
dennisameling, line 4784: I left `--only-32-bit` in place to not break existing CI pipelines and scripts. Ideally these should move to `--only-i686` too.

please.sh
```diff
@@ -4775,20 +4775,24 @@ find_mspdb_dll () { #
...
 		MINGW_ARCH=mingw32
 		export MINGW_ARCH
 		;;
-	--only-64-bit)
+	--only-x86_64|--only-64-bit)
```
dennisameling, line 4788: I left `--only-64-bit` in place to not break existing CI pipelines and scripts. Ideally these should move to `--only-x86_64` too.

installer/release.sh
```diff
@@ -116,7 +116,7 @@ MINGW64)
 CLANGARM64)
 	BITNESS=64
 	ARCH=aarch64
-	inno_defines="$inno_defines$LF#define INSTALLER_FILENAME_SUFFIX 'arm64'"
+	inno_defines="$inno_defines$LF#define INSTALLER_FILENAME_SUFFIX 'ARM64'"
```
dennisameling, line 119: `git-update-git-for-windows` [also expects](https://github.com/git-for-windows/build-extra/blob/f6041a722c0cd6a36d139537f1253aaae74800e9/git-extra/git-update-git-for-windows#L251) uppercase ARM64.

git-extra/PKGBUILD
```diff
@@ -5,7 +5,7 @@ _ver_base=1.1
 pkgver=1.1.614.d551cf0cc
 pkgrel=1
 pkgdesc="Git for Windows extra files"
-arch=('i686' 'x86_64')
+arch=('any')
```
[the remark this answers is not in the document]
dscho: > the `git-extra` package versions are hosted together with the MSYS packages

I just realized that this is a real problem. The reason is that we currently have this setup:

```mermaid
graph TD;
    A[git-sdk-64] -->|gets Pacman packages from| C[x86_64];
    A[git-sdk-64] -->|gets Pacman packages from| E["i686 (MINGW packages only)"];
    B[git-sdk-32] -->|gets Pacman packages from| F["x86_64 (MINGW packages only)"];
    B[git-sdk-32] -->|gets Pacman packages from| D[i686];
    C -->|hosted in| G[x86_64/ subdir of wingit];
    F -->|hosted in| G[x86_64/ subdir of wingit];
    D -->|hosted in| H[i686/ subdir of wingit];
    E -->|hosted in| H[i686/ subdir of wingit];
```

In other words, all of the i686-flavored packages are hosted in the same subdirectory of the `wingit` Pacman repository, and all the x86_64-flavored packages in another subdirectory, and each subdirectory has two package indices, one with the full list of packages and the other one with the MINGW-only packages.

Now, it's relatively easy to imagine ARM64 to come in this way:

```mermaid
graph TD;
    I[git-sdk-arm64] -->|gets Pacman packages from| J["aarch64 (CLANG packages only)"];
    I[git-sdk-arm64] -->|gets Pacman packages from| C[x86_64];
    A[git-sdk-64] -->|gets Pacman packages from| C[x86_64];
    A[git-sdk-64] -->|gets Pacman packages from| E["i686 (MINGW packages only)"];
    B[git-sdk-32] -->|gets Pacman packages from| F["x86_64 (MINGW packages only)"];
    B[git-sdk-32] -->|gets Pacman packages from| D[i686];
    J -->|hosted in| K[aarch64/ subdir of wingit];
    C -->|hosted in| G[x86_64/ subdir of wingit];
    F -->|hosted in| G[x86_64/ subdir of wingit];
    D -->|hosted in| H[i686/ subdir of wingit];
    E -->|hosted in| H[i686/ subdir of wingit];
```

But if we keep hosting `git-extra` in the MSYS parts of the Pacman repositories, then the `aarch64/` subdirectory will never get its own `git-extra` package!

Therefore, we will _have_ to embark on this rather involved package name change first :-(

installer/release.sh
dennisameling: > I'm not so sure about that. The `git-update-git-for-windows` script [only does this if /arm64/ exists](https://github.com/git-for-windows/build-extra/blob/f6041a722c0cd6a36d139537f1253aaae74800e9/git-extra/git-update-git-for-windows#L249), which we're getting away from.
> 
> So it's not much of a precedent/requirement.

I mentioned it because `$arch_bit` [is used to filter the release artifacts](https://github.com/git-for-windows/build-extra/blob/f6041a722c0cd6a36d139537f1253aaae74800e9/git-extra/git-update-git-for-windows#L278) later on in the script. So it will be looking for `Git-2.39.0-ARM64.exe`, for example, if that's the suffix of the installer executable.

> And if I look at the file names Git-2.39.0-arm64.exe or Git-2.39.0-ARM64.exe, I am not so sure that I prefer the latter.
> 
> @dennisameling do you have a strong opinion on this? I don't really have a strong opinion, but I want to "get it right".

I also don't prefer the latter, but (incorrectly) assumed that you preferred this based on some earlier conversations we had, which I can only vaguely remember. It looks like Microsoft themselves are also inconsistent and use [ARM64](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewARM64) (uppercase), [Arm](https://support.microsoft.com/en-us/windows/windows-arm-based-pcs-faq-477f51df-2e3b-f68f-31b0-06f5e4f8ebb5) and [Arm64](https://devblogs.microsoft.com/visualstudio/arm64-visual-studio/). Oh, well. 😅 

The only strong opinion I have is that I'd like to take this opportunity to start moving away from `$BITNESS` and use full architecture suffixes everywhere. I'm fully aware that this has implications all over the place, but I think it's definitely worth it, and we're already slowly moving towards it with things like [the recent changes to `make-file-list.sh`](https://github.com/git-for-windows/build-extra/commit/4cc13ceb60fb85f79d9a96e3c8ed0341bdf941c9).

While I think `Option 1` in the table below is the easiest to read, I guess `Option 2` is a bit easier to implement as it's consistent with the [architectures MSYS2 uses](https://www.msys2.org/docs/environments/). And I quite prefer staying as close as possible to MSYS2's naming and way of working.

Here's some examples of what consistency across the board could look like:

| Current   |      Option 1      |  Option 2 (MSYS2) | Option 3 (Golang) |
|----------|-------------|------|--------|
| Git-2.39.0-64-bit.exe | Git-2.39.0-x64.exe | Git-2.39.0-x86_64.exe | Git-2.39.0-amd64.exe |
| Git-2.39.0-32-bit.exe | Git-2.39.0-x86.exe | Git-2.39.0-i686.exe | Git-2.39.0-386.exe |
| n/a | Git-2.39.0-arm64.exe | Git-2.39.0-aarch64.exe | Git-2.39.0-arm64.exe | Git-2.39.0-arm64.exe |
| PortableGit-2.39.0-64-bit.7z.exe | PortableGit-2.39.0-x64.7z.exe | PortableGit-2.39.0-x86_64.7z.exe | PortableGit-2.39.0-amd64.7z.exe | 
| PortableGit-2.39.0-32-bit.7z.exe | PortableGit-2.39.0-x86.7z.exe | PortableGit-2.39.0-i686.7z.exe | PortableGit-2.39.0-386.7z.exe | 
| n/a | PortableGit-2.39.0-arm64.7z.exe | PortableGit-2.39.0-aarch64.7z.exe | PortableGit-2.39.0-arm64.7z.exe | 

What do you think?

rimrul: > > And if I look at the file names Git-2.39.0-arm64.exe or Git-2.39.0-ARM64.exe, I am not so sure that I prefer the latter.
> > 
> > @dennisameling do you have a strong opinion on this? I don't really have a strong opinion, but I want to "get it right".
> 
> I also don't prefer the latter, but (incorrectly) assumed that you preferred this based on some earlier conversations we had, which I can only vaguely remember. It looks like Microsoft themselves are also inconsistent and use [ARM64](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewARM64) (uppercase), [Arm](https://support.microsoft.com/en-us/windows/windows-arm-based-pcs-faq-477f51df-2e3b-f68f-31b0-06f5e4f8ebb5) and [Arm64](https://devblogs.microsoft.com/visualstudio/arm64-visual-studio/). Oh, well. 😅 

`ARM` is the canonical spelling of the company name, but I agree that lowercase looks better in filenames. Keeping the 64 makes sense to distiguish this from Microsofts earlier short lived 32bit ARM endeavours.

> The only strong opinion I have is that I'd like to take this opportunity to start moving away from `$BITNESS` and use full architecture suffixes everywhere. I'm fully aware that this has implications all over the place, but I think it's definitely worth it, and we're already slowly moving towards it with things like [the recent changes to `make-file-list.sh`](https://github.com/git-for-windows/build-extra/commit/4cc13ceb60fb85f79d9a96e3c8ed0341bdf941c9).

We'll definitely need to move away from `$BITNESS` in the long run. It made sense when `x86` and `AMD64` where the only options we had to care about, but it makes less sense with `ARM64` on the table and it'll completely stop making any sense when we at some point in the future drop `x86` support. 

> While I think `Option 1` in the table below is the easiest to read, I guess `Option 2` is a bit easier to implement as it's consistent with the [architectures MSYS2 uses](https://www.msys2.org/docs/environments/). And I quite prefer staying as close as possible to MSYS2's naming and way of working.

I like the approach of staying close to MSYS2, because it'll probably make our lives easier in the future.

> Here's some examples of what consistency across the board could look like:
> 
>  […]
> 
> What do you think?

If I could mix and match, I'd choose `x86` over `i686` and `386`, `amd64` or maybe `x86_64`  over `x64`, and have no real preference between `aarch64` and `arm64`.

git-extra/PKGBUILD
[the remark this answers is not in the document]
rimrul: I'm not sure I like switching `git-extra` to `any`. There are some parts of `git-extra` that are compiled natively. We could theoretically abuse the `x86` build as `any`, but that feels like a big clutch.

installer/release.sh
dscho: From the users' perspective, I _guess_ `x86` and `amd64` and `arm64` make most sense.

From the correctness' point of view, `i686`, `x86_64` and `aarch64` look the best, but I don't believe that this would serve our users well, but instead only the retentive type of people.

I do not have any strong opinion about what we end up with, though.

The thing I _do_ care about is that we announce this properly and leave at least a cycle or two, where we ship both `-<bitness>-bit` as well as the new artifacts (essentially just adding a copy under the old name), before we drop the old names.

We may want to go the whole nine yards and also rename `git-sdk-32` and `git-sdk-64`, once we decided on the new names.

git-extra/PKGBUILD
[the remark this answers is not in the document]
dscho: > I'm not sure I like switching `git-extra` to `any`. There are some parts of `git-extra` that are compiled natively. We could theoretically abuse the `x86` build as `any`, but that feels like a big clutch.

Right.

The correct thing to do really is to rename it to `mingw-w64-<arch>-git-extra` at the same time as using the `any` suffix. It'll be a ton of work, but I fear that we will have to do that first, before merging this here PR.

installer/release.sh
dennisameling: Sounds good. Let's go with `x86`, `amd64` and `arm64` then, or do you want to discuss with a broader audience first? If the proposed names look good to you, shall I start creating PRs in various places to:
- continue moving away from `$BITNESS` in favor of the full architectures
- update the release scripts to use the new names _and_ ensure that the binaries are shipped under the old _and_ the new names like you mentioned

Is [this](https://github.com/git-for-windows/build-extra/blob/3fad5d8f5fdf94b14715d5d7c9b915ebde05af0c/please.sh#L4030-L4153) `release` function in `please.sh` still the function you're using for putting all the binaries together prior to uploading them to GitHub? Where can I find the script that actually generates the GitHub release? Thanks!

git-extra/PKGBUILD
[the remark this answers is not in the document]
dennisameling: Let's do it 🔥 My productivity levels have been really high lately, so I'd love to keep up the pace. Apologies if this sometimes results in lower accuracy - I really appreciate all your extensive reviews and comments 👍🏼 

I think I found a solution here: https://github.com/git-for-windows/build-extra/pull/455

installer/release.sh
dscho: > Let's go with `x86`, `amd64` and `arm64` then

Sounds good to me!

> Is [this](https://github.com/git-for-windows/build-extra/blob/3fad5d8f5fdf94b14715d5d7c9b915ebde05af0c/please.sh#L4030-L4153) `release` function in `please.sh` still the function you're using for putting all the binaries together prior to uploading them to GitHub?

No, that was only for the local VM-based builds I used starting in 2015, before using automation to build official Git for Windows releases.

These days I am using [the Azure Pipeline build "Git artifacts"](https://dev.azure.com/git-for-windows/git/_build/index?definitionId=34&_a=completed) instead. The definition is unfortunately not really visible, but I ported it over into a GitHub workflow at https://github.com/git-for-windows/git/blob/main/.github/workflows/git-artifacts.yml, which is relatively close to the Azure Pipeline definition (that is not tracked in any Git directory but versioned independently).

The workflow does not yet have the logic to build official releases, though. The short version is that the Pipeline uses:

- `please.sh upgrade --no-build --only-mingw git` for official releases, and [an ad-hoc tagging for snapshots](https://github.com/git-for-windows/git/blob/e7d4c504802f93f5914ed48944ced4e593bcaf6c/.github/workflows/git-artifacts.yml#L66-L88) to generate the bundle artifacts,
- [`please.sh build-mingw-w64-git`](https://github.com/git-for-windows/git/blob/e7d4c504802f93f5914ed48944ced4e593bcaf6c/.github/workflows/git-artifacts.yml#LL202C38-L202C67) to build the actual Pacman packages, and
- [`please.sh make_installers_from_mingw_w64_git`](https://github.com/git-for-windows/git/blob/e7d4c504802f93f5914ed48944ced4e593bcaf6c/.github/workflows/git-artifacts.yml#L425) to create the actual artifacts.

I fear that there might be a lot of cruft in `please.sh` that I accumulated over the years that were long replaced by other methods and were kept only for the transitional period, but then the removal was readily forgotten.

dscho: @dennisameling FWIW my long-term plan was to move over from the Azure Pipeline to the GitHub workflow. With recent developments, I am inclined to accelerate that move. One thing we will definitely want to do is to move `git-artifacts.yml` into git-for-windows-automation, so that we can really restrict our ARM64 runners to that repository.

git-extra/PKGBUILD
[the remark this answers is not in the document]
dscho: > My productivity levels have been really high lately, so I'd love to keep up the pace.

Excellent! I just hope that my time off won't lower your productivity.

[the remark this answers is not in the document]
dennisameling: > Excellent! I just hope that my time off won't lower your productivity.

Haha, I'll be slowing down quite a bit too next week. But I'm super happy that we were able to get some major blockers out of the way before Christmas - that gives some peace of mind and motivation to pick things up further in the new year 😊 

I might still be creating some PRs, but will not expect you or anyone to review them before the new year. If that spams your inbox too much, let me know and I'll just create some branches without creating the actual PRs 👍🏼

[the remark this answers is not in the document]
dscho: Your PRs will always be welcome in my inbox. Even if I may let them sit there until after the holidays ;-)

installer/release.sh
dennisameling: >> Let's go with x86, amd64 and arm64 then
>
> Sounds good to me!

Here's a discussion to announce the change, as well as the implications on GfW's tooling: https://github.com/git-for-windows/git/discussions/4203

> @dennisameling FWIW my long-term plan was to move over from the Azure Pipeline to the GitHub workflow. With recent developments, I am inclined to accelerate that move. One thing we will definitely want to do is to move `git-artifacts.yml` into git-for-windows-automation, so that we can really restrict our ARM64 runners to that repository.

Sounds good to me. Will create a PR.

dennisameling:
This should be good to go now. `git-extra` keeps complaining about its version, though. Bumped the version already but it doesn't seem to help 🤷🏼‍♂️

rimrul:
Looks like both bump commits missed the hash of `HEAD` and instead picked the hash of `HEAD~1`.

dscho (approved):

PR #453, opened 2022-12-17
Issue 4172: handle processes names ending with .exe

jeanmichel-nwsb:
Signed-off-by: Jean-Michel Mercier [jeanmichel.mercier@newsbridge.io](mailto:jeanmichel.mercier@newsbridge.io

This PR fix the detection of running executable as on current version of Windows, the Process.Name may ends with ` .exe` which make the comprison fails, and the processes are not marked to be terminated.

### DIFF
--- a/installer/modules.inc.iss
+++ b/installer/modules.inc.iss
@@ -510,6 +510,7 @@ var
     AppList:array of RM_PROCESS_INFO;
     ReasonList:IdList;
     Success:DWORD;
+    ProcessName:String;
 begin
     SetArrayLength(Processes,0);
     Result:=0;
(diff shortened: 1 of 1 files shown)


dscho:
@jeanmichel-nwsb I edited the commit message and the patch a bit, could you please test to confirm that it still fixes the reported problem?

jeanmichel-nwsb:
Hi @dscho 
I've seen your change
Will test it later today
Thanks, much compact 👍

jeanmichel-nwsb:
Hi @dscho 
So I have tested and it works ... 
But just wanted to be sure with you if it is the inteded behavior:
- it first display the ProcessesPage, 
- and then if I confirm "Install", it offer to stop the process

From the code I was under the impression it would try to terminate the processes before displaying this page but I'm probably wrong with understanding the logic of InnoSetup pages

So for me all good 👍 
Thanks

image
image

dscho:
> But just wanted to be sure with you if it is the inteded behavior:
> 
>     * it first display the ProcessesPage,
> 
>     * and then if I confirm "Install", it offer to stop the process

Yes, it is intended. It says "will be terminated" to notify the user that this is going to happen instead of silently stopping the processes.

PR #456, opened 2022-12-22
Improve ARM64 support of the `sdk` function

dscho:
In Git for Windows' SDK, we [prominently advertise the `sdk` function](https://github.com/git-for-windows/build-extra/blob/53403f702e139577941baf9dda4f5308b20e09ce/git-extra/git-sdk.sh#L62-L63), but it does not _quite_ Do The Right Thing in `git-sdk-arm64` yet. This PR addresses that.


git-extra/git-sdk.sh
```diff
@@ -72,12 +72,27 @@ sdk () {
...
+		MINGW32) shortcut_suffix=" 32-bit";;
+		MINGW64) shortcut_suffix=" 64-bit";;
+		CLANGARM64) shortcut_suffix=" ARM64";;
+		MSYS)
+			case "$(uname -m)" in
+			i686) shortcut_suffix=" 32-bit";;
+			x86_64)
+				if test -d /clangarm64
+				then
+					shortcut_suffix=" ARM64"
```
dennisameling, line 85: Do we want to make this lowercase `arm64` as discussed [here](https://github.com/git-for-windows/build-extra/pull/452#discussion_r1053792396)?

dennisameling (approved):
LGTM - left one minor comment. Thanks!

git-extra/git-sdk.sh
dscho: That's a really good question. I am quite at a loss what we should do here. Maybe you can come up with sensible guidelines when to use upper-case "ARM64" and when to use lower-case "arm64" and when to use "clangarm64" and when to use "aarch64"?

dennisameling: Will create a discussion in `git-for-windows/git` for that later today 👍🏼

dennisameling: Slightly delayed, but here's the discussion: https://github.com/git-for-windows/git/discussions/4203

dennisameling: > That's a really good question. I am quite at a loss what we should do here. Maybe you can come up with sensible guidelines when to use upper-case "ARM64" and when to use lower-case "arm64" and when to use "clangarm64" and when to use "aarch64"?

I just added some guidelines to the [architecture discussion](https://github.com/git-for-windows/git/discussions/4203). It basically means that the artifact suffix becomes `arm64` (lowercase). How does that sound? 😊

dscho: > It basically means that the artifact suffix becomes `arm64` (lowercase). How does that sound? 😊

Sounds good! I did exactly that and force-pushed.

dscho:
> /deploy mingw-w64-git-extra

That didn't work because the directory is still named `git-extra`.

> /deploy git-extra

That worked!

PR #458, opened 2022-12-22
Prepare for uploading Pacman packages targeting Windows/ARM64

dscho:
This should be enough to get that part of the deployment working. The idea is that we will have two new package indices, [a full one](https://wingit.blob.core.windows.net/aarch64/git-for-windows.db) and [a MINGW-only one](https://wingit.blob.core.windows.net/aarch64/git-for-windows-aarch64.db). Until the time when Cygwin starts supporting Windows/ARM64, these will contain the same set of packages.

The idea is that we can then set up `/etc/pacman.conf` in `git-sdk-arm64` to contain these lines:

```ini
[git-for-windows]
# Once Cygwin starts supporting Windows/ARM64, we can switch to aarch64
Server = https://wingit.blob.core.windows.net/x86-64

[git-for-windows-aarch64]
Server = https://wingit.blob.core.windows.net/aarch64
```

### DIFF
--- a/pacman-helper.sh
+++ b/pacman-helper.sh
@@ -604,6 +604,11 @@ quick_add () { # <file>...
 		eval "mingw=\$${arch}_mingw"
 		test -n "$msys$mingw" || continue
 
+		case "$(test aarch64 = $arch && curl -sI "$(arch_url $arch)/git-for-windows.db")" in
+		*404*) initialize_fresh_pacman_repository=t;; # this one is new
+		*) initialize_fresh_pacman_repository=;;
+		esac
+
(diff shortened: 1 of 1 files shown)


dscho:
> Thank you! Though I don't really have a way to test this, the code looks good 👍🏼 let's test with one or two packages whether everything works as expected. Let me know if I can help with that!

@dennisameling ideally, I would like to see https://github.com/git-for-windows/git-for-windows-automation/pull/5 and https://github.com/git-for-windows/git-for-windows-automation/pull/4 merged, and then build a `git-extra` using that.

PR #457, opened 2022-12-22
please.sh: drop a lot of functionality that is no longer used

dscho:
This is in response to [a comment that asked whether `release` is still in use](https://github.com/git-for-windows/build-extra/pull/452#discussion_r1053792396).

PR #461, opened 2022-12-27
Start using new git-extra prefix

dennisameling:
This depends on #457 being merged first, as it removes a lot of code that we'd otherwise have to update.

I used the list of files that need updating from [this comment](https://github.com/git-for-windows/git-for-windows-automation/pull/5#issuecomment-1362580886) (thanks @dscho!)

### DIFF
--- a/commit-msys2.sh
+++ b/commit-msys2.sh
@@ -91,7 +91,7 @@ summarize_commit () {
 		git show "$1" --format=%H \
 			-M15 --raw -- var/lib/pacman/local/\*/desc
 	fi |
-	sed -ne '/.* M\tvar\/lib\/pacman\/local\/git-extra-[1-9].*\/desc$/d' \
+	sed -ne '/.* M\tvar\/lib\/pacman\/local\/\(mingw-w64-.*-\)\?git-extra-[1-9].*\/desc$/d' \
 	 -e '/ R[0-9]*\t/{s/-\([0-9]\)/ (\1/;h;s|-\([0-9][^/]*\)/desc$|\t\1)|;s|.*\t| -> |;x;s|/desc\t.*||;s|.*\t[^\t]*/||;G;s|\n||g;p}' \
 	 -e '/ A\t/{s|.*local/\([^/]*\)/desc|\1|;s|-\([0-9].*\)| (new: \1)|p}' \
 	 -e '/ D\t/{s|.*local/\([^/]*\)/desc|\1|;s|-\([0-9].*\)| (removed)|p}'
@@ -178,6 +178,7 @@ commit)
 	 if git diff-index --exit-code --cached HEAD -- \
 		':(exclude)var/lib/pacman/sync/' \
 		':(exclude)var/lib/pacman/local/git-extra-*/desc' \
+		':(exclude)var/lib/pacman/local/mingw-w64-*-git-extra-*/desc' \
 		':(exclude)etc/rebase.db*'
 	 then
 		# No changes, really, but maybe a new Pacman db
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -53,7 +53,7 @@ source=('inputrc'
         'zzz_rm_stackdumps.sh'
         'proxy-lookup.c'
         'git-askyesno.c'
-        'update-via-pacman.bat'
+        'update-via-pacman.bat.in'
         'git-credential-helper-selector.c'
         'git-credential-helper-selector.manifest'
         'git-credential-helper-selector.rc'
@@ -87,7 +87,7 @@ sha256sums=('8ed76d1cb069ac8568f21c431f5e23caebea502d932ab4cdff71396f4f0d5b72'
             '97e89689d91747ddb5ee873ae864aebcbb8d0364a52fa198db1e439ee2965b9b'
             '9371263a66afaee79d3b7362b2bd455af8faa860230491e21df7f9fbe25bbe9b'
             '0e3af7dacb665306076590ce9aec935929c198cec1bf9c5e161dfe285502ba67'
-            '1a37c34382f759ba4ed22df78dfe568d9508ec403e278f7941968854abb6d9e6'
+            'cb27fa8fd043ebe282ee96771de7e3d02af619542a60b54bddfabcd685c1ae0d'
             '32c834074aa725627f4b86db2aa3cb2ed1556cdc7b0256748ed4b700d1a61490'
             '247dc84535c89d2e00c6a741d4317cd76ae5958c91ac867e2e74d4e73bc3c998'
             '78307cd6c04a16240f68197e9155697bc75d5df8e4c8514ff67712bb93a1cd4a'
@@ -112,6 +112,8 @@ build() {
   export SRCDIR=$srcdir
   export BUILDDIR=build-${MINGW_CHOST}
   mkdir -p ${BUILDDIR}
+  sed -e "s|@@GIT_EXTRA_PACKAGE_NAME@@|${MINGW_PACKAGE_PREFIX}-${_realname}|g" \
+        <$startdir/update-via-pacman.bat.in >$BUILDDIR/update-via-pacman.bat
   make -f Makefile
 }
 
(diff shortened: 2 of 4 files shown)


dennisameling:
**Test results of `please.sh` and `commit-msys2.sh` changes**

I tested this locally by basically mimicking [sync.yml](https://github.com/git-for-windows/git-sdk-64/blob/main/.github/workflows/sync.yml):

```
sh -x usr/src/build-extra/please.sh sync --force --only-64-bit
sh -x /usr/src/build-extra/commit-msys2.sh commit
```

The resulting commit looks good to me:

```diff
$ git show 64498be630d1dc0003beefa306cdb28302f47d28
commit 64498be630d1dc0003beefa306cdb28302f47d28 (HEAD -> main)
Author: Dennis Ameling <dennis@dennisameling.com>
Date:   Tue Dec 27 21:54:32 2022 +0100

    Update 1 package

    git-extra (1.1.616.3f26b1b0c-1 -> 1.1.616.3f26b1b0c-1)

    Signed-off-by: Dennis Ameling <dennis@dennisameling.com>

diff --git a/var/lib/pacman/local/git-extra-1.1.616.3f26b1b0c-1/mtree b/var/lib/pacman/local/git-extra-1.1.616.3f26b1b0c-1/mtree
deleted file mode 100644
index 9fb27af6af..0000000000
Binary files a/var/lib/pacman/local/git-extra-1.1.616.3f26b1b0c-1/mtree and /dev/null differ
diff --git a/var/lib/pacman/local/git-extra-1.1.616.3f26b1b0c-1/desc b/var/lib/pacman/local/mingw-w64-x86_64-git-extra-1.1.616.3f26b1b0c-1/desc
similarity index 93%
rename from var/lib/pacman/local/git-extra-1.1.616.3f26b1b0c-1/desc
rename to var/lib/pacman/local/mingw-w64-x86_64-git-extra-1.1.616.3f26b1b0c-1/desc
index 9379cd8aea..f84650410c 100644
--- a/var/lib/pacman/local/git-extra-1.1.616.3f26b1b0c-1/desc
+++ b/var/lib/pacman/local/mingw-w64-x86_64-git-extra-1.1.616.3f26b1b0c-1/desc
@@ -1,5 +1,5 @@
 %NAME%
-git-extra
+mingw-w64-x86_64-git-extra

 %VERSION%
 1.1.616.3f26b1b0c-1
@@ -14,13 +14,13 @@ Git for Windows extra files
 https://github.com/git-for-windows/build-extra

 %ARCH%
-x86_64
+any

 %BUILDDATE%
 1671657907

 %INSTALLDATE%
-1672110153
+1672174026

 %PACKAGER%
 Johannes Schindelin <johannes.schindelin@gmx.de>
@@ -46,7 +46,10 @@ vim
 filesystem

 %CONFLICTS%
+git-extra
 mingw-w64-clang-aarch64-git-extra
-mingw-w64-x86_64-git-extra
 mingw-w64-i686-git-extra

+%PROVIDES%
+git-extra
+
diff --git a/var/lib/pacman/local/git-extra-1.1.616.3f26b1b0c-1/files b/var/lib/pacman/local/mingw-w64-x86_64-git-extra-1.1.616.3f26b1b0c-1/files
similarity index 100%
rename from var/lib/pacman/local/git-extra-1.1.616.3f26b1b0c-1/files
rename to var/lib/pacman/local/mingw-w64-x86_64-git-extra-1.1.616.3f26b1b0c-1/files
diff --git a/var/lib/pacman/local/git-extra-1.1.616.3f26b1b0c-1/install b/var/lib/pacman/local/mingw-w64-x86_64-git-extra-1.1.616.3f26b1b0c-1/install
similarity index 100%
rename from var/lib/pacman/local/git-extra-1.1.616.3f26b1b0c-1/install
rename to var/lib/pacman/local/mingw-w64-x86_64-git-extra-1.1.616.3f26b1b0c-1/install
diff --git a/var/lib/pacman/local/mingw-w64-x86_64-git-extra-1.1.616.3f26b1b0c-1/mtree b/var/lib/pacman/local/mingw-w64-x86_64-git-extra-1.1.616.3f26b1b0c-1/mtree
new file mode 100644
index 0000000000..c7d0d7279f
```

dennisameling:
**Test results of `update-via-pacman.bat` changes**

When the package was built, the appropriate part of `update-via-pacman.bat` was updated correctly as well: 

```
:FINISH
@ECHO Wrapping up: re-install mingw-w64-x86_64-git-extra
@pacman -S --noconfirm mingw-w64-x86_64-git-extra
@ECHO End of re-install of mingw-w64-x86_64-git-extra
@ECHO ==
@IF ERRORLEVEL 1 GOTO DIE
```

Here's the artifact in case you're interested: 
[git-extra-1.1.618.d98adb1b8-1-x86_64.pkg.tar.zip](https://github.com/git-for-windows/build-extra/files/10310804/git-extra-1.1.618.d98adb1b8-1-x86_64.pkg.tar.zip)

commit-msys2.sh
```diff
@@ -91,7 +91,7 @@ summarize_commit () {
 		git show "$1" --format=%H \
 			-M15 --raw -- var/lib/pacman/local/\*/desc
 	fi |
-	sed -ne '/.* M\tvar\/lib\/pacman\/local\/git-extra-[1-9].*\/desc$/d' \
+	sed -ne '/.* M\tvar\/lib\/pacman\/local\/mingw-w64-*-git-extra-[1-9].*\/desc$/d' \
```
dscho, line 94: I think we need a `.` before the `*` (it's a regular expression, not a shell pattern).

Also, maybe we should be lenient and still accept old package name _in addition_ to the new one?

```suggestion
	sed -ne '/.* M\tvar\/lib\/pacman\/local\/\(mingw-w64-.*-\)\?git-extra-[1-9].*\/desc$/d' \
```

commit-msys2.sh
```diff
@@ -170,7 +170,7 @@ commit)
 	 git add -A . &&
 	 if git diff-index --exit-code --cached HEAD -- \
 		':(exclude)var/lib/pacman/sync/' \
-		':(exclude)var/lib/pacman/local/git-extra-*/desc' \
+		':(exclude)var/lib/pacman/local/mingw-w64-*-git-extra-*/desc' \
```
dscho, line 173: Here, too, I would like to accept both old _and_ new name.

```suggestion
		':(exclude)var/lib/pacman/local/git-extra-*/desc' \
		':(exclude)var/lib/pacman/local/mingw-w64-*-git-extra-*/desc' \
```

git-extra/PKGBUILD
```diff
@@ -111,7 +111,12 @@ build() {
   export CXX=${MINGW_CHOST}-g++
   export SRCDIR=$srcdir
   export BUILDDIR=build-${MINGW_CHOST}
+  export TMPDIR=tmp
   mkdir -p ${BUILDDIR}
+  mkdir -p ${TMPDIR}
+  sed -e "s|@@PACKAGE_NAME@@|${MINGW_PACKAGE_PREFIX}-${_realname}|g" \
+    <update-via-pacman.bat >"$TMPDIR"/update-via-pacman.bat ||
+    (echo "Could not generate update-via-pacman script" && exit 1)
```
dscho, line 119: The canonical way to do this is to have a `update-via-pacman.bat.in` script and generate the `update-via-pacman.bat` version from it. Also, could we use `GIT_EXTRA_PACKAGE_NAME` instead of the less-descriptive `PACKAGE_NAME`?

dscho (changes requested):
We will also have to take care of [`git-sdk-arm64`'s `sync` job](https://github.com/git-for-windows/git-sdk-arm64/blob/a62385e2b3b878f37e6555e17ecc25dec4c18d35/.github/workflows/sync.yml#L76-L77), but that can only happen once we register the aarch64 Pacman repository.

git-extra/PKGBUILD
dennisameling: Makes sense - in https://github.com/git-for-windows/build-extra/pull/461/commits/6746dc5cb33a1b05cc423df1e36cf1058d48c6cf, I went with an approach to copy `update-via-pacman.bat` into the build folder instead of the `startdir`, as its contents will be different for every architecture. So we'll only have an `update-via-pacman.bat.in` in the `startdir` moving forward. WDYT?

dscho: > So we'll only have an `update-via-pacman.bat.in` in the `startdir` moving forward. WDYT?

@dennisameling sounds good! TBH I hope that we can soon replace `update-via-pacman.bat` with the `.ps1` script from https://github.com/git-for-windows/git-sdk-arm64/pull/4 (meaning: we can simply delete it from `git-extra`, as the script will live directly in the respective `git-sdk-*` repository).

dscho (approved):
Wonderful!

PR #455, opened 2022-12-20
git-extra: publish both with and without MINGW prefix

dennisameling:
Following up on [this discussion](https://github.com/git-for-windows/build-extra/pull/452#discussion_r1052009722).

Until now, the `git-extra` package was published under that exact name. However, it's _actually_ a MINGW package which installs things both in e.g. `/mingw64/bin` and `/etc/profile.d`. This is problematic for GfW's upcoming arm64 support, which uses the x64 MSYS shell, so it'll need its own package. At the same time, packages for multiple architectures can't be installed side-by-side like other MINGW-packages, as this one writes to shared folders like `/etc/profile.d` as well. Therefore we mark them as conflicting each other.

With this change, we'll publish the following packages side by side, so the repos will look like this:
- `git-for-windows/git-extra` (old)
- `git-for-windows-mingw32/git-extra` (old)
- `git-for-windows/mingw-w64-x86_64-git-extra` (new)
- `git-for-windows-mingw32/mingw-w64-i686-git-extra` (new)
- `git-for-windows-clangarm64/mingw-w64-aarch64-git-extra` (new)

Publishing side by side allows us to gradually update other parts of GfW to start using the new package names moving forward.

## Test results

After building with `MINGW_ARCH=mingw64 makepkg-mingw`, two files were created:
- `git-extra-1.1.614.d551cf0cc-1-any.pkg.tar.zst` (legacy package: `git-extra`)
- `mingw-w64-x86_64-git-extra-1.1.614.d551cf0cc-1-any.pkg.tar.zst` (new package: `mingw-w64-git-extra`)

Assuming that `git-extra` is already installed (e.g. when working from `git-sdk-64`), let's test these packages.

1. Updating the existing `git-extra` package still works as normal after the changes in this PR:

```
$ pacman -U git-extra-1.1.614.d551cf0cc-1-any.pkg.tar.zst
loading packages...
resolving dependencies...
looking for conflicting packages...

Packages (1) git-extra-1.1.614.d551cf0cc-1

Total Installed Size:  0.47 MiB
Net Upgrade Size:      0.00 MiB

:: Proceed with installation? [Y/n] Y
(1/1) checking keys in keyring                                                      [###############################################] 100%
(1/1) checking package integrity                                                    [###############################################] 100%
(1/1) loading package files                                                         [###############################################] 100%
(1/1) checking for file conflicts                                                   [###############################################] 100%
(1/1) checking available disk space                                                 [###############################################] 100%
:: Processing package changes...
[...]

PR #463, opened 2022-12-28
git-extra: remove traces of old ARM64 logic

dennisameling:
In the past, we tried working towards a situation where arm64 support was added to Git for Windows by relying on some executables in the `/mingw32` folder. That's no longer needed now that there's a native `/clangarm64`. Let's remove and update the old logic.


git-extra/env.sh
```diff
@@ -20,14 +20,12 @@ test -n "$SSH_ASKPASS" || {
...
 		;;
-	ARM64)
+	CLANGARM64)
 		export DISPLAY=needs-to-be-defined
-		if [ -f /arm64/bin/git-askpass.exe ]; then
-			export SSH_ASKPASS=/arm64/bin/git-askpass.exe
-		elif [ -f /mingw32/bin/git-askpass.exe ]; then
-			export SSH_ASKPASS=/mingw32/bin/git-askpass.exe
+		if [ -f /clangarm64/bin/git-askpass.exe ]; then
+			export SSH_ASKPASS=/clangarm64/bin/git-askpass.exe
```
rimrul, line 26: This is currently built with `CC=${MINGW_CHOST}-gcc`. I guess that's internally handled by `clang` on ARM64 through some compatibility layer. Otherwise we might need to tweak the `PKGBUILD` to build those executables natively.

dennisameling: There's a compatibility layer indeed: [`gcc-compat`](https://packages.msys2.org/package/mingw-w64-clang-aarch64-gcc-compat?repo=clangarm64). This ensures that things are handled by `clang` internally. `git-extra` [builds just fine](https://github.com/git-for-windows/build-extra/pull/455) using Clang 👍🏼 

Moving forward, we should update that to `CC=${MINGW_CHOST}-cc` instead, so that either GCC or Clang is used, depending on what `MSYSTEM` we're in. The MSYS2 team has done that for basically all `MINGW-packages` ~1 year ago: https://github.com/msys2/MINGW-packages/commit/7c67d9c46554d10276b2017bcfd0290dfe24a1bf

PR #464, opened 2023-01-01
Add a script to downgrade Pacman packages

dscho:
We just downgraded a couple of packages to accommodate for Git for Windows sticking to OpenSSL v1.1.x while MSYS2 moved on to OpenSSL v3.x

Unfortunately, an upgrade of the `pacman` package undid this, so we had to redo it.

Here is a script that was used to redo it. It takes as first parameter a shell scriptlet that tests whether the issue that needed to be fixed was actually fixed, followed by a list of packages to downgrade to the previous version. This script was used to auto-generate https://github.com/git-for-windows/git-sdk-64/commit/4c5ca7dbd83, by calling

	/usr/src/build-extra/downgrade-pacman-package.sh \
		 'bsdtar -h || test $? != 127' \
		 bsdcpio bsdtar libarchive libserf libssh2 rsync

This closes https://github.com/git-for-windows/git/issues/4197.

### DIFF
--- a/downgrade-pacman-package.sh
+++ b/downgrade-pacman-package.sh
@@ -1,0 +1,78 @@
+#!/bin/sh
+
+die () {
+    echo "$*" >&2
+    exit 1
+}
+
+test $# -gt 1 ||
+die "Usage: $0 <test-scriptlet> <package>..."
+
+test="$1"; shift
+
+git -C / update-index --ignore-submodules --refresh &&
+git -C / diff-files --ignore-submodules &&
+git -C / diff-index --cached --ignore-submodules HEAD ||
+die "Uncommitted changes"
+
+msg=
+for package
+do
+	dir="$(ls -d /var/lib/pacman/local/$package-[0-9]*)" ||
+	die "Package '$package' not installed?"
+
+	current_version=${dir#*/$package-}
+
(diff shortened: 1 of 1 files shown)


dennisameling (commented):
Overall this looks great - thank you!

Note that this will also _upgrade_ packages if the most recent commit downgraded them already:

```
$ /usr/src/build-extra/downgrade-pacman-package.sh \
> 'bsdtar -h || test $? != 127' \
> bsdcpio bsdtar libarchive
Downgrading 'bsdcpio' from 3.6.2-2 to 3.6.2-3
...
```

We might want to add some check to ensure that those commits actually contain _lower_ versions than the currently installed ones. On the other hand, I'd be amazed if we'd have to downgrade packages twice in a row... WDYT?

dscho:
> We might want to add some check to ensure that those commits actually contain _lower_ versions than the currently installed ones. On the other hand, I'd be amazed if we'd have to downgrade packages twice in a row... WDYT?

@dennisameling I'd be amazed, too. I have been amazed such a lot of times in the past, though, that I would not be all too surprised if we needed anything like that. When (or if?) that happens, we can always port [this logic](https://github.com/git-for-windows/build-extra/blob/3cda9eb70cca0e7e8de3d8c3e0076eb68e6c39d3/git-extra/git-update-git-for-windows#L8-L44) over.

PR #467, opened 2023-01-19
git-extra: add the `aarch64` Pacman repository when appropriate

dscho:
Once deployed, this will let git-sdk-arm64 update from Git for Windows' aarch64 Pacman repository.


dscho:
To avoid version clashes, this is based on #456 (and can only be merged after that).

Note: I just realized that we have a problem in our `mingw-w64-clang-aarch64` packages: I had assumed that the Pacman repository was already active, but it wasn't. For that reason, `mingw-w64-curl` was linked against MSYS2's version of `mingw-w64-openssl` (which is already on version v3.x), not against Git for Windows' version (which remains with v1.1.x until support for that version runs out).

Therefore, we will have some manual updating to do.

dscho:
> `mingw-w64-curl` was linked against MSYS2's version of `mingw-w64-openssl` (which is already on version v3.x), not against Git for Windows' version (which remains with v1.1.x until support for that version runs out).
> 
> Therefore, we will have some manual updating to do.

@dennisameling I guess the safest way would be to use [`downgrade-pacman-package.sh`](https://github.com/git-for-windows/build-extra/blob/main/downgrade-pacman-package.sh) to pin `mingw-w64-openssl` and `mingw-w64-curl` to the versions you committed initially, then rebuild the current `mingw-w64-curl`, and then unpin the packages. What do you think?

dennisameling:
> @dennisameling I guess the safest way would be to use [`downgrade-pacman-package.sh`](https://github.com/git-for-windows/build-extra/blob/main/downgrade-pacman-package.sh) to pin `mingw-w64-openssl` and `mingw-w64-curl` to the versions you committed initially, then rebuild the current `mingw-w64-curl`, and then unpin the packages. What do you think?

That should work indeed! Another option would be to merge this PR and have me install GfW's `openssl` version locally, then commit that to `git-sdk-arm64`. Future updates will then automatically use GfW's flavor. WDYT?

dennisameling (approved):

dscho:
> install GfW's `openssl` version locally, then commit that to `git-sdk-arm64`. Future updates will then automatically use GfW's flavor.

That won't work because `git push` will fail due to the missing OpenSSL v3.x DLL that `cURL` depends on... 😁

PR #472, opened 2023-01-29
git-prompt.sh: remove annoying space when MSYSTEM not set

DRSDavidSoft:
Signed-off-by: David Refoua <David@Refoua.me>

This commit removes the annoying double space in the first line of PS1 when `MSYSTEM` is not set.

### DIFF
--- a/git-extra/git-prompt.sh
+++ b/git-extra/git-prompt.sh
@@ -15,6 +15,10 @@ else
 	PS1="$PS1"'\u@\h '             # user@host<space>
 	PS1="$PS1"'\[\033[35m\]'       # change to purple
 	PS1="$PS1"'$MSYSTEM '          # show MSYSTEM
+	if [ ! -z "$MSYSTEM" ]
+	then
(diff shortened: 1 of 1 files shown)


DRSDavidSoft:
CI is failing a `sha256sum` checksum

git-extra/git-prompt.sh
```diff
@@ -15,6 +15,10 @@ else
 	PS1="$PS1"'\u@\h '             # user@host<space>
 	PS1="$PS1"'\[\033[35m\]'       # change to purple
 	PS1="$PS1"'$MSYSTEM '          # show MSYSTEM
```
rimrul, line 17: Wouldn't this line need to be removed for this to work?

dscho (changes requested):

dscho:
> While I suspect that the use case is broken (because MSYSTEM should always be set), I am open to getting convinced that it makes sense.

This PR has not seen any attempt to motivate the change in the past 3 weeks, therefore I'll close it.

PR #475, opened 2023-02-06
create-sdk-artifact: automatically choose appropriate architecture

dscho:
I had totally missed in my review that the default was for `bitness=auto`, and hence `./please.sh create-sdk-artifact build-installers` would automatically pick the appropriate bitness. That default logic was lost, but it is really, really nice to have. Let's have it again, this time with `architecture=auto`.


dennisameling (approved):

chad-baker47:
either moving forward or taking two steps back at least we got programs in place to help us move forward

PR #478, opened 2023-02-08
please.sh: make bundle_pdbs arm64-ready

dennisameling:
As reported [here](https://github.com/git-for-windows/git-for-windows-automation/pull/22#issuecomment-1421268074), the `installer` artifact for `arm64` couldn't be created completely as `please.sh bundle-pdbs` doesn't support it yet. This PR adds the support.

Example artifact: 
[pdbs-for-git-arm64-2.39.1.windows.1.531.gb3665b1bf4-1.zip](https://github.com/git-for-windows/build-extra/files/10689119/pdbs-for-git-arm64-2.39.1.windows.1.531.gb3665b1bf4-1.zip)

### DIFF
--- a/please.sh
+++ b/please.sh
@@ -2848,14 +2848,17 @@ bundle_pdbs () { # [--directory=<artifacts-directory] [--unpack=<directory>] [--
 	artifactsdir="$(cygpath -am "$HOME")" || exit
 
 	test -n "$architectures" ||
-	architectures="i686 x86_64"
+	architectures="i686 x86_64 aarch64"
 
 	versions="$(case $# in 0) pacman -Q;; 1) cat "$1";; esac |
-		sed 's/^\(mingw-w64\)\(-[^-]*\)/\1/' | sort | uniq)"
+		sed 's/^\(mingw-w64\)\(-clang-[^-]*\|-[^-]*\)/\1/' | sort | uniq)"
 	test -n "$versions" ||
 	die 'Could not obtain package versions\n'
 
 	git_version="$(echo "$versions" | sed -n 's/^mingw-w64-git //p')"
+
+	test -n "$git_version" ||
+		die "Could not determine Git version"
 
 	dir="${this_script_path:+$(cygpath -au \
 		"${this_script_path%/*}")/}"cached-source-packages
@@ -2868,15 +2871,28 @@ bundle_pdbs () { # [--directory=<artifacts-directory] [--unpack=<directory>] [--
 
 	for arch in $architectures
 	do
-		test i686 = $arch &&
-		bitness=32-bit ||
-		bitness=64-bit
+		echo "Unpacking .pdb files for $arch..." >&2
 
-		echo "Unpacking .pdb files for $bitness..." >&2
-
-		test x86_64 = $arch &&
-		oarch=x86-64 ||
-		oarch=$arch
+		case $arch in
+			x86_64)
+				oarch=x86-64
+				pacman_arch=x86_64
+				artifact_suffix=64-bit
+				;;
+			i686)
+				oarch=i686
+				pacman_arch=i686
+				artifact_suffix=32-bit
+				;;
+			aarch64)
+				oarch=aarch64
+				pacman_arch=clang-aarch64
+				artifact_suffix=arm64
+				;;
+			*)
+				die "Unhandled architecture: $arch"
+				;;
+		esac
 
 		test -z "$artifactsdir" ||
 		test ! -d $unpack ||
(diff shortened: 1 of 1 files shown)


please.sh
```diff
@@ -2851,12 +2852,15 @@ bundle_pdbs () { # [--directory=<artifacts-directory] [--unpack=<directory>] [--
 	architectures="i686 x86_64"
```
dennisameling, line 2852: I'd actually be in favor of only supporting a single architecture at a time in `bundle_pdbs` to make it consistent with most of the other stuff we've been refactoring. WDYT?

please.sh
```diff
@@ -2890,16 +2909,13 @@ bundle_pdbs () { # [--directory=<artifacts-directory] [--unpack=<directory>] [--
...
 			mingw-w64-*)
-				tar=mingw-w64-$arch-${package#mingw-w64-}-$version-any.pkg.tar.xz
-				dir2="$sdk64/usr/src/MINGW-packages/$name"
+				tar=mingw-w64-$pacman_arch-${package#mingw-w64-}-$version-any.pkg.tar.xz
 				;;
 			*)
-				tar=$package-$version-$arch.pkg.tar.xz
-				test i686 = $arch &&
-				dir2="$sdk32/usr/src/MSYS2-packages/$name" ||
-				dir2="$sdk64/usr/src/MSYS2-packages/$name"
```
dscho, line 2901: I fear that we still need to keep the `dir2` definition in these `case` arms, as the MSYS packages need to be found in `/usr/src/MSYS2-packages/<package-name>/`, not in ``/usr/src/MINGW-packages/<package-name>/`.

The original idea of this `dir2` business is that we cannot download the `-pdb` package for a `mingw-w64-git` package that we just built (and that we might not even deploy, if manual testing finds something incorrect in the Git artifacts and another workflow run has to be started).

So I think we still need this to look in the correct path, even if the common case  is that only a _MINGW_ package (concretely, `mingw-w64-git`) cannot be downloaded and needs to be found locally.

please.sh
```diff
@@ -116,6 +116,7 @@ sdk_path () { # <bitness>
 
 sdk64="$(sdk_path 64)"
 sdk32="$(sdk_path 32)"
+sdkarm64="$(sdk_path arm64)"
```
dscho, line 119: I fear that we'll have to do some major cleanup here.

The original design was for the `please.sh` script to be called in a dedicated VM that was the home of both i686 and x86_64 SDK, and both should be handled in synchrony. For example: whenever I built `msys2-runtime`, it was to be built in both SDKs, and installed in both SDKs. That's why there is an `sdk_path` to begin with.

However, we're safely outside this original scenario. Long gone is the time that I even ran VMs on my own machine.

And only on a Windows/ARM64 machine could we even _use_ side-by-side i686/x86_64/aarch64 SDKs. But I don't have such a machine (except a relatively slow Azure VM that I only power up when I need to work on Windows/ARM64 stuff).

The common case, therefore, is that the "sdk path" is `/`.

So I think that soon we will need to do some major spring cleaning and rip out this idea that more than one SDK is available, and that one should be "callable" from another one. Read: we will want to get rid of `sdk_path`, `sdk64`, `sdk32` and `sdkarm64` as well as all of their users.

dscho (commented):
This looks quite good!

I think we will still need to move the `dir2` assignments back into their respective `case` arms, so that the script looks for `mingw-w64-*-pdb` packages in `/usr/src/MINGW-packages/<package-name>` and for the other packages in `/usr/src/MSYS2-packages/<package-name>`.

Other than that, this is ready to merge, methinks!

please.sh
dennisameling: Whoops, I completely overlooked the fact that it distinguished between `MINGW-packages` and `MSYS2-packages` in the first place. Apologies.

Given that `MSYS2-packages` can only be built for `x86_64` and `i686` at this point (as opposed to `MINGW-packages`, which supports `aarch64` as well), shall I just restore the original logic for this arm?

And for the `mingw-w64-*` arm, I think I can just replace `$sdk64/usr/src/MINGW-packages/$name` with `/usr/src/MINGW-packages/$name`, right? Same for the `$sdk64/git-cmd.exe` invocations below. If that works, we also don't have to introduce the `sdkarm64` variable above. WDYT?

dscho: > Given that `MSYS2-packages` can only be built for `x86_64` and `i686` at this point (as opposed to `MINGW-packages`, which supports `aarch64` as well), shall I just restore the original logic for this arm?

Yes, please!

> And for the `mingw-w64-*` arm, I think I can just replace `$sdk64/usr/src/MINGW-packages/$name` with `/usr/src/MINGW-packages/$name`, right? Same for the `$sdk64/git-cmd.exe` invocations below. If that works, we also don't have to introduce the `sdkarm64` variable above. WDYT?

That sounds really good!

dennisameling:
Think we're good to go now. I tested on arm64 and everything's looking good. Would appreciate if you could do some testing as well just to be sure. Thanks!

dscho:
> Would appreciate if you could do some testing as well just to be sure.

```
[...]
Retrieving mingw-w64-clang-aarch64-git-pdb-2.39.1.1.b03dafd9c2-1-any.pkg.tar.xz...
curl error mingw-w64-clang-aarch64-git-pdb-2.39.1.1.b03dafd9c2-1-any.pkg.tar.xz (22)
```

... because those have not yet been built ;-)

I'll just merge it and we'll take it from there.

PR #479, opened 2023-02-08
installer: drop support for Windows 7 and 8.0

dennisameling:
[As announced in July 2022](https://github.com/git-for-windows/build-extra/blob/main/ReleaseNotes.md#changes-since-git-for-windows-v2371-july-12th-2022), Git for Windows is dropping support for these older Windows versions, [following MSYS2's lead](https://github.com/msys2/msys2.github.io/blob/source/web/news.md#2023-01-15---dropping-support-for-windows-7-and-80). Let's make sure that the installer actively blocks the installation process on these versions. The minimal Windows version is now 8.1 (6.3).

Looking ahead: if the minimum version reaches `10.0`, GfW could choose to switch to `CLANG64` instead of the current `MINGW64`, [which is using the more modern `ucrt` C library](https://www.msys2.org/docs/environments/). Our recent work on `CLANGARM64` support paves the way for this potential future switch ✨

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -95,6 +95,7 @@ WizardImageBackColor=clWhite
 WizardImageStretch=no
 WizardImageFile={#SourcePath}\git.bmp
 WizardSmallImageFile={#SourcePath}\gitsmall.bmp
+MinVersion=6.3
 
 [Types]
 ; Define a custom type to avoid getting the three default types.


dscho:
> [As announced in July 2022](https://github.com/git-for-windows/build-extra/blob/main/ReleaseNotes.md#changes-since-git-for-windows-v2371-july-12th-2022), Git for Windows is dropping support for these older Windows versions, [following MSYS2's lead](https://github.com/msys2/msys2.github.io/blob/source/web/news.md#2023-01-15---dropping-support-for-windows-7-and-80)

While that is true, I forgot to repeat that warning in quite a few releases since then. That's why I [added the warning again, for the next release](https://github.com/git-for-windows/build-extra/commit/990e62b97a13056a9ddabb484d3822d94ae98e0f) yesterday.

So I think that this PR might need a while before it can be merged.

But you got me thinking: maybe we should warn users who install the next Git for Windows version if their Windows version will soon no longer be supported? Something similar to [the warning about a downgrade](https://github.com/git-for-windows/build-extra/blob/9ea75ff24db0fdadf771bf7ce4b6c1f5826bf4ec/installer/install.iss#L1109-L1121), except instead of checking for the _Git for Windows version_ we would check for the _Windows version_, similar to what we do [when detecting support for symbolic links](https://github.com/git-for-windows/build-extra/blob/9ea75ff24db0fdadf771bf7ce4b6c1f5826bf4ec/installer/install.iss#L1255-L1262))?

And once the time actually comes to drop Windows 7/8.0 support, the big change will be to upgrade to MSYS2 runtime v3.5.* (which is not even available yet because Cygwin did not release any v3.5.* version yet). But maybe we should follow Cygwin's lead in prominently declaring one version "the last one to support Windows 7"? Like they do on their [home page](https://cygwin.com/):

image

> Looking ahead: if the minimum version reaches `10.0`, GfW could choose to switch to `CLANG64` instead of the current `MINGW64`, [which is using the more modern `ucrt` C library](https://www.msys2.org/docs/environments/). Our recent work on `CLANGARM64` support paves the way for this potential future switch ✨

Given how finicky even something like the `malloc()` performance of different CRT versions is (in my `mimalloc` work, I was able to replicate some poor performance numbers even with recent CRT versions), I am not quite as enthusiastic about that switch. It sounds like quite a risk to me, to be honest, especially in conjunction with switching from GCC to clang (which implies using a low-level virtual machine whose performance characteristics I never thoroughly investigated).

dscho:
Given that MSYS2 runtime v3.5.0 will drop support for Windows 7/8, but v3.4.0 _already_ dropped i686 support, I realized that the latter is much more pressing. I opened https://github.com/git-for-windows/git/issues/4279 for that.

dennisameling:
Ok, let's just mark this PR as draft then. Feel free to close this completely if you feel this is premature 👍🏼

dscho:
> Ok, let's just mark this PR as draft then. Feel free to close this completely if you feel this is premature 👍🏼

Oh, let's keep it open. I would not even have minded keeping it in non-draft mode ;-)

rimrul:
> Given that MSYS2 runtime v3.5.0 will drop support for Windows 7/8

https://github.com/git-ecosystem/git-credential-manager/pull/1418

Git Credential Manager is now also working on dropping support for Windows 7/8.

rimrul:
/add relnote bug The installer now correctly blocks the installation on Windows 7 and Windows 8 as these versions of Windows are no longer supported since Git for Windows v2.47.0

PR #481, opened 2023-02-11
Really only add the aarch64 Pacman repository for git-sdk-arm64

dscho:
In https://github.com/git-for-windows/git-sdk-64/commit/40af5e29d1efdfb384e25b21891edc64a254099f, the `filesystem` package got updated, which contains the `/clangarm64` directory (which contains quite a few subdirectories, but no files).

This fooled the post-install script of `git-extra` into believing that git-sdk-64 actually targets Windows/ARM64 and hence it added Git for Windows' aarch64 Pacman repository.

I [manually reverted](https://github.com/git-for-windows/git-sdk-64/commit/cce08432ac15f6d68b9b5a2b4a5a7ee4e3823084) that to _not_ register the aarch64 Pacman repository.

However, the next update of the `filesystem` would trigger the same issue.

Let's make the post-install script of `git-extra` smarter so that it does not get fooled anymore.

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -5,7 +5,7 @@ pkgbase="mingw-w64-${_realname}"
 pkgname=($_realname
         "${MINGW_PACKAGE_PREFIX}-${_realname}")
 _ver_base=1.1
-pkgver=1.1.628.f599c8ce7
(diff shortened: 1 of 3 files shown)


dennisameling (approved):
Thank you!

PR #482, opened 2023-02-21
installer: warn about 32-bit deprecation

dscho:
As per https://github.com/git-for-windows/git/issues/4279, Git for Windows is deprecating the 32-bit variant. Let's teach the installer to detect when a user is installing this variant and show an appropriate warning in that case. On a 32-bit Windows, they will see this type of dialog box:

image

On a 64-bit Windows, they will see this dialog box:

image

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -44,6 +44,8 @@ This package contains software from a number of other projects including Bash, z
 
 As announced previously, Git for Windows will drop support for Windows 7 and for Windows 8 in one of the next versions, following [Cygwin's and MSYS2's lead](https://www.msys2.org/docs/windows_support/) (Git for Windows relies on MSYS2 for components such as Bash and Perl).
 
+Also following the footsteps of the MSYS2 and Cygwin projects on which Git for Windows depends, the 32-bit variant of Git for Windows [is nearing its end of support](https://gitforwindows.org/32-bit.html).
+
 ### New Features
 
 * In the olden Git days, there were "dashed" Git commands (e.g. `git-commit` instead of `git commit`). These haven't been supported for interactive use in a really, really long time. But they still worked in Git aliases and hooks ("scripts"). Since Git v1.5.4 (released on February 2nd, 2008), it was discouraged/deprecated to use dashed Git commands even in scripts. As of this version, Git for Windows [no longer supports these dashed commands](https://github.com/git-for-windows/git/pull/4252).
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1091,6 +1091,10 @@ begin
     UpdateInfFilenames;
 #if BITNESS=='32'
     Result:=True;
+    if not IsWin64 then
+        SuppressibleMsgBox('Git for Windows (32-bit) is nearing its end of support.'+#13+'More information at https://gitforwindows.org/32-bit.html',mbError,MB_OK or MB_DEFBUTTON1,IDOK)
+    else if SuppressibleMsgBox('Git for Windows (32-bit) is nearing its end of support. It is recommended to install the 64-bit variant of Git for Windows instead.'+#13+'More information at https://gitforwindows.org/32-bit.html'+#13+'Continue to install the 32-bit variant?',mbError,MB_YESNO or MB_DEFBUTTON2,IDNO)=IDNO then
+        Result:=False;
 #else
     if not IsWin64 then begin
         LogError('The 64-bit version of Git requires a 64-bit Windows. Aborting.');


mjcheetham (approved):
👍 Looks great! Don't suppose adding a link to info about the end of support is possible, given this is a simple MsgBox?

dscho:
> 👍 Looks great! Don't suppose adding a link to info about the end of support is possible, given this is a simple MsgBox?

@mjcheetham unfortunately, [it seems that this is not possible](https://stackoverflow.com/questions/1493198/how-to-show-a-hyperlink-in-inno-setup). The suggested workaround would be to let one of the buttons of the message box open a link, but out of [the available buttons](https://jrsoftware.org/ishelp/topic_isxfunc_msgbox.htm) ("OK", "Cancel", "Abort", "Retry", "Ignore", "Yes", and "No"), nothing strikes me as an obvious good choice.

But I guess I could reword the message box thusly:

> Git for Windows (32-bit) is nearing its end of support.
> Would you like to open the ticket with more information?
> <kbd>Yes</kbd> <kbd>No</kbd>

Another question is whether I should mention v2.40.x explicitly, or whether it is better to stay vague because we will continue to build 32-bit Git for Windows even with v2.41.x, it's just that the MSYS2 runtime won't be the latest and greatest anymore...

What do you think?

mjcheetham:
> > 👍 Looks great! Don't suppose adding a link to info about the end of support is possible, given this is a simple MsgBox?
> 
> @mjcheetham unfortunately, [it seems that this is not possible](https://stackoverflow.com/questions/1493198/how-to-show-a-hyperlink-in-inno-setup). The suggested workaround would be to let one of the buttons of the message box open a link, but out of [the available buttons](https://jrsoftware.org/ishelp/topic_isxfunc_msgbox.htm) ("OK", "Cancel", "Abort", "Retry", "Ignore", "Yes", and "No"), nothing strikes me as an obvious good choice.

Yeah none of those make any sense.

> But I guess I could reword the message box thusly:
>
> > Git for Windows (32-bit) is nearing its end of support.
> > Would you like to open the ticket with more information?
> > Yes No

That may work too (except the term "ticket" sounds weird..), but what about the 32-on-64 message? How about:

> Git for Windows (32-bit) is nearing its end of support.
> Read more information at gitforwindows.org/32bit.
> <kbd>OK</kbd>

..and..

> Git for Windows (32-bit) is nearing its end of support.
> It is recommended to install the 64-bit variant of Git for Windows instead.
> Read more information at gitforwindows.org/32bit.
> Continue to install the 32-bit variant?
> <kbd>Yes</kbd> <kbd>No</kbd>

> Another question is whether I should mention v2.40.x explicitly, or whether it is better to stay vague because we will continue to build 32-bit Git for Windows even with v2.41.x, it's just that the MSYS2 runtime won't be the latest and greatest anymore...
> 
> What do you think?

Staying vague is probably better.

dscho:
@mjcheetham I force-pushed an update (and with this update, this here PR needs https://github.com/git-for-windows/git-for-windows.github.io/pull/43 to be merged first). The dialog boxes now look like this:

image

image

mjcheetham:
> The dialog boxes now look like this

Looks great! I think adding the URL is helpful as overtime you can update the webpage with more concrete dates and version numbers etc.

dscho:
Technically, it _would_ be [possible to implement a custom message box with an "Open link" button](https://stackoverflow.com/questions/47107177/inno-setup-create-custom-message-box-yes-no). However, the styling would be different, and to support the "Suppressible" part of [`SuppressibleMsgBox`](https://jrsoftware.org/ishelp/topic_isxfunc_suppressiblemsgbox.htm) we would have to add even more code.

I've decided that there are more pressing things that I need to attend to ;-)

mjcheetham:
> Technically, it _would_ be [possible to implement a custom message box with an "Open link" button](https://stackoverflow.com/questions/47107177/inno-setup-create-custom-message-box-yes-no). However, the styling would be different, and to support the "Suppressible" part of [`SuppressibleMsgBox`](https://jrsoftware.org/ishelp/topic_isxfunc_suppressiblemsgbox.htm) we would have to add even more code.

By "link" I mean adding the URL, which you have done. No need to add a hyperlink!

> I've decided that there are more pressing things that I need to attend to ;-)

100%

dscho:
> > Technically, it _would_ be [possible to implement a custom message box with an "Open link" button](https://stackoverflow.com/questions/47107177/inno-setup-create-custom-message-box-yes-no). However, the styling would be different, and to support the "Suppressible" part of [`SuppressibleMsgBox`](https://jrsoftware.org/ishelp/topic_isxfunc_suppressiblemsgbox.htm) we would have to add even more code.
> 
> By "link" I mean adding the URL, which you have done. No need to add a hyperlink!

Still, a hyperlink would be nicer for users.

asjimene:
Hello, please let me know if you would prefer an issue opened for this. It seems this change breaks installing the 32-bit version silently with /SUPPRESSMSGBOXES, as it defaults to "No" and cancels the installation when running silently.

Was this the intended outcome of this change?

rimrul:
> It seems this change breaks installing the 32-bit version silently with /SUPPRESSMSGBOXES, as it defaults to "No" and cancels the installation when running silently.

Note that this behaviour only affects 64bit Windows, as the message box has no "No" option on 32bit Windows. See also #492 and https://github.com/git-for-windows/git/issues/4347.

According to https://github.com/git-for-windows/git/issues/4347#issuecomment-1471434626 this is intentional.

asjimene:
Ahh, that is what I needed to know. Thanks for pointing that out and apologies for not finding it on my own!

PR #483, opened 2023-02-23
Add winget to default Win32 aliases

Trenly:
[winget](/microsoft/winget-cli) is the Microsoft Windows Package Manager and is included with newer versions of Windows by default. The application requires a win32 console, so including it here as a default will reduce user friction


Trenly:
I see the build is failing; Is this not the repository this change should be made in?

dscho:
@Trenly you simply need to call `updpkgsums` and commit the changed checksum in `PKGBUILD`, too. That'll fix it.

Trenly:
I must be doing something wrong, since I get this error when trying to use `updpkgsums`
```
==> ERROR: pkgname is not allowed to start with a hyphen.
==> ERROR: Extra package() function for split package 'mingw-w64-git-extra'
==> ERROR: Missing package_-git-extra() function for split package '-git-extra'
==> ERROR: Failed to generate new checksums
```

I'm a windows user just trying to figure out how to contribute here, so I appreciate your patience and guidance

dscho:
> I get this error when trying to use `updpkgsums`
> 
> ```
> ==> ERROR: pkgname is not allowed to start with a hyphen.
> ```

Most likely `MSYSTEM` was not set to `MINGW64`. And therefore I suspect [`MINGW_PACKAGE-PREFIX`](https://github.com/git-for-windows/build-extra/blob/4d1a8a1d2b7ca3ff607d065e1cadc4083d9b4e72/git-extra/PKGBUILD#L6) not to be set.

The remaining problem is that the `pkgver` is automatically updated when new commits are added that touch `git-extra/`.

I do agree that none of these details should be required to be known to contributors. We're constantly improving these mechanics; In response to this here PR, I opened https://github.com/git-for-windows/gfw-helper-github-app/issues/28 with a plan to remove a lot of friction in the future.

Trenly:
Thank you very much for the detailed explanation and your assistance!

martin-braun:
So, how do you install `git-extra` via `winget` though?

dscho:
> So, how do you install `git-extra` via `winget` though?

You don't. `git-extra` is a Pacman package and lives within the context of the Git for Windows SDK, which is a close derivative of the [MSYS2](https://msys2.org/) project.

Git for Windows' official releases are built from subsets of those Pacman packages in Git for Windows' SDK.

`winget` is a completely different package management system, with completely different boundaries: it apparently has [an unofficial `Git` package](https://github.com/microsoft/winget-pkgs/tree/master/manifests/g/Git/Git) that is fed Git for Windows' official releases via community contributions. As such, it will only receive `git-extra` updates once they made it into an official Git for Windows release. The next such official release is scheduled for this coming Thursday or Friday, depending on the official _Git_ release (which is [scheduled](https://gh.io/gitCal) for Thursday _some time of day_).

martin-braun:
@dscho Thank you for the detailed response, I think I mis-understood what `git-extra` is then. The reason why I got here is I was hoping that you can install Git for Windows SDK (incl. pacman) via `winget`. I made an issue about that on `winget`. I'm very amazed by possibilities of Git for Windows SDK, and I wonder why `pacman` is not shipped with the normal Git for Windows, because it enables so many great features like installing `tmux` and so on. I guess you want Git for Windows to be just that, Git for Windows, but it is so much for me to be honest. 

No I don't want to use regular MSYS2 as I value stability over feature set which is what you are going for. Thanks so much, your work makes Windows fine to use, which I have to use to compile some Windows stuff for my work.

PR #484, opened 2023-02-26
Fix git extra.install quoting

dscho:
This faulty quoting somehow slipped in and for some reason only showed up yesterday in [the automated builds](https://github.com/git-for-windows/git-sdk-64/actions/runs/4268074874/jobs/7430553454#step:6:1034):

> ```
> /var/lib/pacman/local/mingw-w64-x86_64-git-extra-1.1.632.0366f1328-1/install: line 292: unexpected EOF while looking for matching ``'
```

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -5,7 +5,7 @@ pkgbase="mingw-w64-${_realname}"
 pkgname=($_realname
         "${MINGW_PACKAGE_PREFIX}-${_realname}")
 _ver_base=1.1
-pkgver=1.1.632.0366f1328
+pkgver=1.1.634.5b4d8fffc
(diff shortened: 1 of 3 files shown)


dscho:
Seems that [this error](https://github.com/git-for-windows/git-sdk-64/actions/runs/4273106721/jobs/7438791786#step:3:143) was missed in the `sync` workflow (probably because error checking needs to be improved?).

PhilipOakley:
So glad it wasn't just me. 

I'd been trying to ensure that I'd got my SDK up to date, and has hit that as one of the many 'little' problems on a merry-go-round of confusing (to me) errors. Very hopeful I'll be back up running after this.

Many thanks.

PR #492, opened 2023-04-15
installer: allow silent install of 32-bit installer on ARM64

rimrul:
When Installing 32-bit Git for Windows we show a Warning that we're deprecating 32-bit support, but there are two Versions of the 32-bit warning. The Version shown on 32-Bit Windows allows a silent install, while the Version on 64-Bit Windows does not. This is fine on AMD64, but for Windows 10 on ARM64 we don't have a good alternative offer. So show them the same message as we show on i686.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1091,7 +1091,7 @@ begin
     UpdateInfFilenames;
 #if BITNESS=='32'
     Result:=True;
-    if not IsWin64 then
+    if not IsX64 then
         SuppressibleMsgBox('Git for Windows (32-bit) is nearing its end of support.'+#13+'More information at https://gitforwindows.org/32-bit.html',mbError,MB_OK or MB_DEFBUTTON1,IDOK)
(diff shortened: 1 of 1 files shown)


dscho:
> for Windows 10 on ARM64 we don't have a good alternative offer.

Do we really have to care about Windows 10 for ARM64? It was my understanding that it was still marked as experimental, and that you'd really would need to go for Windows 11 if you want to use a full-featured Windows/ARM64.

_If_ we want to allow a silent install of 32-bit Git for Windows on ARM64, then I would insist on doing that only on Windows 10. On Windows 11 users should install the 64-bit variant.

rimrul:
> Do we really have to care about Windows 10 for ARM64?

I'd treat it as regular x86 Windows 10, as far as the installer is concerned.

> _If_ we want to allow a silent install of 32-bit Git for Windows on ARM64, then I would insist on doing that only on Windows 10. On Windows 11 users should install the 64-bit variant.

I've thought about that, but I think I'd prefer to recommend the upcomming native ARM64 version there, once we have official ARM64 releases and avoid recommending a stepping stone in between. Though I guess the intermediary step via AMD64 is not the worst thing.

dscho:
> the upcomming native ARM64 version

There is still the blocker that we have unexplained (and unfixed) hangs in the MSYS2 runtime when code-signing the executables. Before that is resolved, we cannot move forward with Git for Windows/ARM64. And while I'd love to work on this (it seems like a good debugging challenge) I fear I currently lack the time and energy to do so.

PR #287, opened 2020-05-19
Address the fallout from the upgrade to MSYS2 runtime v3.1.4

dscho:
The main two things in this PR: we now offer an experimental option to opt-in to the pseudo console support, and we silence the `pull.rebase` warning proactively by forcing the user to choose in the installer.

While at it, I also slipped in a change accommodating for https://github.com/git-for-windows/MINGW-packages/pull/40: the `edit-git-bash.exe` helper is now included in the `mingw-w64-git` package (originally I had planned to modify the `git-bash.exe` via `edit-git-bash` so that it would enable pseudo consoles, but that would be confusing, as it would affect _only_ `git-bash.exe` (and not, say, `/cmd/git.exe -c alias.mintty='!mintty' mintty`).

PR #498, opened 2023-05-09
Clarify that `ff-only` is git's default pull behavior

kaste:
Fixes https://github.com/git-for-windows/git/issues/4407

We stated that "Merge" would be the default `git pull` behavior but it is not.  The default mode is in fact "ff-only" and we follow here.

image

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -2281,13 +2281,13 @@ begin
     GitPullBehaviorPage:=CreatePage(PrevPageID,'Choose the default behavior of `git pull`','What should `git pull` do by default?',TabOrder,Top,Left);
 
     // 1st choice
-    RdbGitPullBehavior[GP_GitPullMerge]:=CreateRadioButton(GitPullBehaviorPage,'Default (fast-forward or merge)','This is the standard behavior of `git pull`: fast-forward the current branch to'+#13+'the fetched branch when possible, otherwise create a merge commit.',TabOrder,Top,Left);
+    RdbGitPullBehavior[GP_GitPullMerge]:=CreateRadioButton(GitPullBehaviorPage,'Fast-forward or merge','Fast-forward the current branch to the fetched branch when possible,'+#13+'otherwise create a merge commit.',TabOrder,Top,Left);
 
     // 2nd choice
     RdbGitPullBehavior[GP_GitPullRebase]:=CreateRadioButton(GitPullBehaviorPage,'Rebase','Rebase the current branch onto the fetched branch. If there are no local'+#13+'commits to rebase, this is equivalent to a fast-forward.',TabOrder,Top,Left);
 
     // 3rd choice
-    RdbGitPullBehavior[GP_GitPullFFOnly]:=CreateRadioButton(GitPullBehaviorPage,'Only ever fast-forward','Fast-forward to the fetched branch. Fail if that is not possible.',TabOrder,Top,Left);
+    RdbGitPullBehavior[GP_GitPullFFOnly]:=CreateRadioButton(GitPullBehaviorPage,'Only ever fast-forward','Fast-forward to the fetched branch. Fail if that is not possible.'+#13+'This is the standard behavior of `git pull`.',TabOrder,Top,Left);
 
     // Restore the setting chosen during a previous install.
     case ReplayChoice('Git Pull Behavior Option','Merge') of


rimrul:
> I think git-for-windows choose the wrong default from the beginning:  You wanted to save the user from the new nag messages, but the nag messages were introduced preliminary to change the default (to ff-only).

It's simpler than that. We added a way to choose in the setup, way before  the messages existed. And we chose our  default based on the default at the time. When the nag messages came we saved the users from them, as a side effect, because they had already explicitly set a default.

kaste:
Ah okay, I thought git introduced it https://github.com/git/git/commit/d18c950a69f3a24e1e3add3d9fc427641f53e12b (Mar 10, 2020) and [`6ab14d9` (#287)](https://github.com/git-for-windows/build-extra/pull/287/commits/6ab14d9a169d7aeb3fdcf861148ec4d1a0de91cf) (May 20, 2020) here would be in reply of that.  Didn't look further.

dscho:
So we're at an impasse.

@kaste wants to unleash a new default to all Git for Windows users, based on the rationale that upstream Git forces users to make their choice explicit when trying to pull a divergent branch.

Git for Windows, on the other hand, asks users up front what Git's default pull behavior should be. And with millions of users, Git for Windows has to be careful when unleashing such a change. Disgruntled users will not be mad at @kaste, after all, they will be mad at the Git for Windows maintainers when their workflows and scripts need to be adapted.

So I am mildly negative about this PR, in particular because there was no room for a compromise to first reword the explanations (I'd be happy with a "(RECOMMENDED!) ... This is Git's default behavior.", for example) and then see where we're at in a year or so and maybe flip the default _then_.

You can convince me with a gentler deprecation path, but not with a firm no-compromises-possible stance.

kaste:
I don't see the "no-compromise" Haltung that much as I go for the lengthy discussion about wordings after the initial "It is not a wording issue."  

Now, three years later, I feel it might be just a wording issue more than before though, as we currently state "Default (fast-forward or merge)  This is the standard behavior of `git pull`..." and that might be just wrong and irritating.  

How about 

```
[x] Fast-forward or merge:  Fast-forward the current branch to the fetched branch when possible, otherwise create a merge commit.
[ ] Rebase:  Rebase the current branch onto the fetched branch. If there are no local commits to rebase, this is equivalent to a fast-forward.
[ ] Only ever fast-forward:  Fast-forward to the fetched branch. Fail if that is not possible. This is the standard behavior of `git pull`.   
```

I usually like a "Do nothing - decide later" option because it is the best option for new users (you can't decide the question if you don't know git) but you did not comment on that (and it is not part of the PR anyway).

dscho:
> How about
> 
> - [x] Fast-forward or merge:  Fast-forward the current branch to the fetched branch when possible, otherwise create a merge commit.
> - [ ] Rebase:  Rebase the current branch onto the fetched branch. If there are no local commits to rebase, this is equivalent to a fast-forward.
> - [ ] Only ever fast-forward:  Fast-forward to the fetched branch. Fail if that is not possible. This is the standard behavior of `git pull`.   

I like it.

> I usually like a "Do nothing - decide later" option because it is the best option for new users (you can't decide the question if you don't know git) but you did not comment on that (and it is not part of the PR anyway).

In my experience, this "do nothing, force the user to decide later, right when they just need to Get Stuff Done and Don't Want To Be Bothered With Configuring Git _Again_" stance of Git is a big, gigantic source of (totally avoidable) friction.

kaste:
Okay, had some time to do this today.  It now looks like

image


> right when they just need to Get Stuff Done

Well yeah, that's a bit the problem here again.  "Get Stuff Done" in an installer is to just install something quickly.  For *new* users to git, this pull-behavior question is impossible to answer. Unlikely they know what "rebase" or "ff-only" is.  For them, it makes the install more stressful and slower.  Should I google this question?  How can I change that later?  And honestly, you can live in git for years and only ever have fast-forward pulls if that's your workflow.  

I think the nag hints git now shows at various places *are* for new users in the first place.  Hints or confirmations are always a bit annoying and disturbing for "pros".  But there are not included and built for you anyway.  

Anyway, rewording helps a lot already.

dscho:
/add relnote bug The installer [no longer claims that "fast-forward or merge" is the default `git pull` behavior](https://github.com/git-for-windows/build-extra/pull/498): The default behavior has changed in Git a while ago, to "fast-forward only".

dscho:
@kaste thank you!

kaste:
Yeah!

goyalyashpal:
> new users (you can't decide the question if you don't know git)

yeahhh, soo true. obviously i too didn't know back then; but have seen at various places in github/gitlab where this behaviour of "merge" on git-pull created _lotsss_ of headaches.

now that i know git-local well, and was just doing the update from v2.39.1 to v2.44.0 now; have changed following things during the install:

(i hadn't changed the symlink one before too, but i guess it wasn't enabled in the windows system developer settings)

```diff
--- git-config-2023
+++ git-config-2024
@@ -1,5 +1,5 @@
 $ git -v
-git version 2.39.1.windows.1
+git version 2.44.0.windows.1
 
 $ git config -l
 diff.astextplain.textconv=astextplain
@@ -8,14 +8,15 @@
 filter.lfs.process=git-lfs filter-process
 filter.lfs.required=true
 http.sslbackend=openssl
-http.sslcainfo=.../Git/mingw64/ssl/certs/ca-bundle.crt
-core.autocrlf=true
+http.sslcainfo=.../Git/mingw64/etc/ssl/certs/ca-bundle.crt
+core.autocrlf=input
 core.fscache=true
-core.symlinks=false
-pull.rebase=false
+core.symlinks=true
+core.editor="C:\\Program Files\\VSCodium\\bin\\codium" --wait
+pull.rebase=true
 credential.helper=manager
 credential.https://dev.azure.com.usehttppath=true
-init.defaultbranch=master
+init.defaultbranch=main
 credential.https://codeberg.org.provider=generic
 core.repositoryformatversion=0
 core.filemode=false

```

PR #499, opened 2023-05-09
Stop assuming that Pacman package names cannot contain -<digit>

dscho:
We currently have a rather big issue, where the Pacman repositories are broken.

The underlying problem is that the Pacman databases are corrupt. To understand why: Pacman databases (i.e. `git-for-windows.db` which live in the Pacman repositories and are synced into `/var/lib/pacman/sync/`) are `.tar` files containing information about the current package versions. This package information is contained in the `.tar` archives as `desc` files stored in directories whose names have the format `<pkgname>-<pkgver>-<pkgrel>` where `pkgver` and `pkgrel` correspond to the definitions in the original `PKGBUILD` file: `pkgver` refers to the upstream package version, while `pkgrel` is an incrementing number (starting with 1) that allows for rebuilding the packages even while the `pkgver` is unchanged.

Now, the issue was that those Pacman repositories were "sanitized", de-duplicating packages, but mistaking `msys2-runtime-3.3-3.3.6-1` and `msys2-runtime-3.3.6-4` to refer to different versions of the same package, `msys2-runtime`. To make matters worse, `msys2-runtime-3.3-devel-3.3.6-1` was _not_ mistaken to be a different version of the same package as `msys2-runtime-devel-3.3.6-4`, which means that e.g. git-sdk-64's `sync` job now consistently [fails](https://github.com/git-for-windows/git-sdk-64/actions/runs/4921830149) with:

```
[...]
warning: terminate other MSYS2 programs before proceeding
resolving dependencies...
looking for conflicting packages...
error: failed to prepare transaction (could not satisfy dependencies)
:: installing msys2-runtime (3.4.6-2) breaks dependency 'msys2-runtime=3.3.6' required by msys2-runtime-devel
[...]
```

This PR fixes this, and all related problems that are also caused by my original incorrect assumption about Pacman package naming.

I might have missed some, or done some incorrectly, though, so I appreciate _every_ thorough review.


dscho:
> We currently have a rather big issue, where the Pacman repositories are broken.
> 
> The underlying problem is that the Pacman databases are corrupt.

FWIW I have fixed this manually in https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4930413416/jobs/8811298893.

This PR is _still_ kind of semi-urgent because it's blocking any further package deployment in Git for Windows (unless we want to see the same breakage happen again 😃)

jeffhostetler (approved):
I think this look fine.

make-file-list.sh
```diff
@@ -140,6 +140,16 @@ mingw-w64-$PACMAN_ARCH-zstd"
 	return $res
 }
 
+has_pacman_package () {
```
rimrul, line 143: There seem to be no callers for this.

dscho: Gaah, you're right, I forgot to adjust the code below that function to add two intended callers! [Fixed](https://github.com/git-for-windows/build-extra/compare/701c6302d172010b25e414bcb49f12e90bc275a7..6769038c55aed5f067fec4309f96cc1ee0f7bbba)

PR #501, opened 2023-05-14
git-extra: more really only add the aarch64 repository for git-sdk-arm64

PhilipOakley:
This follows up on https://github.com/git-for-windows/git/issues/4408#issuecomment-1538451797.

The referenced commit included additional grepping for installing part of the aarch64 repository which I think isn't necessary here, but I could easily be wrong.

I'm about to go off-line for 4 days, and with -rc0 imminent I thought it worthwhile to at least get started, even if I haven't had time to test it.

257a670d (git-extra: really only add the aarch64 repository for git-sdk-arm64, 2023-02-11) missed two additional places that needed the extra checks.

This fixes issue: #4408 `sdk gitk interaction`

Helped-by: Johannes Schindelin <johannes.schindelin@gmx.de>

### DIFF
--- a/git-extra/git-sdk.sh
+++ b/git-extra/git-sdk.sh
@@ -80,7 +80,7 @@ sdk () {
 			case "$(uname -m)" in
 			i686) shortcut_suffix=" 32-bit";;
 			x86_64)
-				if test -d /clangarm64
+				if test -z "$(find /clangarm64 -type f -print -quit)" || # if /clangarm64 exists and contains at least one file
 				then
 					shortcut_suffix=" arm64"
 				else
@@ -221,7 +221,7 @@ sdk () {
 			MINGW_MOUNT_POINT=/mingw32
 			;;
 		x86_64)
-			if test -d /clangarm64
+			if test -z "$(find /clangarm64 -type f -print -quit)" || # if /clangarm64 exists and contains at least one file
 			then
 				MSYSTEM=CLANGARM64
 				MINGW_MOUNT_POINT=/clangarm64


PhilipOakley:
Hmm. forgot the "Validating source files with sha256sums..." Need to update those checksums. Though don't have time just now.

git-extra/git-sdk.sh
```diff
@@ -80,7 +80,7 @@ sdk () {
 			case "$(uname -m)" in
 			i686) shortcut_suffix=" 32-bit";;
 			x86_64)
-				if test -d /clangarm64
+				if test -z "$(find /clangarm64 -type f -print -quit)" || # if /clangarm64 exists and contains at least one file
```
dscho, line 83: It's probably worth refactoring this into its own function (something like `is_aarch64_setup`).

PhilipOakley: I'll look into that refactoring. 

There was one issue I saw with the code that may be due to my ignorance of some bash idioms. When I copied the change into my `C:\git-sdk-64\etc\profile.d\git-sdk.sh` (where this is installed), I found that I'd get a syntax error at startup, probably due to the way the double quotes and `||` (or) interact with the `#` comment.

In the original https://github.com/git-for-windows/build-extra/commit/257a670d3762817c5e9553edcbfaf41b89d8d47f (git-extra: really only add the aarch64 repository for git-sdk-arm64, 2023-02-11) there is only a single opening 'double quote' for the `"$(find ...)`, which looks 'odd' to me - is that (the possible use of the || as an alternate close quote) an idiom I'm not aware of?

If I 'correct' the conditional to either add an `echo "here"`, or a `:` (colon) just after the `||` (or), then the error disappears (obvs. fixed in two places).

What am I missing?

wayneforrest:
can't wait for this one ... I really would like to have a MINGW64 ARM install.

git-extra/git-sdk.sh
dscho: > I found that I'd get a syntax error at startup, probably due to the way the double quotes and `||` (or) interact with the `#` comment.

No, the reason is that the `||` should not be here. That's a logical "or" operator, and there is no expression to "or it against". You need to remove it.

> In the original [257a670](https://github.com/git-for-windows/build-extra/commit/257a670d3762817c5e9553edcbfaf41b89d8d47f) (git-extra: really only add the aarch64 repository for git-sdk-arm64, 2023-02-11) there is only a single opening 'double quote' for the `"$(find ...)`, which looks 'odd' to me - is that (the possible use of the || as an alternate close quote) an idiom I'm not aware of?

That was a bug, [hence fixed](https://github.com/git-for-windows/build-extra/commit/5b4d8fffcefe79f4960d49df4b59e0700a68d7d4) ("View git blame" can help you find such answers).

dscho:
This PR is most likely obsolete, and even if not, work would have to be put into it. Let's just close it instead.

PR #512, opened 2023-07-22
Implemented Git Bash support for systems with mandatory ASLR enabled

AndyBorah38:
Implemented Git Bash support for systems with mandatory ASLR enabled by modifying the post installation Batch script to add mandatory ASLR security exceptions for the executables in the "usr/bin" directory.

Fixes:
- https://github.com/git-for-windows/git/issues/1412
- https://github.com/mintty/wsltty/issues/6#issuecomment-415961012

### DIFF
--- a/post-install.bat
+++ b/post-install.bat
@@ -42,6 +42,47 @@
 	usr\bin\dash.exe -c '/usr/bin/dash usr/bin/rebaseall -p'
 )
 
+@REM Checks if this script is running with administrative privileges and if not ask the user if they wish to attempt privilege escalation using a message box.
+@REM This is required in case this script was called by the portable installer.
+@REM
+@REM Then if running with administrative privileges adds mandatory ASLR security exceptions for the executables in the "usr/bin" directory.
+@REM This is required for Git Bash to work when running on a Windows system with mandatory ASLR enabled.
+@REM
+@REM Mandatory ASLR is a Windows security feature that is disabled by default.
+@REM https://learn.microsoft.com/microsoft-365/security/defender-endpoint/exploit-protection-reference?view=o365-worldwide#force-randomization-for-images-mandatory-aslr
+@REM
+@REM Doing this significantly slows down the load time of the program settings list in the exploit protection section of the Windows security application but it will load eventually.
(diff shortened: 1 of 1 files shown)


rimrul:
Do we still need ASLR exceptions with 64 Bit Git for Windows 2.41.0 and newer? The Cygwin 3.4.* based Msys2 runtime [should work with full ASLR](https://cygwin.com/pipermail/cygwin-announce/2022-December/010821.html).

https://github.com/git-for-windows/git/issues/1412#issuecomment-355568038

> Please, and I must reiterate, do not automatically deactivate, disable, or otherwise circumvent security options users or administrators have set up.

This user does have a point, we shouldn't just add security exceptions without asking when run as an administrator.

AndyBorah38:
> Do we still need ASLR exceptions with 64 Bit Git for Windows 2.41.0 and newer? The Cygwin 3.4.* based Msys2 runtime [should work with full ASLR](https://cygwin.com/pipermail/cygwin-announce/2022-December/010821.html).
> 
> [git-for-windows/git#1412 (comment)](https://github.com/git-for-windows/git/issues/1412#issuecomment-355568038)
> 
> > Please, and I must reiterate, do not automatically deactivate, disable, or otherwise circumvent security options users or administrators have set up.
> 
> This user does have a point, we shouldn't just add security exceptions without asking when run as an administrator.

Yes we still need ASLR exceptions with 64 Bit Git for Windows 2.41.0 and newer or at least I need them on my Windows 10 Pro x64 machine and I didn't consider that but I guess it would be proper to always prompt the user unless they supply the silent flags.

I'll get on editing it as soon as possible.

dscho:
> The Cygwin 3.4.* based Msys2 runtime [should work with full ASLR](https://cygwin.com/pipermail/cygwin-announce/2022-December/010821.html).

See msys2/msys2-runtime@a556be71db77b88ddf4121d431ee860faa1be789, though.

rimrul:
> See msys2/msys2-runtime@a556be71db77b88ddf4121d431ee860faa1be789, though.

I forgot about that. That would explain it.

PR #515, opened 2023-08-06
Installer: detect manually-configured default branch

dscho:
As reported in https://github.com/git-for-windows/git/issues/4525, during upgrades the Git for Windows installer may overwrite the current default branch name if it has been set manually in the system config (as opposed to configuring it in the previous Git for Windows installer run).

This PR addresses that by letting the system config override the values from the previous installer run.


rimrul (approved):

dscho:
/add relnote bug When `init.defaultBranch` is changed manually in the system config, subsequent Git for Windows upgrades [would overwrite that change](https://github.com/git-for-windows/git/issues/4525). This has been [fixed](https://github.com/git-for-windows/build-extra/pull/515).

PR #517, opened 2023-08-09
install.iss: when silent, do not prompt for process closure

derrickstolee:
This block that is being edited is already within an "if WizardSilent()" block, so we should not be prompting via a message box.

Instead, fail with a log message indicating the problem.

The diff looks a bit nasty since we are removing a retry loop at the same time. Hide whitespace changes to make it more reasonable.

---

I still need to generate an installer with this change so I can test its behavior in silent mode.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -2566,30 +2566,25 @@ begin
                         Log(Msg);
                     ExitEarlyWithSuccess();
                 end;
-                if WizardSilent() then
-                    while True do begin
-                        Msg:='';
-                        for j:=i to GetArrayLength(Processes)-1 do
-                            if not Processes[j].Restartable then
-                                Msg:=Msg+#13+#10+Processes[j].Name+' (PID '+IntToStr(Processes[j].ID)+')'
-                            else if (Processes[j].ToTerminate) and (not TerminateProcessByID(Processes[i].ID)) then
-                                Msg:=Msg+#13+#10+Processes[j].Name+' (PID '+IntToStr(Processes[j].ID)+')';
+                if WizardSilent() then begin
+                    Msg:='';
+                    for j:=i to GetArrayLength(Processes)-1 do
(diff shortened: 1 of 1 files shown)


rimrul:
Why do we need this? The code in question was already using `SuppressibleMsgBox()`, so `/VERYSILENT /SUPPRESSMSGBOXES` should already make this not prompt.

derrickstolee:
> TBH I am a bit puzzled by that code. It looks like it wants to go through a loop asking the user to terminate the applications, until all applications are terminated, but only in _silent_ mode? I'll want to do a bit of digging to figure out whether the intention was actually to write `if not WizardSilent()` instead.

The `else` case includes this message: `'Setup cannot continue until you close at least those applications in the list that are marked as "closing is required".` This makes me think that in the Wizard we have a list presented in the window and the "retry" is simply to hit "next" on the wizard again.

> In any case, you can test this code by running `./release.sh -d Processes` in `installer/`, which will build a custom installer to specifically test only the "Processes" page.

Thanks! I'll give this a try.

derrickstolee:
I was able to make and test the manifest update and it in fact does not require a change here in `git-for-windows/build-extra`. The installation logs include this message (and only once):

```
2023-08-09 13:18:29.491   Defaulting to Cancel for suppressed message box (Retry/Cancel):
                          The following process(es) use Git for Windows:
                          
                          bash.exe (PID 17664)
                          ssh-agent.exe (PID 6948)
                          MSYS2 terminal (PID 23020)
                          bash.exe (PID 14908)
                          ssh-agent.exe (PID 15748)
                          bash.exe (PID 2208)
                          ssh-agent.exe (PID 17104)
                          winpty.exe (PID 9088)
                          
                          Please terminate those processes and retry.
                          Alternatively, cancel to abandon setup altogether.
```

Thanks for the pointers in the right direction, @rimrul!

PR #521, opened 2023-08-22
signtool.sh: add verbose logging to osslsigncode

dennisameling:
In the issue mentioned below, we've been investigating a problem where code signing _sometimes_ hangs on arm64, but not consistently. It seems to hang somewhere around the osslsigncode.exe process. Let's add some verbose logging there to see if it helps us in the debugging process somehow...

Ref: https://github.com/git-for-windows/git/issues/3107#issuecomment-1634695759


dscho (approved):
Thank you!

PR #522, opened 2023-08-24
installer: fix the broken "Enable Pseudo Console Support" option

dscho:
As reported in https://github.com/git-for-windows/git/issues/4571, the Pseudo Console support checkbox does the exact opposite of what it is intended to do.

Turns out that this bug has been with us since Git for Windows v2.41.0, i.e. since June 1st 2023 (or, if you count the pre-releases, since since May 17th, 2023).

This PR fixes the logic and adds a check to prevent regressions in the future.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -186,7 +186,7 @@ jobs:
         shell: pwsh
         run: |
           $exePath = Get-ChildItem -Path installer-${{ matrix.bitness }}/*.exe | %{$_.FullName}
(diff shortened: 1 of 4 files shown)


dscho:
/add release note bug The installer option to enable support for Pseudo Consoles [has been handled incorrectly](https://github.com/git-for-windows/git/issues/4571) since Git for Windows v2.41.0, which [has been fixed](https://github.com/git-for-windows/build-extra/pull/522).

Serenitychic:
gh pr checkout 522

PR #524, opened 2023-08-24
Fix the PR build installing and validating the 32-bit flavor of Git for Windows

dscho:
Over in https://github.com/git-for-windows/build-extra/pull/522, I was implementing a relatively straight-forward bug fix and added an extra precaution to verify via `run-checklist.sh` that the installer works as intended.

Guess how big my surprise was when not only the [`sdk-artifacts (build-installers, 32)` job failed](https://github.com/git-for-windows/build-extra/actions/runs/5961265753/job/16170153237?pr=522#logs) (and _only_ that one, the 64-bit one was fine!), but when I went down a lengthy rabbit hole once I realized that the silent installation of the 32-bit Git for Windows that is part of the validation _failed_, and must have done so ever since c800a3f6004dc0690ec9bcf0ef79f696f53eeb66, i.e. for over half a year!

Turns out that we did not, in fact, validate what we thought we were validating. In many, entertainingly different ways did we not do that.

This PR reflects some of my journey down that particular rabbit hole and I hope you enjoy the ride.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -186,7 +186,7 @@ jobs:
         shell: pwsh
         run: |
           $exePath = Get-ChildItem -Path installer-${{ matrix.bitness }}/*.exe | %{$_.FullName}
-          Start-Process -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART /SUPPRESSMSGBOXES /ALLOWDOWNGRADE=1 /LOG=installer.log"
+          Start-Process -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART /SUPPRESSMSGBOXES /ALLOWDOWNGRADE=1 /ALLOWINSTALLING32ON64=1 /LOG=installer.log"
           "$env:ProgramFiles\Git\usr\bin" | Out-File -Encoding ascii -Append $env:GITHUB_PATH
           "$env:ProgramFiles\Git\mingw${{env.ARCH_BITNESS}}\bin" | Out-File -Encoding ascii -Append $env:GITHUB_PATH
       - name: validate
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -203,6 +203,7 @@ jobs:
           echo '::endgroup'
 
           set -x &&
+          grep 'Installation process succeeded' installer.log &&
           cygpath -aw / &&
           git.exe version --build-options >version &&
           cat version &&
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -193,6 +193,10 @@ jobs:
         if: matrix.artifact == 'build-installers'
         shell: bash
         run: |
+          echo '::group::installer.log'
+          cat installer.log
+          echo '::endgroup'
+
           set -x &&
           cygpath -aw / &&
           git.exe version --build-options >version &&
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -186,7 +186,12 @@ jobs:
         shell: pwsh
         run: |
           $exePath = Get-ChildItem -Path installer-${{ matrix.bitness }}/*.exe | %{$_.FullName}
-          Start-Process -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART /SUPPRESSMSGBOXES /ALLOWDOWNGRADE=1 /ALLOWINSTALLING32ON64=1 /LOG=installer.log"
+          $installer = Start-Process -PassThru -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART /SUPPRESSMSGBOXES /ALLOWDOWNGRADE=1 /ALLOWINSTALLING32ON64=1 /LOG=installer.log"
+          $exitCode = $installer.ExitCode
+          if ($exitCode -ne 0) {
+            Write-Host "::error::Installer failed with exit code $exitCode!"
+            exit 1
+          }
           "$env:ProgramFiles\Git\usr\bin" | Out-File -Encoding ascii -Append $env:GITHUB_PATH
           "$env:ProgramFiles\Git\mingw${{env.ARCH_BITNESS}}\bin" | Out-File -Encoding ascii -Append $env:GITHUB_PATH
       - name: validate
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -204,6 +204,7 @@ jobs:
 
           set -x &&
           grep 'Installation process succeeded' installer.log &&
+          ! grep -iw failed installer.log &&
           cygpath -aw / &&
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -195,7 +197,7 @@ jobs:
           cygpath -aw / &&
           git.exe version --build-options >version &&
           cat version &&
-          grep "$(sed -e 's|^v||' -e 's|-|.|g' <bundle-artifacts/next_version)" version &&
+          grep "${{ steps.build-artifact.outputs.git-version }}" version &&
```
jeffhostetler, line 216: on line 198 we have `--build-options` and up where the `...output.git-version` is set we do not use that option.  granted this does not matter since the grep will match.

should we consider using `--build-options` in both places and then verify both the version string *and* the `build from commit` lines ??

While at it, we could also verify that the `cpu` line matches the expected value from the matrix.  (This might be overkill.)

.github/workflows/main.yml
```diff
@@ -186,7 +186,12 @@ jobs:
         shell: pwsh
         run: |
           $exePath = Get-ChildItem -Path installer-${{ matrix.bitness }}/*.exe | %{$_.FullName}
-          Start-Process -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART /SUPPRESSMSGBOXES /ALLOWDOWNGRADE=1 /ALLOWINSTALLING32ON64=1 /LOG=installer.log"
+          $installer = Start-Process -PassThru -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART /SUPPRESSMSGBOXES /ALLOWDOWNGRADE=1 /ALLOWINSTALLING32ON64=1 /LOG=installer.log"
+          $exitCode = $installer.ExitCode
+          if ($exitCode -ne 0) {
+            Write-Host "::error::Installer failed with exit code $exitCode!"
+            exit 1
+          }
```
jeffhostetler, line 194: Good grief!  How obscure!  Good find.

jeffhostetler (approved):
A few nits, but looks good.

.github/workflows/main.yml
dscho: @jeffhostetler Right. We could change how it was done previously (which _is_ where this paradigm is coming from, grepping for the version instead of requiring an exact match of the output with an expected result).

However, there are two reasons why I don't want to do that here:

- The purpose of this PR is to fix the brokenness in a minimal fashion. Before, we grepped for the empty string, which was incorrect. Now we grep for the version, which is correct.
- This code is actually not really stand-alone. It was copy/edited [from `git-sdk-32`](https://github.com/git-for-windows/git-sdk-32/blob/d818936674d95a984af49beaec8fb55a701cd068/.github/workflows/git-artifacts.yml#L219-L231) and that is where I need to make follow-up changes because I want to keep both workflows somewhat in sync.

dscho: Oh, and the `cat version` is trying to help debugging, that's also the reason why `--build-options` is used there.

.github/workflows/main.yml
```diff
@@ -192,8 +192,13 @@ jobs:
...
             exit 1
           }
-          "$env:ProgramFiles\Git\usr\bin" | Out-File -Encoding ascii -Append $env:GITHUB_PATH
-          "$env:ProgramFiles\Git\mingw${{env.ARCH_BITNESS}}\bin" | Out-File -Encoding ascii -Append $env:GITHUB_PATH
+          if ("${{ matrix.bitness }}" -eq 32) {
+            "${env:ProgramFiles(x86)}\Git\usr\bin" | Out-File -Encoding ascii -Append $env:GITHUB_PATH
+            "${env:ProgramFiles(x86)}\Git\mingw32\bin" | Out-File -Encoding ascii -Append $env:GITHUB_PATH
+          } else {
+            "$env:ProgramFiles\Git\usr\bin" | Out-File -Encoding ascii -Append $env:GITHUB_PATH
+            "$env:ProgramFiles\Git\mingw${{env.ARCH_BITNESS}}\bin" | Out-File -Encoding ascii -Append $env:GITHUB_PATH
```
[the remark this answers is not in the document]
dscho: > I'm not sure what these command lines do.

@jeffhostetler In PowerShell, when you call `echo 1 >>out.txt`, the appended text is encoded in UTF-16, something that GitHub Actions are distinctly unable to handle in `GITHUB_PATH`.

Therefore I pipe the to-be-appended lines to the `Out-File` command that can be asked to use ASCII encoding.

> [...] do we now know here that the bitness will be 64 and could replace the mingw${{env.ARCH_BITNESS}} with just mingw64 for clarity ??

It is still my hope that we will support Windows/ARM64 in the near future, in which case even the `mingw` part needs to be re-thought: the Windows/ARM64 bits we're building live in `/clangarm64/`.

So today I want to just leave this stuff unchanged (apart from indentation), and cross that particular bridge only once I reach it.

dscho:
> A few nits, but looks good.

Thank you for looking over it, and poking at iffy spots!

rimrul (approved):
One nit about `479d33f`: The bug isn't that we need to use `-PassThru`, That is expected and [documented](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/start-process?view=powershell-7.3#-passthru) (though unintuitive) behaviour. The bug is that without `-Wait` the `ExitCode`property isn't populated.

dscho:
> One nit about `479d33f`: The bug isn't that we need to use `-PassThru`, That is expected and [documented](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/start-process?view=powershell-7.3#-passthru) (though unintuitive) behaviour. The bug is that without `-Wait` the `ExitCode`property isn't populated.

Good point.

<details><summary>Range-diff</summary>



* 1:  479d33fcd ! 1:  2f1c99859 PR build (run the installer): exit with error if the installer failed

   ``````diff
   @@ Commit message
        Let's check explictly if the installer failed.
    
        Note: We cannot simply use `$?` but need to look at the `ExitCode`
   -    attribute of the process returned by `Start-Process`, _and_ we need to
   -    pass `-PassThru` for this to work, due to a bug in `Start-Process`. See
   +    attribute of the process returned by `Start-Process` _and_ the `-Wait`
   +    option for this to work, due to a bug in `Start-Process` where the
   +    `ExitCode` attribute is not populated properly without `-Wait`. And to
   +    get to the `ExitCode` we naturally need the `-PassThru` option, too. See
        https://stackoverflow.com/a/16018287/1860823 for a more complete
        explanation.
    
   ``````

* 2:  f5110ccfc = 2:  9526d1850 PR build (run the installer): verify that the installer claims success
* 3:  41d121eff = 3:  92c2f9aa5 installer: avoid trying to copy a non-existent file
* 4:  74a61f135 = 4:  44a9c8985 installer: adjust a misleading log message
* 5:  209fa84bc = 5:  d034bbbcc PR build (run the installer): report any failure of any kind
* 6:  93cd487de = 6:  8e51e6565 PR build (run the installer): do validate the correct installation

</details>

PR #526, opened 2023-08-25
7-Zip: update SFX component to match 7-Zip 23.01

dscho:
Update not only "the SFX component", but actually add another one, too. The reason is that `7zSD.sfx`, the component we used to use all the time, is supposed to be linking to Microsoft's C runtime dynamically (the "D" stands for "DLL" or "dynamic").

And it used to be the case, right until we fixed that bug that was reported in https://github.com/git-for-windows/git/issues/1127, where the runtime we linked against was `vcruntime140.dll`, which is unfortunately not ubiquituous.

We then fixed that via the commit at https://github.com/git-for-windows/7-Zip/commit/3ba0e05a95b5 that was unfortunately incorrect: all it did was change the `7zSD.sfx` from linking dynamically to the C runtime to linking statically. But that' exactly the only difference between `7zSD.sfx` and `7zS.sfx`.

Undoing that mistake, I used the opportunity of the sync with 7-Zip v23.01 to build `7zSD.sfx` so it links dynamically again. Since we need the statically-linked version, I also built `7zS.sfx` and copied it into `build-extra`, and then updated all the locations where we used `7zSD.sfx` by mistake.

These artifacts come straight from https://github.com/git-for-windows/7-Zip/actions/runs/5977124234#artifacts, built from the commit c77d78f7 (Merge branch 'readme', 2023-08-25) in the `v32.01-VS2022-sfx` branch of https://github.com/git-for-windows/7-Zip.


dscho:
@rimrul I wanted to wait with this until the v2.42.0 release(s) settled down, and think that now would be a good time. Do you concur?

rimrul (approved):

PR #535, opened 2023-11-22
Update git-update-git-for-windows

MPGaby23:
Need to replace //SILENT command with //VERYSILENT in order to make the update silent.

### DIFF
--- a/git-extra/git-update-git-for-windows
+++ b/git-extra/git-update-git-for-windows
@@ -346,7 +346,7 @@ update_git_for_windows () {
 		echo "Downloading $filename" >&2
 	fi
 	curl -# -L -o $installer $download || return
-	start "" "$installer" //SILENT //NORESTART
+	start "" "$installer" //VERYSILENT //NORESTART
 
 	# Kill all Bash processes (which will let MinTTY quit, too)"
 	#


rimrul:
I don't think we want `/VERYSILENT` we explicitly got rid of it in https://github.com/git-for-windows/build-extra/pull/318/commits/9d3660bcb68c8dcbbc0a5e7db556ac1442f83f2a

git-extra/git-update-git-for-windows
```diff
@@ -346,7 +346,7 @@ update_git_for_windows () {
 		echo "Downloading $filename" >&2
 	fi
 	curl -# -L -o $installer $download || return
-	start "" "$installer" //SILENT //NORESTART
+	start "" "$installer" //VERYSILENT //NORESTART
```
dscho, line 349: As pointed out by @rimrul, this contradicts https://github.com/git-for-windows/build-extra/commit/9d3660bcb68c8dcbbc0a5e7db556ac1442f83f2a.

When essentially reverting a previous commit, it is _particularly_ important to make a good, convincing case that it should be reverted. This is typically done in the commit message. For guidance how to improve the commit message substantially, see https://github.blog/2022-06-30-write-better-commits-build-better-projects/.

Also, instead of leaving the comment about the [DCO](https://git-scm.com/docs/SubmittingPatches#dco) in the PR template, it would have made sense to heed the advice (i.e. to sign off on the commit) and delete the advice...

dscho (changes requested):
In the current form, this PR would revert a previous change that was made on purpose, and therefore cannot be merged.

However, I suspect that there may be good reasons to introduce a command-line option to make the installation "very silent". This could not be the default, but would potentially come in handy for users wishing to automate the updates in an even less-intrusive way.

git-extra/git-update-git-for-windows
dscho: @MPGaby23 are you prepared to make the changes we requested?

PR #548, opened 2024-02-27
getgit: flag for installation without modifying the environment

LunarLambda:
passing `-e` has the following effects:

1. `/etc/profile.d` is not processed.
2. The context menu scripts `ctxmenu.bat` and `rm-ctxmenu.bat` are not created.
3. `/cmd` is not added to the path, programs in `/cmd` are installed to `[/usr]/bin` instead

### DIFF
--- a/git-extra/getgit
+++ b/git-extra/getgit
@@ -13,6 +13,7 @@ function usage(){
     echo "  -k                   keep all downloaded, extracted files in \"/tmp\""
     echo "  -s                   skip downloading"
     echo "  -f                   forcibly update no matter what version it is"
+    echo "  -e                   don't modify the environment (no /cmd in path, /etc/profile.d)"
     echo "  -h                   this prints me..."
     exit
 }
@@ -108,6 +109,8 @@ until [ $# -eq 0 ]; do
             force_update=1; shift;;
         -s)    
             skip_download=1; shift;;
+        -e)
+            no_env=1; shift;;
         -h)
             usage; shift;;    
         *) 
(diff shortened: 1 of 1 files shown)


git-extra/getgit
```diff
@@ -274,4 +284,4 @@ fi
 echo "Git is now available! If you want to add \"Git GUI Here\" and \"Git Bash"
```
rimrul, line 284: These messsges should probably be modified if the context menu scripts aren't created.

git-extra/getgit
```diff
@@ -13,6 +13,7 @@ function usage(){
     echo "  -k                   keep all downloaded, extracted files in \"/tmp\""
     echo "  -s                   skip downloading"
     echo "  -f                   forcibly update no matter what version it is"
+    echo "  -e                   don't modify the environment (no /cmd in path, /etc/profile.d)"
```
rimrul, line 16: "no /cmd in path" sounds like that part is only about not modifying `$PATH`, whereas the code doesn't create `/cmd` at all.

git-extra/getgit
```diff
@@ -188,72 +191,79 @@ process "/mingw$bit/share/git-core"
 process "/mingw$bit/share/git-gui"
 process -f "/mingw$bit/share/perl5"
 process -f "/mingw$bit/ssl"
-process "/cmd"
-process -f "/etc/profile.d"
+if [[ ! x$no_env == x1 ]]; then
+    process "/cmd"
+    process -f "/etc/profile.d"
+else
+    cp -f $gitupk/cmd/* $bin
```
rimrul, line 198: Does it make sense to copy these files to `/usr/bin` or `/bin`? Wouldn't they break MSYS2/Cygwin `git`? And wouldn't they be broken in `/usr/bin` anyways, because they expect `../mingw64/bin/git.exe` etc to exist?

rimrul: Most files in `/cmd` are compiled from [`git-wrapper.c`](https://github.com/git-for-windows/MINGW-packages/blob/main/mingw-w64-git/git-wrapper.c) and can't rely on `PATH` to define `PATH`.

LunarLambda: Ah. Would the PR be acceptable if /cmd is left as is (and not copied to somewhere else) i.e. only skips /etc/profile.d, the registry PATH setting, and the context menu scripts? I can add a message informing the user they need to add /cmd to path themselves.

dscho:
Let's not pursue this anymore but instead rally what resources we have to https://github.com/msys2/MINGW-packages/discussions/16383, which would accomplish the same goal, except better.

PR #150, opened 2017-08-09
installer: Add registry for Windows Libraries

YueLinHo:
See https://github.com/git-for-windows/git/issues/1229

PR #551, opened 2024-04-01
installer: delete context menu entries created for library folder on uninstall

nshcr:
A PR(#150) from a long time ago added some context menu entries for the Windows library folder but did not include scripts for their removal, leading to residual registry data after the software was uninstalled.

This PR is intended to address that issue.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -566,11 +566,13 @@ begin
         RootKey:=HKEY_CURRENT_USER;
     end;
 
-    SetArrayLength(Keys,4);
+    SetArrayLength(Keys,6);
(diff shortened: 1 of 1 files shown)


dscho (approved):
❤️

dscho:
/add relnote bug Since v2.14.0(2), Git for Windows' installer registers the _Open Git Bash here_ and _Open Git GUI here_ context menu items also in the special [Libraries folders](https://msdn.microsoft.com/en-us/library/windows/desktop/dd758096.aspx), but the uninstaller never removed them from those folders, [which was fixed](https://github.com/git-for-windows/build-extra/pull/551).

khanomiposhak-tech:
TANK YOU

PR #553, opened 2024-04-25
Update aliases judgment for winpty

250h:
Mintty can change their type of $TERM easily in settings, it's better use `disable_pcon` to get winpty aliases.

Signed-off-by: E. H [94212663+250h@users.noreply.github.com](mailto:94212663+250h@users.noreply.github.com)

### DIFF
--- a/git-extra/aliases.sh
+++ b/git-extra/aliases.sh
@@ -5,11 +5,11 @@
 alias ls='ls -F --color=auto --show-control-chars'
 alias ll='ls -l'
 
-case "$TERM" in
-xterm*)
+case "$MSYS" in
+*disable_pcon*)
 	# The following programs are known to require a Win32 Console
 	# for interactive usage, therefore let's launch them through winpty
-	# when run inside `mintty`.
+	# when disable pseudo console with `disable_pcon`.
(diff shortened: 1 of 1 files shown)


dscho:
> Signed-off-by: E. H [94212663+250h@users.noreply.github.com](mailto:94212663+250h@users.noreply.github.com)

Please sign off the _commit_, not the PR: Add your sign-off to the commit message. But if you use a pseudonym and a non-working email address, then it is not actually a sign-off.

git-extra/aliases.sh
```diff
@@ -5,11 +5,11 @@
 alias ls='ls -F --color=auto --show-control-chars'
 alias ll='ls -l'
 
-case "$TERM" in
-xterm*)
+case "$MSYS" in
+*disable_pcon*)
```
dscho, line 9: The trouble with this is: by default `MSYS` is _unset_, and Pseudo Console support is _still_ disabled. The new check misses that rather common scenario, and therefore will break them.

I did not find a way to reliably determine whether Pseudo Console support is enabled. We may need to patch the MSYS2 runtime to be able to tell, first.

PR #555, opened 2024-05-17
portable: make sure that `package-versions.txt` is included

dscho:
In https://github.com/git-for-windows/git/issues/4955 it was pointed out that the i686 version ("32-bit") of Portable Git v2.45.1 is missing the `package-versions.txt` file.

While it is unclear why that happened (the logs of the -- embargoed -- build do not reveal any clue about this issue), let's at least ensure that it does not happen again.

/cc @mcgitty


dscho:
@mcgitty any chance that you can review this?

mcgitty:
I see. It will force an investigation if it's missing in future builds.

mcgitty (commented):
I see. It will force an investigation if it's missing in future builds.

PR #558, opened 2024-06-06
getgit: Fix user path reg query result potentially getting truncated

Morilli:
This reg query result for the user path environment variable could potentially get truncated if a space was contained anywhere inside the path variable. This is especially problematic because this incorrect value is then written back, effectively destroying part of the users path.

For example with path containing `D:\important\entry;C:\my\very            spacey\path`:
Before: `reg query "HKEY_CURRENT_USER\Environment" //v Path 2>/dev/null | awk 'NR==3 {print $NF}'` -> `spacey\path`

After: `reg query "HKEY_CURRENT_USER\Environment" //v Path 2>/dev/null | awk 'NR==3 {print}' | cut -d ' ' -f 13-` -> `D:\important\entry;C:\my\very            spacey\path`

### DIFF
--- a/git-extra/getgit
+++ b/git-extra/getgit
@@ -248,7 +248,7 @@ echo "be removed from the context menu is available as $rm_ctxmenu "
 echo ""
 # Setting ENV
 echo -n "Setting env ..."
-userpath=$(reg query "HKEY_CURRENT_USER\Environment" //v Path 2>/dev/null | awk 'NR==3 {print}' | cut -d ' ' -f 13-)
+userpath="$(powershell -command '(Get-ItemProperty -Path 'HKCU:\Environment' -Name Path).Path')"
 if [[ "$userpath" == "" ]]; then
     setx PATH "$(cygpath -m /cmd)" > /dev/null 2>&1
 elif [[ "$userpath" != *"$(cygpath -m /cmd)"* ]]; then
--- a/git-extra/getgit
+++ b/git-extra/getgit
@@ -248,7 +248,7 @@ echo "be removed from the context menu is available as $rm_ctxmenu "
 echo ""
 # Setting ENV
 echo -n "Setting env ..."
-userpath=$(reg query "HKEY_CURRENT_USER\Environment" //v Path 2>/dev/null | awk 'NR==3 {print $NF}')
+userpath=$(reg query "HKEY_CURRENT_USER\Environment" //v Path 2>/dev/null | awk 'NR==3 {print}' | cut -d ' ' -f 13-)
 if [[ "$userpath" == "" ]]; then
     setx PATH "$(cygpath -m /cmd)" > /dev/null 2>&1
 elif [[ "$userpath" != *"$(cygpath -m /cmd)"* ]]; then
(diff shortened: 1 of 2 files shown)


dscho:
> image

[Oops](https://github.com/git-for-windows/git-for-windows-automation/pull/82).

git-extra/getgit
```diff
@@ -248,7 +248,7 @@ echo "be removed from the context menu is available as $rm_ctxmenu "
 echo ""
 # Setting ENV
 echo -n "Setting env ..."
-userpath=$(reg query "HKEY_CURRENT_USER\Environment" //v Path 2>/dev/null | awk 'NR==3 {print $NF}')
+userpath=$(reg query "HKEY_CURRENT_USER\Environment" //v Path 2>/dev/null | tail -n +3 | awk '{$1=$2=""; print substr($0,3)}')
```
dscho, line 251: @Morilli would you mind explaining this a bit more in the commit message? Feel free to follow [this guidance](https://github.blog/2022-06-30-write-better-commits-build-better-projects/):
image

Morilli: Have force pushed with a full explanation.

git-extra/getgit
```diff
@@ -248,7 +248,7 @@ echo "be removed from the context menu is available as $rm_ctxmenu "
 echo ""
 # Setting ENV
 echo -n "Setting env ..."
-userpath=$(reg query "HKEY_CURRENT_USER\Environment" //v Path 2>/dev/null | awk 'NR==3 {print $NF}')
+userpath=$(reg query "HKEY_CURRENT_USER\Environment" //v Path 2>/dev/null | awk 'NR==3 {print}' | cut -d ' ' -f 13-)
```
dscho, line 251: Would the following work, too?

```sh
reg query "HKEY_CURRENT_USER\Environment" //v Path | sed -n 's/^ *Path *REG_EXPAND_SZ *//p'
```

I would have a preference for that if it worked, because:
1. It avoids spawning _two_ processes, and
2. It spells out what is expected instead of relying on the magic row number 3 and the magic column number 13.

Morilli: Looks like that doesn't work all the time because the path variable may be of type `REG_SZ` or `REG_EXPAND_SZ`, depending on whether or not variables like `%USERPROFILE%` are used (see https://learn.microsoft.com/en-us/windows/win32/sysinfo/registry-value-types for registry value types). I *assume* those are the only valid types for that key that the system accepts, although `REG_MULTI_SZ` also sounds possible to me. If you prefer the `sed` approach, `reg query "HKEY_CURRENT_USER\Environment" //v Path | sed -n 's/^ *Path *REG\(_EXPAND\)\?_SZ *//p'` has worked for both `REG_SZ` and `REG_EXPAND_SZ` in my testing.

It's a bit unfortunate that the `reg query` output is inherently badly formatted for automated reading, but I think the assumptions that have worked so far are safe (value always on line 3, separators are exactly 4 spaces...).

dscho: > It's a bit unfortunate that the `reg query` output is inherently badly formatted for automated reading, but I think the assumptions that have worked so far are safe (value always on line 3, separators are exactly 4 spaces...).

It might be better to turn this into a PowerShell scriptlet, anyway. Something like this:

```sh
userpath="$(powershell -command '(Get-ItemProperty -Path 'HKCU:\Environment' -Name Path).Path')"
```

What do you think?

PR #564, opened 2024-07-02
Reduce redundancy/clarify Win 7/8/32-bit deprecation notices.

pkasting:
Version 2.44 and 2.45 both claimed to be the "last version" to support Win 7/8. Fix by changing the 2.44 text to match earlier releases.

Remove duplicate deprecation warnings from point release notes, since these aren't really "changes since the last release".


dscho (approved):
Thank you!

PR #566, opened 2024-07-04
Let `pacman-helper.sh` run in Linux

dscho:
The idea is of course to help https://github.com/git-for-windows/git-for-windows-automation/pull/86

### DIFF
--- a/pacman-helper.sh
+++ b/pacman-helper.sh
@@ -32,10 +32,6 @@ die () {
 fifo_find="/var/tmp/disowned.find"
 fifo_pacman="/var/tmp/disowned.pacman"
 
-# MSYS2's mingw-w64-$arch-ca-certificates seem to lag behind ca-certificates
-CURL_CA_BUNDLE=/usr/ssl/certs/ca-bundle.crt
-export CURL_CA_BUNDLE
-
 mode=
 case "$1" in
 fetch|add|remove|push|files|dirs|orphans|push_missing_signatures|file_exists|lock|unlock|break_lock|quick_add|sanitize_db)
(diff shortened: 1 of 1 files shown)


pacman-helper.sh
```diff
@@ -57,7 +53,16 @@ upload)
 	;;
 esac
 
-this_script_dir="$(cygpath -am "${0%/*}")"
+this_script_dir="$(case "$(uname -s)" in
+	MSYS|MINGW*)
+		# MSYS2's mingw-w64-$arch-ca-certificates seem to lag behind ca-certificates
+		CURL_CA_BUNDLE=/usr/ssl/certs/ca-bundle.crt
+		export CURL_CA_BUNDLE
```
rimrul, line 60: Doesn't the `$()` command substitution run in a subshell? Wouldn't that make the export useless?

dscho: The `export` does not run in a subshell, but its effect extends to all subshells, so it's very useful: every `curl` call in this script, as well as in the `wingit-snapshot-helper.sh` script, will now heed that path.

rimrul: Isn't this whole hunk one big subshell?

dscho: Ouch. You're right. Let me fix that.

dscho: How does [this](https://github.com/git-for-windows/build-extra/compare/ae824da594a500309b3c67cfe30df044003f369b..78ef4e3cda962813a5520f76403b4e189fd89676) look to you, @rimrul?

rimrul (approved):

PR #571, opened 2024-07-18
installer/portable: exclude `sshd`

dscho:
The OpenSSH daemon (i.e. the server-side) is not actually something Git for Windows needs per se, nor wants to support. So let's stop shipping it.

### DIFF
--- a/make-file-list.sh
+++ b/make-file-list.sh
@@ -194,6 +194,7 @@ grep -v -e '\.[acho]$' -e '\.l[ao]$' -e '/aclocal/' \
 	-e '^/usr/lib/python' -e '^/usr/lib/ruby' -e '^/\(mingw\|clang\)[^/]*/lib/python' \
 	-e '^/usr/share/subversion' \
 	-e '^/etc/skel/' -e '^/\(mingw\|clang\)[^/]*/etc/skel/' \
+	-e '^/etc/sshd_config' \
 	-e '^/usr/bin/svn' \
 	-e '^/usr/bin/xml.*exe$' \
 	-e '^/usr/bin/xslt.*$' \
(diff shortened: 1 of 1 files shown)


mjcheetham (approved):

dscho:
/add relnote feature The server-side component of OpenSSH, which had been shipped with Git for Windows for historical reasons only, is [now no longer distributed with it](https://github.com/git-for-windows/build-extra/pull/571).

bzanin-wdc:
I understand [the logic for this change](https://lore.kernel.org/git/a658fd0a-59bd-c162-874c-cc5b9926acd5@gmx.de/T/#t) and in general I support it, but I should mention that Git for Windows shipping `sshd` has created the opportunity to work around some very thorny issues at `$JOB`.

In particular we sometimes encounter workflows where developers investigate CI failures on specific worker nodes where prototype hardware is attached for full-stack validation, discover tuning or logic issues, craft trial patches in situ, then wish to push their experimental changes back upstream to a feature branch.  They are blocked in this desire for convenience by worker nodes lacking their GPG and SSH keys for signing commits and pushing, respectively; but being able to spin up a one-shot SSH daemon process to briefly support tunneling their local `ssh-agent` and `gpg-agent` sockets to the node have served them well.  Once Git for Windows stops packaging `sshd` this convenience will disappear.

I still support this change, I just wanted to call out that its impact will be non-zero in at least one shop.

dscho:
@bzanin-wdc I did suspect that there were users of `sshd.exe` out there (cf. [Hyrum's law](https://www.hyrumslaw.com/)).

In your instance, would spinning up a [`git daemon`](https://git-scm.com/docs/git-daemon) work as well? After all, the IP seems to be accessible, the ports open, and this scenario is precisely what `git daemon` was designed for. The only difference vs `sshd` is that is unauthenticated.

bzanin-wdc:
> In your instance, would spinning up a [`git daemon`](https://git-scm.com/docs/git-daemon) work as well? After all, the IP seems to be accessible, the ports open, and this scenario is precisely what `git daemon` was designed for. The only difference vs `sshd` is that is unauthenticated.

The lack of authentication would make this option a non-starter, but otherwise it looks like a good match, conceptually.  The hypothetical workflow could be that the developer works on the lab machine for a bit, commits their changes, spins up `git-daemon`, pulls their changes to their personal work machine, amends the commit to GPG-sign it, then pushes it back upstream to the central repo connected to the CI/CD pipeline.

With the authentication step missing and therefore `git daemon` being off the table, _most_ of that workflow could still be performed.  The dev could sub out the `git daemon` step for generating a git bundle, transferring it to their local workstation, then pulling from it, amending/signing the commit, and pushing it back upstream.

It's a lot of steps but it might do the trick.

Thank you for the pointer.

dscho:
> The lack of authentication would make this option a non-starter

You could also go for the opposite direction: starting up a (regular, Win32) OpenSSH server on the developer's machine, add a private key on the CI machine and add its public key to the server's `authorized_keys`, then push rather than pull.

hidromagnetismo:
When I install "Git for Windows SDK v1.0.8" it installs sshd.exe version 9.8p1 (OpenSSH), does this binary have the vulnerability?

dscho:
> sshd.exe version 9.8p1 (OpenSSH), does this binary have the vulnerability?

See https://github.com/git-for-windows/git/issues/5094#issuecomment-2274255080

PR #572, opened 2024-08-08
please.sh: add asciidoctor to build-installers

dennisameling:
[This pipeline](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10159591709/job/28094144095) failed because `asciidoctor` wasn't available. However, it _is_ [available](https://github.com/git-for-windows/git-sdk-arm64/tree/e31bf6358ef639c35716ced851bc98e3bb4e10d4/var/lib/pacman/local/mingw-w64-clang-aarch64-asciidoctor-2.0.23-1) in the `git-sdk-arm64` repo. Looks like it just doesn't end up in the `build-installers` SDK flavor which is used by that pipeline.

In an earlier commit, Git for Windows switched from its own *-asciidoctor-extensions package to the MSYS2-provided *-asciidoctor package.

However, ARM64 pipelines that use the build-artifacts SDK flavor now complain that the asciidoctor binary isn't available.

This commit should ensure that Asciidoctor becomes available.

Ref: https://github.com/git-for-windows/build-extra/commit/eb16e991b8f0bfebc367a4d900585883547032ba

### DIFF
--- a/please.sh
+++ b/please.sh
@@ -3327,6 +3327,14 @@ create_sdk_artifact () { # [--out=<directory>] [--git-sdk=<directory>] [--archit
 
 		# BusyBox
 		$PREFIX/bin/busybox.exe
+
+		# Asciidoctor (requires Ruby to run)
+		$PREFIX/bin/asciidoctor
+		$PREFIX/bin/asciidoctor.bat
+		$PREFIX/lib/ruby/
+		$PREFIX/bin/ruby.exe
+		$PREFIX/bin/ruby*.dll
+
 		EOF
 		mkdir -p "$output_path/.sparse" &&
 		cp "$sparse_checkout_file" "$output_path/.sparse/build-installers"


please.sh
```diff
@@ -3327,6 +3327,14 @@ create_sdk_artifact () { # [--out=<directory>] [--git-sdk=<directory>] [--archit
 
 		# BusyBox
 		$PREFIX/bin/busybox.exe
+
+		# Asciidoctor (requires Ruby to run)
+		$PREFIX/bin/asciidoctor
+		$PREFIX/bin/asciidoctor.bat
+		$PREFIX/lib/ruby/gems/*/gems/asciidoctor-*/
+		$PREFIX/bin/ruby.exe
+		$PREFIX/bin/ruby*.dll
```
dennisameling, line 3336: Looking at the [package details](https://packages.msys2.org/package/mingw-w64-clang-aarch64-asciidoctor?repo=clangarm64), I _think_ this should do the job. Curious to hear if you have any thoughts or concerns about this.

dennisameling: Was able to build `build-installers` locally. Looks like it's still missing a few things:

```
$ asciidoctor --version
`RubyGems' were not loaded.
`error_highlight' was not loaded.
`did_you_mean' was not loaded.
C:/repos/build-installers/clangarm64/bin/asciidoctor:16:in `require': cannot load such file -- rubygems (LoadError)
        from C:/repos/build-installers/clangarm64/bin/asciidoctor:16:in `<main>'
```

dennisameling: Looks like I need to add `$PREFIX/lib/ruby/` (the entire folder) instead of just `$PREFIX/lib/ruby/gems/*/gems/asciidoctor-*/`, so I've just force-pushed to make that change. Would that be OK for you?

dscho: Would you happen to know the difference of the size of the `build-installers` artifact before and after this patch? I am only a bit concerned about making the artifact unnecessarily large.

dscho:
Hmm. I took a deep look at why this is needed at all, when it seems to work all right for the i686 and x86_64 builds. And I think I have the answer.

In https://github.com/git-for-windows/git-sdk-arm64/pull/6, we replaced the AsciiDoctor paths in `.sparse/makepkg-git` in `git-sdk-arm64` with the AsciiDoc equivalents. The reason was that AsciiDoctor (or, more precisely, Ruby) was not available in `git-sdk-arm64`.

However, in the meantime AsciiDoctor did become supported, and we even switched over from AsciiDoc to AsciiDoctor (installing the latter in https://github.com/git-for-windows/git-sdk-arm64/commit/cbf9ddf1864affea6dad93d3c4144d7f8bd0ef75).

Which means that [the AsciiDoc paths](https://github.com/git-for-windows/git-sdk-arm64/blob/425cc1d67d0c25bab5dce51f5630930d6519333c/.sparse/makepkg-git#L55-L64) need to be replaced by the AsciiDoctor paths in `.sparse/makepkg-git` in `git-sdk-arm64` (essentially, `s/mingw64/clangarm64/`, `s/x64-msvcrt-//` and `s/x64-mingw32/aarch64-mingw-ucrt/` on [these lines](https://github.com/git-for-windows/git-sdk-64/blob/HEAD/.sparse/makepkg-git#L63-L89)).

With these changes in git-for-windows/git-sdk-arm64, I believe we can drop this here PR.

dscho:
> With these changes in git-for-windows/git-sdk-arm64, I believe we can drop this here PR.

@dennisameling I have opened https://github.com/git-for-windows/git-sdk-arm64/pull/21 (without testing because I'd have to spin up a VM for that and that takes more time than I have right now), could you give it a whirl? Feel free to force-push as needed.

dscho:
@dennisameling now that [I verified](https://github.com/git-for-windows/git-sdk-arm64/pull/21#issuecomment-2286032542) that the `makepkg-git` fixes I suggested do work as intended (and keeping in mind that the `build-installers` SDK artifact is a superset of the `makepkg-git` SDK artifact), I believe that this here PR can be closed in favor of https://github.com/git-for-windows/git-sdk-arm64/pull/21 (and https://github.com/git-for-windows/build-extra/pull/574).

dennisameling:
Closing in favor of https://github.com/git-for-windows/git-sdk-arm64/pull/21. Thanks so much @dscho!

PR #574, opened 2024-08-13
makepkg-git: support Windows/ARM64 better

dscho:
As of e292201a (please.sh: change from bitness to architecture in create_sdk_artifact, 2022-12-28), we want to use the CPU architecture (or equivalent) instead of the "bitness", but there was one forgotten instance that still needs to be fixed.


dscho:
@dennisameling when I tried building the `mingw-w64-git` package on Windows/ARM64 with the `makepkg-git` SDK artifact, I noticed that the `/usr/bin/git` wrapper pointed to a totally bogus location. This is my attempt to fix that.

dennisameling:
Good catch, thank you!!

dennisameling (approved):

PR #575, opened 2024-08-13
7z.sfx: update to v24.08

dscho:
This corresponds to the commit 4e0cd6fa (Add a Dependabot configuration, 2024-06-28) of
https://github.com/git-for-windows/7-Zip/commits/v24.08-VS2022-sfx and the artifacts have been copied from the workflow run at https://github.com/git-for-windows/7-Zip/actions/runs/10374851525.

This addresses git-for-windows/git#5104


dscho:
I just got notified of https://www.cve.org/CVERecord?id=CVE-2024-11477, and am glad that we got this here PR merged in time for Git for Windows v2.46.1. Having said that, my assessment is that Git for Windows would not have been affected anyway:

- We do not ship 7-Zip, only PortableGit uses the SFX module.
- The vulnerability affects only the Zstandard decompression (which the SFX module does not even include)
- It requires a crafted archive (but we craft those archives ourselves, thanks very much 😉)
- The bug was fixed in version 24.07 (we upgraded to 24.08 via this here PR, in time for Git for Windows v2.46.1).

PR #578, opened 2024-10-27
installer: skip Experimental Options page, if empty

dscho:
As of a75602aa0 (installer: hide the "experimental" FSMonitor setting by default, 2024-07-17), the FSMonitor option is hidden from the options page unless the user had enabled it previously.

However, this then left the page blank (or: all options hidden), not a good user experience. So let's skip it if it would appear blank.

This fixes https://github.com/git-for-windows/git/issues/5231.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1884,6 +1884,18 @@ end;
 
 procedure QueryUninstallValues; forward;
 
+function IsHiddenExperimentalOptionsPageEmpty:Boolean;
+var
+    i:Integer;
+begin
+    Result:=True;
+#ifdef HAVE_EXPERIMENTAL_OPTIONS
+    for i:=1 to GP_MaxExperimental do
+        if (RdbExperimentalOptions[i]<>nil) and RdbExperimentalOptions[i].Visible then
+            Result:=False;
+#endif
+end;
+
 procedure InitializeWizard;
 var
     PrevPageID,TabOrder,TopOfLabels,Top,Left:Integer;
@@ -2589,7 +2601,10 @@ begin
             Result:=False
         else
             Result:=(PageID<>wpInfoBefore) and (PageID<>wpFinished);
-    end else
+    end else if (PageID=ExperimentalOptionsPage.ID) and IsHiddenExperimentalOptionsPageEmpty then
+        // Skip experimental options page if all options are hidden
+        Result:=True
+    else
         Result:=False;
 #ifdef DEBUG_WIZARD_PAGE
     Result:=PageID<>DebugWizardPage
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -336,17 +336,17 @@ end;
 
 const
     // Git Editor options.
-    GE_Nano           = 0;
-    GE_VIM            = 1;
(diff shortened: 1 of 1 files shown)


installer/install.iss
```diff
@@ -423,6 +423,7 @@ const
     GP_BuiltinAddI     = 4;
     GP_EnablePCon      = 5;
     GP_EnableFSMonitor = 6;
+    GP_Max             = 6;
```
rimrul, line 426: This is a great idea, but the name is slightly confusing since we use the same `GP_` prefix for other things.

rimrul (approved):
The code looks great. We should at some point rework the `GP_` prefix recycling though to avoid confusion.

installer/install.iss
dscho: I revamped this rather completely. Could you have another look, @rimrul?

mjcheetham (approved):
Good fix and nice QOL improvements too :-)

dscho:
/add relnote bug Due to a bug introduced in the v2.47 cycle, [the installer showed an empty "experimental options" page](https://github.com/git-for-windows/git/issues/5231), which was [fixed](https://github.com/git-for-windows/build-extra/pull/578).

jerryfleurival:
@dscho 
> As of [a75602a](https://github.com/git-for-windows/build-extra/commit/a75602aa089e9a119764187484c8fcb775750231) (installer: hide the "experimental" FSMonitor setting by default, 2024-07-17), the FSMonitor option is hidden from the options page unless the user had enabled it previously.
> 
> However, this then left the page blank (or: all options hidden), not a good user experience. So let's skip it if it would appear blank.
> 
> This fixes [git-for-windows/git#5231](https://github.com/git-for-windows/git/issues/5231).

Yes, this indeed need to improve user experience of the last page. As for all, This does not effect the installation process or nothing wrong with the installation as you can go ahead and just hit the "Install" button and installation will begin. Though, at some point in the future, this (user experience) needs to be address. 

As for enabling experimental options, I would assume the user set `core.fsmonitor` to `true`  either  by system, global or repository(local) levels  in the command line themselves. Experimental option such as `fsmonitor.daemon`.

dscho:
> > This fixes [git-for-windows/git#5231](https://github.com/git-for-windows/git/issues/5231).
> 
> Yes, this indeed need to improve user experience of the last page.

@jerryfleurival I don't understand... Do you mean to say that this is _not_ fixed for you? For me, the last page is:

image

After that, I only get the following page if I have a Git Bash open, otherwise it immediately starts to install Git for Windows:

image

At no point do I see an empty "Experimental Options" page. Is this different for you?

jerryfleurival:
@dscho I'm on version 2.47.0.2. I left those two unchecked, and then came a blank Experimental Options page. 
Ref: https://github.com/git-for-windows/git/issues/5273#issue-2687957789

dscho:
> I'm on version 2.47.0.2.

Well, the [fix](https://github.com/git-for-windows/git/issues/5231) is [in v2.47.1](https://github.com/git-for-windows/git/milestone/95?closed=1), that's why you don't have it. You need to upgrade.

PR #586, opened 2025-01-04
installer: really default to Secure Channel

rimrul:
Commit 8a7ab99 (installer: switch the libcurl default to Secure Channel, 2023-05-15) attempted to change the default libcurl backend for clean new installs to Secure channel, but only adjusted one of two fallbacks in the installer. There is one default value that gets passed to ReplayChoice() to return if it finds no previous choice and one value that gets used if ReplayChoice() returns an unexpected value. 8a7ab99 changed the latter to use Secure Channel, but left the former to default to OpenSSL. Adjust the default value to match the intention of that patch.

This fixes https://github.com/git-for-windows/git/issues/5295


dscho (approved):
Excellent find, thank you for that and the fix!

PR #587, opened 2025-01-10
Update InnoSetup from 6.1.0 to 6.4.0

dennisameling:
These changes are the output of running `./update-inno-setup.sh` locally.

Release notes: https://jrsoftware.org/files/is6-whatsnew.htm

Tested a `git-artifacts` [pipeline ](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/12713549285/job/35442493562#step:7:37)with the updated installer version, and everything seems to be working fine.


rimrul:
#568 updated to 6.3.3, but apparently `/open pr` didn't pick up the `.isl` files.

dscho:
> #568 updated to 6.3.3, but apparently `/open pr` didn't pick up the `.isl` files.

Maybe we should fix that? I do not really understand what is broken, though, because https://github.com/git-for-windows/git-for-windows-automation/blob/c47fca11b13202133c63df453efcafdb9511e1fd/update-scripts/version/innosetup#L11 looks fine to me...

dscho:
Also: I would like to hold off upgrading from InnoSetup v6.3.3 to v6.4.0 for a bit, for these reasons:

- As I mentioned [here](https://github.com/git-for-windows/git/issues/5350#issuecomment-2578383528), the new version offers an opportunity to drop the finicky [ExecWithCapture() function](https://github.com/git-for-windows/build-extra/blob/84ed494f4a7d9667bf3d54ed87d1da7196ec47cb/installer/exec-with-capture.inc.iss#L134-L211). This will need some work, and especially thorough testing.
- I released Git for Windows v2.48.0-rc2 _yesterday_. It already was delayed a lot by many issues I faced, from a `gettext` update requiring a new library I had not included in the `minimal-sdk` artifact to build problems caused by _big_ changes integrated very, very late in Git's release cycle, to outright test failures, to new flaky behavior. And this is all without any meaningful testing because Git's release cycle did not care that it was the winter holidays and sane people did not touch their computers. I would really like to keep a chance alive that the upcoming Git for Windows isn't completely broken, and to bring that about I am now extremely reluctant to integrate even more new stuff in addition to the gifts imparted by the Git project.
- As a rule, I like to integrate bigger changes a week or two _after_ releases, to give those changes maximal exposure before being part of an official Git for Windows release.

PR #588, opened 2025-02-05
please.sh build-mingw-w64-git: assume that all dependencies are installed

dscho:
The idea of having `git-sdk-*` repositories that track full MSYS2 environments is not only to have a convenient and fast way to bootstrap a build environment in which to build and release Git for Windows, it also enables straight-forward processes for embargoed builds, where we have to "time-travel" back to the state from which the preceding version was built and assembled.

To guarantee that this time traveling continues to work, the build process of a regular Git for Windows release must be prevented from "conveniently" installing missing dependencies; If they are not part of the `git-sdk-*` repository (or not in the `build-installers` sparse checkout), that needs to be changed before releasing, so that the embargoed release process can continue to work as intended.

### DIFF
--- a/please.sh
+++ b/please.sh
@@ -3428,6 +3428,7 @@ build_mingw_w64_git () { # [--only-i686] [--only-x86_64] [--only-aarch64] [--ski
 	sed_makepkg_e=
 	force=
 	src_pkg=
+	maybe_sync=
 	while case "$1" in
 	--only-i686|--only-32-bit)
 		MINGW_ARCH=mingw32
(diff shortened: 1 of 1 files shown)


dscho:
I verified that this works [in `git-sdk-64`](https://github.com/git-for-windows/git-sdk-64/actions/runs/13157178763), and [in `git-sdk-32`](https://github.com/git-for-windows/git-sdk-32/actions/runs/13157774103), and [in `git-sdk-arm64`](https://github.com/git-for-windows/git-sdk-arm64/actions/runs/13159140512) (for some reason, in `git-sdk-arm64` the installer failed to overwrite the already-existing uninstaller, but the rest of the build worked).

mjcheetham (approved):
I assume the new `--maybe-sync` option will be added in a related PR where necessary?

dscho:
> I assume the new `--maybe-sync` option will be added in a related PR where necessary?

It won't actually be necessary, I expect. I just added this option in case somebody relies on the `-s --noconfirm` stanza in any existing scenario (which I doubt).

PR #590, opened 2025-02-05
Release Notes: announce that `git svn` will be phased out

dscho:
It poses a substantial maintenance burden, what with the need not only to keep maintaining Perl, but also subversion (and apr, apr-util, and swig, and the Perl bindings of subversion).

At the same time, less and less `git svn` tickets (and no PRs worth talking about) have trickled in.

Therefore it is a good time to say good-bye to `git svn` and send it off to a cozy, relaxing retirement on the beach. It has served us well and can now go on to enjoy life instead of work.

I'd like this to be part of v2.48.1's release announcement.


mjcheetham (approved):

dscho:
For the record: This PR is part of https://github.com/git-for-windows/git/issues/5405.

PR #593, opened 2025-02-15
installer: use modern innosetup features

### LINKED ISSUE #5421: installer: use new InnoSetup functionality
In [InnoSetup v6.4.0](https://jrsoftware.github.io/issrc/whatsnew.htm):

> * Added new `ExecAndCaptureOutput` support function to execute a program or batch file and capture its _stdout_ and _stderr_ outputs separately.

This is of particular interest to me, as it will hopefully allow us to drop the finicky [`ExecWithCapture()` function](https://github.com/git-for-windows/build-extra/blob/84ed494f4a7d9667bf3d54ed87d1da7196ec47cb/installer/exec-with-capture.inc.iss#L134-L211). 

 In [InnoSetup v6.3](https://jrsoftware.org/files/is6.3-whatsnew.htm):

> ## Support for Arm64 systems improved, and related enhancements
>
> -   Setup now officially supports the installation of x64 apps on Arm64 Windows 11 systems, which are able to run x64 binaries via emulation. To enable your x64 app installers to run properly on Arm64 Windows 11, some minor changes may be needed in your scripts. Most importantly:
>     1.  In `ArchitecturesAllowed` and `ArchitecturesInstallIn64BitMode`, change any use of `x64` to `x64compatible`.
>     2.  In `Check` parameters and `[Code]`, change any use of `IsX64` to `IsX64Compatible`.
>     3.  In `[Code]`, if there are any `ProcessorArchitecture = paX64` comparisons, replace them with calls to `IsX64Compatible`.
>
>     The key difference between `x64`/`IsX64` and the new `x64compatible`/`IsX64Compatible` is that the latter matches both x64 Windows and Arm64 Windows 11.
>
>     In most cases, **you should make the above changes**, because otherwise, users on Arm64 systems may not be able to run your installers. For example, an `ArchitecturesAllowed=x64` setting will only allow the installer to run on x64 Windows --- not on Arm64 Windows 11. Or, if you ship x86 and x64 versions of your app in the same installer, the 32-bit x86 version may be chosen instead of the expected x64 version when running on Arm64 Windows 11.
> -   The `[Setup]` section directives `ArchitecturesAllowed` and `ArchitecturesInstallIn64BitMode` have been enhanced:
>     -   Six new architecture identifiers have been introduced. Briefly:
>         -   `arm32compatible` matches systems capable of running 32-bit Arm binaries.
>         -   `x64compatible` matches systems capable of running x64 binaries.
>         -   `x64os` matches systems running x64 Windows only. (Equivalent to the existing `x64` identifier, which is now deprecated.)
>         -   `x86compatible` matches systems capable of running 32-bit x86 binaries.
>         -   `x86os` matches systems running 32-bit x86 Windows only. (Equivalent to the existing `x86` identifier.)
>         -   `win64` matches systems running 64-bit Windows, regardless of architecture.See the new *Architecture Identifiers* [help topic](https://jrsoftware.org/ishelp/index.php?topic=archidentifiers) for further details on each.
>     -   Boolean expressions are now supported. Through use of the `and` operator, for example, it is possible to require two architecture identifiers to match at the same time. See the `ArchitecturesAllowed` [help topic](https://jrsoftware.org/ishelp/index.php?topic=setup_architecturesallowed) for usage examples.
> -   The `x64` architecture identifier is now deprecated. If it is used, the compiler will issue a warning and substitute `x64os`, which has the same effect. But as mentioned above, in most cases, **scripts should be changed** to use `x64compatible` because it matches both x64 Windows and Arm64 Windows 11.
> -   The 64-bit example scripts (`64Bit*.iss`) have all been updated to use `x64compatible` as preferred.
> -   Certain 64-bit functionality that previously only worked on x64 Windows now works on Arm64 Windows 11 as well. This includes:
>     -   The `Permissions` parameter of the `[Dirs]` section when running in 64-bit install mode.
>     -   The `Permissions` parameter of the `[Files]` section when running in 64-bit install mode or when the `64bit` flag is used.
>     -   The `Permissions` parameter of the `[Registry]` section when running in 64-bit install mode or when the value of the `Root` parameter ends in `64`.
>     -   The `regtypelib` flag of the `[Files]` section when running in 64-bit install mode or when the `64bit` flag is used.Note that all of the above remains unsupported on Arm64 Windows 10.
> -   Setup now logs the machine types supported by the system --- that is, what types of EXEs can be executed, either natively or via emulation. For example, when running on an Arm64 Windows 11 system, it logs: `Machine types supported by system: x86 x64 Arm32 Arm64`.
> -   The `OnlyOnTheseArchitectures` message is not used anymore. Instead, the `WindowsVersionNotSupported` message is now shown when Setup is started on an architecture that is not allowed by the `ArchitecturesAllowed` expression. (But please do not remove the message from translation files.)
> -   Pascal Scripting change: Added new `IsArm32Compatible`, `IsX64Compatible`, `IsX64OS`, `IsX86Compatible`, and `IsX86OS` support functions. The `IsX64` support function still exists but is now deprecated as explained above. Example testing all architecture identifiers:
>
>     [Code]
>     function InitializeSetup: Boolean;
>     begin
>       if IsArm32Compatible then Log('IsArm32Compatible');
>       if IsArm64 then Log('IsArm64');
>       if IsX64OS then Log('IsX64OS');
>       if IsX64Compatible then Log('IsX64Compatible');
>       if IsX86 then Log('IsX86');
>       if IsX86OS then Log('IsX86OS');
>       if IsX86Compatible then Log('IsX86Compatible');
>       if IsWin64 then Log('IsWin64');
>       Result := True;
>     end;

Would be nice to add better Windows/ARM64 support (even though it seems that Delphi, i.e. the development environment underlying InnoSetup, does not have Windows/ARM64 target support yet).

rimrul:
Innosetup 6.4 introduced a New function ExecAndCaptureOutput that executes commands and captures there standard output and standard error streams. We've been using our own howmgrown solution called ExecWithCapture for this purpose, but it's a bit finicky. Switch to the official implementation.


installer/install.iss
```diff
@@ -2584,9 +2588,9 @@ begin
             if DirExists(AppDir) then begin
                 if not FileExists(ExpandConstant('{tmp}\blocked-file-util.exe')) then
                     ExtractTemporaryFile('blocked-file-util.exe');
-                Cmd:='"'+ExpandConstant('{tmp}\blocked-file-util.exe')+'" blocking-pids "'+AppDir+'"';
-                if not ExecWithCapture(Cmd,StdOut,StdErr,Res) or (Res<>0) then begin
-                    Msg:='Skipping installation because '+AppDir+' is still in use:'+#13+#10+StdErr;
+                Cmd:='"'+ExpandConstant('{tmp}\blocked-file-util.exe')+'"';
+                if not ExecAndCaptureOutput(Cmd, 'blocking-pids "'+AppDir+'"', '', SW_SHOWNORMAL, ewWaitUntilTerminated, Res, Output) or (Res<>0) then begin
```
rimrul, line 2592: It might make sense to use AppDir as the working dir here

installer/install.iss
```diff
@@ -604,10 +603,12 @@ var
 // Returns true if at least one FSMonitor daemon was shut down successfully
 function ShutdownFSMonitorDaemons():Boolean;
 var
+    StdOut:String;
     FindRec:TFindRec;
-    ExitCode:DWORD;
-    Path,Str:String;
+    ExitCode:Integer;
+    Path:String;
```
rimrul, line 609: I removed `Str`, realised that `TExecOutput.StdOut` and `TExecOutput.StdErr` aren't `String`, but `TArrayOfString` and had to reintroduce a String here. I should probably at least re-join the lines here.

installer/install.iss
```diff
@@ -837,12 +838,13 @@ begin
...
 
-    if not ExecWithCapture('"'+AppDir+'\{#MINGW_BITNESS}\bin\git.exe" config -l -z '+ExtraOptions,StdOut,StdErr,ExitCode) then begin
+    if not ExecAndCaptureOutput('"'+AppDir+'\{#MINGW_BITNESS}\bin\git.exe"', 'config -l -z '+ExtraOptions, '', SW_SHOWNORMAL, ewWaitUntilTerminated, ExitCode, Output) then begin
         if FileExists(AppDir+'\{#MINGW_BITNESS}\bin\git.exe') then
-            LogError('Unable to get system config (exit code '+IntToStr(ExitCode)+'):'+#13+#10+StdErr);
+            LogError('Unable to get system config (exit code '+IntToStr(ExitCode)+'):'+#13+#10+StringJoin(#13+#10,Output.StdErr));
     end;
 
     // Split NUL-delimited key/value pairs, extract LF that denotes end of key
+    StdOut:=StringJoin(#10,Output.StdOut);
```
rimrul, line 847: We probably want to iterate over `Output.StdOut` directly instead of joining and re-slitting this, but that was the easiest Q&D way to at least get it to compile again.

rimrul (commented):
This is entirely untested, I just made sure it builds.

installer/install.iss
rimrul: Looking at https://github.com/git-for-windows/build-extra/blob/30a1958fea52277914b4d4439f1442d332ecd9f1/git-extra/blocked-file-util.c#L79-L110 ,
`blocked-file-util.exe` doesn't care about working dir.

installer/install.iss
rimrul: We do that now, but there's probably at least one off-by-one error in there.

installer/install.iss
```diff
@@ -54,7 +54,7 @@ SolidCompression=yes
 #endif
 SourceDir={#SOURCE_DIR}
 #if BITNESS=='64' || INSTALLER_FILENAME_SUFFIX=='arm64'
-ArchitecturesInstallIn64BitMode=x64 arm64
+ArchitecturesInstallIn64BitMode=x64compatible arm64
```
dennisameling, line 57: With the new `x64compatible`, there should be [no need](https://jrsoftware.org/ishelp/index.php?topic=setup_architecturesinstallin64bitmode) to explicitly set `arm64` anymore:

> **x64compatible**
> Matches systems capable of running x64 binaries. This includes systems running x64 Windows, and also Arm64-based Windows 11 systems, which have the ability to run x64 binaries via emulation.

dscho: Makes sense to me!

dennisameling:
@rimrul with all the recent updates to the installer, there's some conflicts on this PR now. Do you have bandwidth to rebase it onto the latest main branch? Happy to help if not.

rimrul:
Yeah, I should get around to it in the next few days.

installer/install.iss
```diff
@@ -615,22 +615,22 @@ begin
             // Find out which form to use.
             if (BuiltinFSMonitorStopOption='') then begin
                 BuiltinFSMonitorStopOption:='(huh?)';
-                if not ExecWithCapture('"'+AppDir+'\cmd\git.exe" fsmonitor--daemon -h',Str,Str,ExitCode) or (ExitCode<>129) then begin
+                if not ExecAndCaptureOutput('"'+AppDir+'\cmd\git.exe"', 'fsmonitor--daemon -h', '', SW_SHOWNORMAL, ewWaitUntilTerminated, ExitCode, Output) or (ExitCode<>129) then begin
```
dscho, line 618: I see that the deleted code assigned `StartupInfo.wShowWindow:=SW_HIDE;`, so I was a bit concerned about the `SW_SHOWNORMAL` here. But [the documentation](https://jrsoftware.org/ishelp/index.php?topic=isxfunc_execandcaptureoutput) of `ExecAndCaptureOutput()` says:

> Console programs are always hidden and the ShowCmd parameter only affects GUI programs, so always using SW\_SHOWNORMAL instead of SW\_HIDE is recommended.

So all is good. If at all, I would document that in the commit message. But I could live with it in the present shape, too.

dscho (approved):

dscho (approved):
Thank you!

PR #597, opened 2025-02-19
pacman-helper.sh: add quick_remove mode

jeremyd2019:
add an action parameter to the prior quick_add function, and use if checks to see if we are adding or removing.  Provide wrapper functions for quick_add and quick_remove.

For https://github.com/git-for-windows/git-sdk-32/issues/39#issuecomment-2664159567

### DIFF
--- a/pacman-helper.sh
+++ b/pacman-helper.sh
@@ -431,9 +431,18 @@ quick_action () { # <action> <file>...
 		 printf '%s\n' $msys $mingw |
 		 sed '/\.pkg\.tar/{
 			s/-[^-]*-[^-]*-[^-]*\.pkg\.tar\.\(xz\|zst\)$/-[0-9]*/
-			b
+			b1
 		 }
-		 s/$/-[0-9]*/' |
+		 s/$/-[0-9]*/
+		 :1
+		 p
+		 # Prevent false positives (e.g. deleting `msys2-runtime-3.3` when
+		 # updating `msys2-runtime`) by requiring the suffix to be of the form
+		 # `-<pkgver>-<pkgrel>-<arch><pkgext>`. Sadly, there are no non-greedy
+		 # wildcards, therefore do this via an "exclude pattern" instead:
+		 # `:(exclude)<pkgname>-[0-9]*-*-*-*`
+		 s/$/-*-*-*/
+		 s/^/:(exclude)/' |
 		 xargs git rm --sparse --cached -- ||
 		 die "Could not remove the existing versions from the Git branch in $arch"
 
--- a/pacman-helper.sh
+++ b/pacman-helper.sh
@@ -30,13 +30,14 @@ die () {
 
 mode=
 case "$1" in
-lock|unlock|break_lock|quick_add)
+lock|unlock|break_lock|quick_add|quick_remove)
 	mode="$1"
 	shift
 	;;
 *)
-	die "Usage:\n%s\n%s\n" \
+	die "Usage:\n%s\n%s\n%s\n" \
 		" $0 quick_add <package>..." \
+		" $0 quick_remove <package>..." \
 		" $0 ( lock | unlock <id> | break_lock )"
 	;;
 esac
@@ -107,6 +108,23 @@ repo_add () {
 	"$this_script_dir/repo-add" "$@"
 }
 
+repo_remove () {
+	if test ! -s "$this_script_dir/repo-remove"
+	then
+		# Make sure that GPGKEY is used unquoted
+		 sed 's/"\(\${\?GPGKEY}\?\)"/\1/g' </usr/bin/repo-remove >"$this_script_dir/repo-remove"
+	fi &&
+	"$this_script_dir/repo-remove" $(for arg
+	do
+		# repo-remove only accepts package _names_, but we are potentially given _files_.
+		# Handle this by distilling the package names from filenames.
+		case "$arg" in
+		*.pkg.tar.xz|*.pkg.tar.zst) echo "${arg%-*-*-*}";;
+		*) echo "$arg";;
+		esac
+	done)
+}
+
 sanitize_db () { # <file>...
 	perl -e '
 		foreach my $path (@ARGV) {
@@ -174,9 +192,16 @@ sanitize_db () { # <file>...
 	fi
 }
 
-quick_add () { # <file>...
-	test $# -gt 0 ||
+quick_action () { # <action> <file>...
+	test $# -gt 1 ||
 	die "Need at least one file"
+
+	label="$1"
+	shift
+	case "$label" in
+	add|remove) action=repo_$label;;
+	*) die "Unknown action '$action'";;
+	esac
 
 	if test -z "$PACMANDRYRUN$azure_blobs_token"
 	then
@@ -212,7 +237,8 @@ quick_add () { # <file>...
 	x86_64_msys=
 	all_files=
 
-	# Copy the file(s) to the temporary directory, and schedule their addition to the appropriate index(es)
+	# Copy the file(s) to the temporary directory, and schedule their addition to the appropriate index(es),
+	# or for `remove`: schedule their removal from the appropriate index(es).
 	for path in "$@"
 	do
 		file="${path##*/}"
@@ -259,12 +285,12 @@ quick_add () { # <file>...
 		*) echo "Skipping file with unknown arch: $file" >&2; continue;;
 		esac
 
-		echo "Copying $file to $arch/..." >&2
 		test -z "$key" || eval "$key=\$$key\\ $file"
 		all_files="$all_files $arch/$file"
 
 		if test ! -d "$dir/$arch"
 		then
+			echo "Initializing $dir/$arch..." >&2
 			git -C "$dir" rev-parse --quiet --verify refs/remotes/origin/$arch >/dev/null ||
 			git -C "$dir" fetch --depth=1 origin x86_64 aarch64 i686 ||
 			die "$dir: could not fetch from pacman-repo"
@@ -273,6 +299,15 @@ quick_add () { # <file>...
 			die "Could not initialize $dir/$arch"
 		fi
 
+		case "$label" in
+		remove)
+			test -z "$GPGKEY" ||
+			all_files="$all_files $arch/$file.sig"
+			continue
+			;;
+		esac
+
+		echo "Copying $file to $arch/..." >&2
 		cp "$path" "$dir/$arch" ||
 		die "Could not copy $path to $dir/$arch"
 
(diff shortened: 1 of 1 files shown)


pacman-helper.sh
```diff
@@ -212,88 +224,134 @@ quick_add () { # <file>...
...
+				;;
+			git-extra)
+				# TODO we have no way to infer the arch from an msys pkgname
+				arch=i686
+				key=${arch}_mingw
+				;;
+			*)
+				# TODO we have no way to infer the arch from an msys pkgname
+				arch=i686
+				key=${arch}_msys
```
jeremyd2019, line 325: I don't know what to do about the `arch` here... do I need to add another argument to specify which arch msys packages should be removed from?

jeremyd2019: yeah, I just started looking at build-and-deploy.yml, and I think adding an architecture argument would be the way to go.

dscho:
Hey @jeremyd2019 good stuff!

I worked on this a bit more (and this time I was able to push the PR branch! Yay!). In particular, I wanted to make this a bit more flexible, so that would could unshadow, say, `p7zip` from both the i686 and the x86_64 repository in one go.

The main idea is to skip the `<architecture>` argument and use either full package filenames or partial package names _that include the architecture_. For example, removing `p7zip` could now be accomplished like this:

```console
$ pacman-helper.sh quick_remove p7zip-i686 p7zip-x86_64
```

To keep things simpler, I want to avoid near-duplicating the `for` loop, and I also wanted to avoid the `if`/`else`/`fi` pattern in favor of storing `quick_add` or `quick_remove` in the `action` variable and then calling the respective function via `$action <argument>...`.

What do you think of that new version? If you're fine with it, I'll auto-squash the `fixup!`/`amend!` and merge.

jeremyd2019:
OK.

I don't necessarily want to hold up this PR, because it isn't really related to the changes here, but some of these patterns would seem to cause issues for packages with numbers in their names.  Specifically, I think the next update of msys2-runtime will `git rm` msys2-runtime-3.3 files as well.  Something to think about...

jeremyd2019:
repo-add does have `-R|--remove` option, to remove the old file when adding a new version of a package.  I don't know if that would work for you, and doesn't help the repo-remove case.

If you want to get deep into it, you can see repo-add does this to figure out what to remove:
```bash
find_pkgentry() {
        local pkgname=$1
        local pkgentry

        for pkgentry in "$tmpdir/db/$pkgname"*; do
                name=${pkgentry##*/}
                if [[ ${name%-*-*} = "$pkgname" ]]; then
                        echo $pkgentry
                        return 0
                fi
        done
        return 1
}

                        if (( RMEXISTING )); then
                                local oldfilename="$(sed -n '/^%FILENAME%$/ {n;p;q;}' "$pkgentry/desc")"
                                local oldfile="$(dirname "$1")/$oldfilename"
                        fi
```

pacman-helper.sh
```diff
@@ -273,14 +317,23 @@ quick_add () { # <file>...
 			die "Could not initialize $dir/$arch"
 		fi
 
+		case "$label" in
+		remove)
+			test -z "$GPGKEY" ||
```
jeremyd2019, line 322: I think this would be better
```suggestion
			test -f "$arch/$file.sig" &&
```

but it doesn't really matter, the only time there's not a GPGKEY should be when it's a dry run, so it'll only change output of what actions it would have done.

pacman-helper.sh
```diff
@@ -248,6 +274,24 @@ quick_add () { # <file>...
 			# skip explicit signatures; we copy them automatically
 			continue
 			;;
+		*-i686|*-x86_64|*-aarch64)
+			test remove = "$label" || die "Cannot add $path"
+			arch=${file##*-}
+			file=${file%-$arch}
+			file=${file%-[0-9]*}
+			file=${file%-[0-9]*}
```
jeremyd2019, line 282: Would it be better/safer to do
```suggestion
			file=${file%-[0-9]*-[0-9]*}
```
and require that if you give a version, you give it in full (including pkgrel)?

If you can enable `extglob` you could maybe use a pattern like `-*([^-])-*([^-])`?  And match filenames with `-*([^-])-*([^-])-*([^-]).pkg.tar.*([^.])` instead of `-[0-9]*`?  (some of those glob patterns seem to be interpreted by git instead of by bash before calling git, so I don't know if extglob would work in that case)

pacman-helper.sh
dscho: That would not actually work because we're in a sparse checkout; Only the package database is present at this stage.

pacman-helper.sh
dscho: That is a valid concern. For example, `quick_remove msys2-runtime-3.3-x86_64` would clearly like to remove the `msys2-runtime-3.3` package, not the `msys2-runtime` package. I'll apply your change and squash it in.

dscho:
> [...] some of these patterns would seem to cause issues for packages with numbers in their names. Specifically, I think the next update of msys2-runtime will `git rm` msys2-runtime-3.3 files as well. Something to think about...

@jeremyd2019 Very valid point! I added https://github.com/git-for-windows/build-extra/pull/597/commits/0ef4d2a844e35f991aa278ad10145c0f0a9e152c to address that.

I verified this by doing a dry-run of `quick_add`ing a (stale, but doesn't matter as this was only for testing) `msys2-runtime` archive:

```console
$ PACMANDRYRUN=1 ./pacman-helper.sh quick_add /var/cache/pacman/pkg/msys2-runtime-3.5.5-3-x86_64.pkg.tar.xz
[...]
Leaving temporary directory /tmp/tmp.6yspBjDZPQ/ for inspection

$ cd /tmp/tmp.6yspBjDZPQ/x86_64

$ git show --raw
commit 09472db379e348f218dc43894b9ce553c617becc (HEAD -> x86_64)
Author: Johannes Schindelin <johannes.schindelin@gmx.de>
Date:   Thu Feb 20 11:25:20 2025 +0100

    Update 1 package(s)
    
    msys2-runtime -> 3.5.5-3
    
:100644 100644 e807a8f f786a67 M        git-for-windows-x86_64.db
:100644 100644 e807a8f f786a67 M        git-for-windows-x86_64.db.tar.xz
:100644 100644 2f4fa87 a268e91 M        git-for-windows-x86_64.files
:100644 100644 2f4fa87 a268e91 M        git-for-windows-x86_64.files.tar.xz
:000000 100644 0000000 a5482f3 A        msys2-runtime-3.5.5-3-x86_64.pkg.tar.xz
:000000 100644 0000000 55f4c8d A        msys2-runtime-3.5.5-3-x86_64.pkg.tar.xz.sig
:100644 000000 49574cf 0000000 D        msys2-runtime-3.5.7-1-x86_64.pkg.tar.xz
:100644 000000 18416eb 0000000 D        msys2-runtime-3.5.7-1-x86_64.pkg.tar.xz.sig

$ diff -u \
  <(git show HEAD^:git-for-windows-x86_64.db | tar tJvf - | sort -k 6) \
  <(git show HEAD:git-for-windows-x86_64.db | tar tJvf - | sort -k 6)
```
```diff
--- /dev/fd/63  2025-02-20 11:29:12.000000000 +0100
+++ /dev/fd/62  2025-02-20 11:29:12.000000000 +0100
@@ -128,8 +128,8 @@
 -rw-r--r-- root/root      1525 2023-05-09 23:00 msys2-runtime-3.3-3.3.6-1/desc
 drwxr-xr-x root/root         0 2023-05-09 23:00 msys2-runtime-3.3-devel-3.3.6-1/
 -rw-r--r-- root/root      1576 2023-05-09 23:00 msys2-runtime-3.3-devel-3.3.6-1/desc
-drwxr-xr-x root/root         0 2025-01-30 12:28 msys2-runtime-3.5.7-1/
--rw-r--r-- root/root       711 2025-01-30 12:28 msys2-runtime-3.5.7-1/desc
+drwxr-xr-x root/root         0 2025-02-20 11:25 msys2-runtime-3.5.5-3/
+-rw-r--r-- root/root      1521 2025-02-20 11:25 msys2-runtime-3.5.5-3/desc
 drwxr-xr-x root/root         0 2025-01-30 12:28 msys2-runtime-devel-3.5.7-1/
 -rw-r--r-- root/root       760 2025-01-30 12:28 msys2-runtime-devel-3.5.7-1/desc
 drwxr-xr-x root/root         0 2025-02-18 13:13 openssh-9.9p2-1/
```

I also verified that the `quick_remove` operation does what it is supposed to do:

```console
$ PACMANDRYRUN=1 ./pacman-helper.sh quick_remove cyrus-sasl-2.1.28-2-i686
[...]
Leaving temporary directory /tmp/tmp.BxRTsyW8g0/ for inspection

$ cd /tmp/tmp.BxRTsyW8g0/i686/

$ git show --raw
commit acb31edf6d6e097ed0435551cf9f5c8bbb5d02fb (HEAD -> i686)
Author: Johannes Schindelin <johannes.schindelin@gmx.de>
Date:   Thu Feb 20 11:29:48 2025 +0100

    Remove 1 package(s)

    cyrus-sasl

:100644 000000 ce16dbd 0000000 D        cyrus-sasl-2.1.28-2-i686.pkg.tar.xz
:100644 000000 b797413 0000000 D        cyrus-sasl-2.1.28-2-i686.pkg.tar.xz.sig
:100644 100644 90ce06d 48b9dbe M        git-for-windows-i686.db
:100644 100644 90ce06d 48b9dbe M        git-for-windows-i686.db.tar.xz
:100644 100644 89a39de 65eb381 M        git-for-windows-i686.files
:100644 100644 89a39de 65eb381 M        git-for-windows-i686.files.tar.xz

$ diff -u \
  <(git show HEAD^:git-for-windows-i686.db | tar tJvf - | sort -k 6) \
  <(git show HEAD:git-for-windows-i686.db
 | tar tJvf - | sort -k 6)
```
```diff
--- /dev/fd/63  2025-02-20 11:30:44.000000000 +0100
+++ /dev/fd/62  2025-02-20 11:30:44.000000000 +0100
@@ -6,8 +6,6 @@
 -rw-r--r-- root/root      1381 2023-01-26 23:03 ca-certificates-20211016-2/desc
 drwxr-xr-x root/root         0 2025-02-17 21:15 curl-8.12.1-2/
 -rw-r--r-- root/root       604 2025-02-17 21:15 curl-8.12.1-2/desc
-drwxr-xr-x root/root         0 2023-06-23 14:16 cyrus-sasl-2.1.28-2/
--rw-r--r-- root/root      1368 2023-06-23 14:16 cyrus-sasl-2.1.28-2/desc
 drwxr-xr-x root/root         0 2023-02-21 11:17 docbook-xsl-ns-1.79.2-2/
 -rw-r--r-- root/root      1308 2023-02-21 11:17 docbook-xsl-ns-1.79.2-2/desc
 drwxr-xr-x root/root         0 2024-06-14 11:31 fido2-tools-1.15.0-1/
```

So: all good to go no!

PR #603, opened 2025-03-01
pacman-helper.sh: compose release body/notes

jeremyd2019:
Do some trivial sed escaping of characters that might cause issues in JSON.  This is not comprehensive, but the text in question is controlled by this script and package names and versions being published, so we can reasonably be assured of a limited set of characters being used.  (In fact, it's probably unnecessary to escape quotes and backslashes).   

I noticed that the version of sed has a `-z` option, which nominally makes it operate on `\0`-delimited input instead of `\n`, but effectively removes the need for a loop to special case `\n`.

### DIFF
--- a/pacman-helper.sh
+++ b/pacman-helper.sh
@@ -349,6 +349,7 @@ quick_action () { # <action> <file>...
 	test -z "$GPGKEY" || sign_option=--sign
 	dbs=
 	to_push=
+	>release_notes.txt
 	for arch in $architectures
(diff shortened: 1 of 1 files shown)


dscho (approved):
`-z` is a great idea!

jeremyd2019:
it is non-standard (sed is a posix standard), but it's probably a safe bet that we're using GNU sed anywhere this script is likely to be run.

PR #611, opened 2025-04-20
Chinese Simplified for installer

TC999:

```
```

Thank you very much.


dscho:
Thank you for your contribution @TC999!

However, Git for Windows merely consumes these translations from InnoSetup. We do not add our own versions here. I see two _unofficial_ Chinese translations in that project: [simplified](https://github.com/jrsoftware/issrc/blob/main/Files/Languages/Unofficial/ChineseSimplified.isl) and [traditional](https://github.com/jrsoftware/issrc/blob/main/Files/Languages/Unofficial/ChineseTraditional.isl). It would most likely make more sense to contribute there (specifically, see the information at https://jrsoftware.org/files/istrans/, a link that you even quoted in the added file), and to try to get the translation to a level where it can be included in the official InnoSetup versions, from which Git for Windows will then automatically consume it.

In that light, I do not want to merge it here (which would be premature before getting the official blessing of the InnoSetup project).

PR #612, opened 2025-04-21
check-for-missing-dlls: support running on Windows/ARM64

dscho:


check-for-missing-dlls.sh
```diff
@@ -22,15 +22,25 @@ die "Could not enumerate system .dll files"
...
-x86_64) MINGW_PREFIX=mingw64;;
-aarch64) MINGW_PREFIX=clangarm64;;
-*) die "Unhandled architecture: $ARCH";;
-esac
+if test -z "$MINGW_PREFIX"
+then
+	case "$MSYSTEM" in
+	MINGW64) MINGW_PREFIX=mingw64;;
+	CLANGARM64) MINGW_PREFIX=clangarm64;;
+	MINGW32) MINGW_PREFIX=mingw32;;
```
dennisameling, line 30: I'm assuming this change is needed because `uname -m` will most likely report `x86_64` on an ARM64 machine given that Cygwin [doesn't natively support ARM64 yet](https://www.msys2.org/wiki/arm64/), correct?

dennisameling (approved):
This is very exciting, thank you!

PR #614, opened 2025-05-06
Update outdated text in install.iss

Tapped:
The proposal to change the term "master" to something else began six years ago. At this point, claiming the change will happen "in the near future" has clearly proven inaccurate.

There are likely other places in Git where similar outdated language still appears, however I think this is the only occurrence in this repository.


dscho (approved):
You are correct, and it makes me really sad that you are.

PR #615, opened 2025-05-09
Switch to the new Pacman repository

dscho:
I meant to do this a lot earlier: In February, I started migrating off of the Pacman repository that is hosted in my personal Azure account, to a new Pacman repository that is hosted on GitHub Releases.

With this PR, Git for Windows' SDK is moved over to the new Pacman repository.

Note that the `get-sources.sh` script is left alone, for now, as it will require substantial work (because it is no longer as easy to find the source package for a given package _version_; It will require [using the GitHub REST API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases), probably via `gh` so that the `--jq` option can be used to filter by the desired asset filename).

### DIFF
--- a/git-extra/git-extra.install
+++ b/git-extra/git-extra.install
@@ -99,30 +99,39 @@ GITATTRIBUTES
 	case "$uname_m" in
 	i686)
 		arch=i686
-		otherarch=x86-64
+		otherarch=x86_64
 		otherpacman=git-for-windows-mingw64
 		;;
 	x86_64)
-		arch=x86-64
+		arch=x86_64
 		otherarch=i686
 		otherpacman=git-for-windows-mingw32
 		;;
 	esac
 
+	! grep -q '^\[git-for-windows\]$' etc/pacman.conf ||
+	sed -i -e 's/^\(\[git-for-windows\)\(\]\)$/\1-'"$arch"'\2/' etc/pacman.conf
+
+	! grep -q wingit.blob.core.windows.net etc/pacman.conf ||
+	sed -i -e '/https:\/\/wingit\.blob\.core\.windows\.net\//{
+		s|x86-64|x86_64|
+		s|https://[^/]*|https://raw.githubusercontent.com/git-for-windows/pacman-repo/refs/heads|
+	}' etc/pacman.conf
+
 	grep -q git-for-windows[^-] etc/pacman.conf ||
-	sed -i -e '/^\[mingw32\]/i[git-for-windows]\nServer = https://wingit.blob.core.windows.net/'$arch'\n' etc/pacman.conf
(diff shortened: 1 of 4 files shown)


git-extra/git-extra.install
```diff
@@ -99,30 +99,36 @@ GITATTRIBUTES
...
 		;;
 	x86_64)
-		arch=x86-64
+		arch=x86_64
 		otherarch=i686
 		otherpacman=git-for-windows-mingw32
 		;;
 	esac
 
+	grep -q ! wingit.blob.core.windows.net ||
```
jeremyd2019, line 112: I'm not sure what this is trying to do

dscho: Yes, I am still in the process of developing this ;-)

dscho: @jeremyd2019 _now_ it's ready 😀

jeremyd2019:
> Note that the `get-sources.sh` script is left alone, for now, as it will require substantial work (because it is no longer as easy to find the source package for a given package _version_; It will require [using the GitHub REST API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases), probably via `gh` so that the `--jq` option can be used to filter by the desired asset filename).

That sounds like it'll suck - `--jq` is client-side filtering, so if you don't know what release you're looking for, you may have to try to list all releases & their assets from the server and use jq to find the one you want.  That doesn't scale well.

Also, lately I've been seeing HTTP 429 responses from github when attempting to get commits/PRs as `.patch` files - I hope they're not going to slap a rate limit on the raw files that'll cause issues with this.

dscho:
> > Note that the `get-sources.sh` script is left alone, for now, as it will require substantial work (because it is no longer as easy to find the source package for a given package _version_; It will require [using the GitHub REST API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases), probably via `gh` so that the `--jq` option can be used to filter by the desired asset filename).
> 
> That sounds like it'll suck - `--jq` is client-side filtering, so if you don't know what release you're looking for, you may have to try to list all releases & their assets from the server and use jq to find the one you want. That doesn't scale well.

Yes, this wouldn't scale well. I was considering caching the results in some way (and tracking that cache in Git).

If only GitHub Pages allowed 302s... But all I can find are `<meta>`-based redirects...

@jeremyd2019 do you know whether there is a way to let `pacman` handle `<meta>`-based redirects? If so, we could even use GitHub Pages to serve the Pacman repository...

jeremyd2019:
no, I did a quick search and I didn't see any way to get curl to do so, and pacman uses libcurl.

PR #616, opened 2025-05-13
git-extra(vimrc): adjust $PATH to remove Git's libexec

benknoble:
As described on the Git mailing list [1] and elsewhere [2], Git adjusts
PATH before invoking programs, and those adjustments are inherited by
editors if they are invoked as subprocesses. In most cases this is
harmless, but it can create confusion when spawning subprocesses from
the editor (as is common in Vim) and when the user's shell looks at PATH
or the location of a "git" binary.

Include portable code to strip these entries from PATH on startup.

[1]: https://public-inbox.org/git/CALnO6CDtGRRav8zK2GKi1oHTZWrHFTxZNmnOWu64-ab+oY3_Lw@mail.gmail.com/
[2]: https://benknoble.github.io/blog/2020/05/22/libexec-git-core-on-path/

Signed-off-by: D. Ben Knoble <ben.knoble+github@gmail.com>

---

This code comes from my own Dotfiles [3] via 2 authors, myself and
https://github.com/Konfekt, on their recommendation [4].

The only major user-facing change would be that `:!git-commit` and
similar probably no longer work, since libexec-style dir containing
those binaries wouldn't be on PATH anymore. But Git doesn't support that
usage officially, anyway?

I took the liberty of swapping `const` for `let` in this version to make
it a little more portable between versions of Vim; I'm not sure which
version Git for Windows ships, so if I need to convert the lambda
expressions into "v:val" strings, I'll do so.

[3]: https://github.com/benknoble/Dotfiles
[4]: https://github.com/benknoble/Dotfiles/issues/143#issuecomment-2869525481

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -5,7 +5,7 @@ pkgbase="mingw-w64-${_realname}"
 pkgname=($_realname
         "${MINGW_PACKAGE_PREFIX}-${_realname}")
 _ver_base=1.1
-pkgver=1.1.653.48e2403b3
+pkgver=1.1.654.aca5df34e
 pkgrel=1
 pkgdesc="Git for Windows extra files"
 arch=('any')
@@ -62,7 +62,7 @@ source=('inputrc'
         'git-askpass.h'
         'git-askpass.rc')
 sha256sums=('8ed76d1cb069ac8568f21c431f5e23caebea502d932ab4cdff71396f4f0d5b72'
-            'e36a3b93b8a33b0a74619f2449ed6d56ed54e4e2938b97070bce4926f1f44054'
+            '7736786309a58112b7ac60627b3df049eb5eac3e0627c364939277477aaa03f1'
             '640d04d2a2da709419188a986d5e5550ad30df23c7ea9be1a389c37a6b917994'
             '17c90698d4dd52dd9f383e72aee54ecea0f62520baf56722de5c83285507c0d9'
             '3cd83627f1d20e1108533419fcf33c657cbcf777c3dc39fa7f13748b7d63858a'
--- a/git-extra/vimrc
+++ b/git-extra/vimrc
@@ -52,4 +52,36 @@ if has("autocmd")
       autocmd Filetype diff
       \ highlight WhiteSpaceEOL ctermbg=red |
       \ match WhiteSpaceEOL /\(^+.*\)\@<=\s\+$/
+
+      " When Git starts up, it prepends its libexec dir to PATH to allow it to
+      " find external commands.
+      "
+      " Thus, if Vim is invoked via a Git process (such as the contrib git-jump,
+      " or any other usage of GIT_EDITOR/VISUAL/EDITOR in Git commands, be they
+      " scripts or internals--with the exception of manually invoking the script
+      " yourself, without using Git: sh .../git-jump), $PATH will contain
+      " something like libexec/git-core.
+      "
+      " We don't generally want it in Vim's $PATH, though, as it is passed down
+      " to *all* subprocesses, including shells started with :terminal or
+      " :shell.
+      function s:fix_git_path() abort
+          let slash = exists('+shellslash') && !&shellslash ? '\\' : '/'
+          let git_core_base = printf('%slib\%%(exec\)\?%sgit-core', slash, slash)
+          " optimization: early return
+          if $PATH !~# git_core_base
+              return
+          endif
+          let path_sep = has('win32') ? ';' : ':'
+          let new_path = split($PATH, path_sep)
+                      \ ->filter({_, d -> d !~# git_core_base..'$' })
+                      \ ->join(path_sep)
+          let $PATH = new_path
+      endfunction
+
+      augroup fix_git_path
+          autocmd!
+          autocmd VimEnter * call s:fix_git_path()
+      augroup end
+
 endif " has("autocmd")


benknoble:
<details><summary>range-diff: also update sha256sum</summary>

```
1:  2b4b5cf5 ! 1:  f8b44295 git-extra(vimrc): adjust $PATH to remove Git's libexec
    @@ Commit message
     
         Signed-off-by: D. Ben Knoble <ben.knoble+github@gmail.com>
     
    + ## git-extra/PKGBUILD ##
    +@@ git-extra/PKGBUILD: source=('inputrc'
    +         'git-askpass.h'
    +         'git-askpass.rc')
    + sha256sums=('8ed76d1cb069ac8568f21c431f5e23caebea502d932ab4cdff71396f4f0d5b72'
    +-            'e36a3b93b8a33b0a74619f2449ed6d56ed54e4e2938b97070bce4926f1f44054'
    ++            '7736786309a58112b7ac60627b3df049eb5eac3e0627c364939277477aaa03f1'
    +             '640d04d2a2da709419188a986d5e5550ad30df23c7ea9be1a389c37a6b917994'
    +             '17c90698d4dd52dd9f383e72aee54ecea0f62520baf56722de5c83285507c0d9'
    +             '3cd83627f1d20e1108533419fcf33c657cbcf777c3dc39fa7f13748b7d63858a'
    +
      ## git-extra/vimrc ##
     @@ git-extra/vimrc: if has("autocmd")
            autocmd Filetype diff
```

</details>

benknoble:
<details><summary>range-diff: and the pkgver? took this from CI; I'm probably missing the "workflow" here</summary>

```
1:  f8b44295 ! 1:  1bcec9ae git-extra(vimrc): adjust $PATH to remove Git's libexec
    @@ Commit message
         Signed-off-by: D. Ben Knoble <ben.knoble+github@gmail.com>
     
      ## git-extra/PKGBUILD ##
    +@@ git-extra/PKGBUILD: pkgbase="mingw-w64-${_realname}"
    + pkgname=($_realname
    +         "${MINGW_PACKAGE_PREFIX}-${_realname}")
    + _ver_base=1.1
    +-pkgver=1.1.653.48e2403b3
    ++pkgver=1.1.654.aca5df34e
    + pkgrel=1
    + pkgdesc="Git for Windows extra files"
    + arch=('any')
     @@ git-extra/PKGBUILD: source=('inputrc'
              'git-askpass.h'
              'git-askpass.rc')
```

</details>

Konfekt:
On a general side note, is an augroup for an VimEnter autocmd needed ? Though wrapping autocmd into augroups is considered best practice, this is meant to prevent double loading in the same Vim instance, but then VimEnter is only fired once anyway ?

dscho (commented):
Thank you for contributing!

I was a bit on the fence after I read the PR title (which didn't manage to convince me that this is a good change, it rather did the opposite: why _on Earth_ should it be _vim_'s job to clean up the `PATH`?), but can see that this might make sense.

Since I am unfamiliar with vim's arcane script language, the code in question is too hard to understand for me, and additional code comments would be much appreciated. For example, I wonder whether this code now adds a start-up penalty to _all_ invocations of `vim` and not just those where `GIT_EXEC_PATH` is set.

Also, I am still uncertain that this fix is in the right layer: It works around when calling `vim` as Git editor, but only in Git for Windows! What about Linux, macOS? What about _other_ editors?

In essence, I am beginning to suspect that the most appropriate place to fix the underlying issue is [`launch_specified_editor()`](https://github.com/git/git/blob/v2.49.0/editor.c#L60-L125) instead. There's a slight complication there in that it already takes an `env` parameter _and_ it is not a `strvec` but a fixed, constant array (but then, in [all the callers](https://github.com/search?q=repo%3Agit/git%20/launch(_sequence)?_editor/&type=code), there is [only one](https://github.com/git/git/blob/v2.49.0/builtin/commit.c#L1103) that actually passes a non-`NULL` `env` parameter, and it comically constructs a `strvec` already).

Of course, the most versatile way to implement this would _not even be_ in `launch_specified_editor()`, but instead only set a (to-be-introduced) flag in the `child_process` struct (something like `no_git_exec_path`) and then handle that flag [at the beginning of `start_command()`](https://github.com/git/git/blob/v2.49.0/run-command.c#L678), something along these lines:

```c
if (cmd->no_git_exec_path) {
  struct strbuf buf = STRBUF_INIT;
  const char *exec_path = git_exec_path();
  size_t len = strlen(exec_path);
  char *p;

  strbuf_addstr(&buf, getenv("PATH"));
  for (p = strstr(buf.buf, exec_path); p; p = strstr(p, exec_path)) {
    if ((p[len] && p[len] != PATH_SEP) || (p != buf.buf && p[-1] != PATH_SEP))
      p += len; /* false positive, skip */
    else {
      size_t offset = p - buf.buf, l = len;
      if (p != buf.buf) {
        /* include the preceding path separator */
        offset--;
        l++;
      } else if (p[len] == PATH_SEP) {
        /* include the path separator following GIT_EXEC_PATH */
        l++;
      }
      strbuf_splice(&buf, offset, l, "", 0);
    }
    /* now use this as `PATH` value in the child process' env */
  }
}
```

Of course, the devil lies in the details and it may well be better to integrate this logic into [`prep_childenv()`](https://github.com/git/git/blob/v2.49.0/run-command.c#L451-L498) instead (passing through the `no_git_exec_path` parameter).

benknoble:
> Thank you for contributing!

Thank you for your detailed reply!

> I was a bit on the fence after I read the PR title (which didn't manage to convince me that this is a good change, it rather did the opposite: why _on Earth_ should it be _vim_'s job to clean up the `PATH`?), but can see that this might make sense.

Right: I wasn't sure this belonged here, anyway, but my old mail never generated any conversation :)

> Since I am unfamiliar with vim's arcane script language, the code in question is too hard to understand for me, and additional code comments would be much appreciated. For example, I wonder whether this code now adds a start-up penalty to _all_ invocations of `vim` and not just those where `GIT_EXEC_PATH` is set.

I can further explain, but based on below will try to upstream a different change first.

To answer the performance question: we take a minor hit when queuing and executing the callback, but we bail out of that as fast as we can when we don't find the exec path in `PATH`. If we can rely on `GIT_EXEC_PATH` to know when to adjust `PATH` (and even which value to adjust), we can probably do even better.

As a side note, you can profile with `--startuptime <output>` to Vim: my entire 600-line vimrc takes 23ms to execute. Unfortunately, I don't _think_ `VimEnter` commands are included in this profile. Yet something like

```
vim --cmd 'profile start myprof' --cmd 'profile func fix_git_path'
# press "ZZ" once Vim opens
view myprof
```

does work to profile the function execution: when not launched by Git, this function executes for 0.000042000s (0.042ms = 42μs).

> Also, I am still uncertain that this fix is in the right layer: It works around when calling `vim` as Git editor, but only in Git for Windows! What about Linux, macOS? What about _other_ editors?
>
> In essence, I am beginning to suspect that the most appropriate place to fix the underlying issue is [`launch_specified_editor()`](https://github.com/git/git/blob/v2.49.0/editor.c#L60-L125) instead. There's a slight complication there in that it already takes an `env` parameter _and_ it is not a `strvec` but a fixed, constant array (but then, in [all the callers](https://github.com/search?q=repo%3Agit/git%20/launch(_sequence)?_editor/&type=code), there is [only one](https://github.com/git/git/blob/v2.49.0/builtin/commit.c#L1103) that actually passes a non-`NULL` `env` parameter, and it comically constructs a `strvec` already).
>
> Of course, the most versatile way to implement this would _not even be_ in `launch_specified_editor()`, but instead only set a (to-be-introduced) flag in the `child_process` struct (something like `no_git_exec_path`) and then handle that flag [at the beginning of `start_command()`](https://github.com/git/git/blob/v2.49.0/run-command.c#L678), something along these lines:
>
> ```c
> if (cmd->no_git_exec_path) {
>   struct strbuf buf = STRBUF_INIT;
>   const char *exec_path = git_exec_path();
>   size_t len = strlen(exec_path);
>   char *p;
>
>   strbuf_addstr(&buf, getenv("PATH"));
>   for (p = strstr(buf.buf, exec_path); p; p = strstr(p, exec_path)) {
>     if ((p[len] && p[len] != PATH_SEP) || (p != buf.buf && p[-1] != PATH_SEP))
>       p += len; /* false positive, skip */
>     else {
>       size_t offset = p - buf.buf, l = len;
>       if (p != buf.buf) {
>         /* include the preceding path separator */
>         offset--;
>         l++;
>       } else if (p[len] == PATH_SEP) {
>         /* include the path separator following GIT_EXEC_PATH */
>         l++;
>       }
>       strbuf_splice(&buf, offset, l, "", 0);
>     }
>     /* now use this as `PATH` value in the child process' env */
>   }
> }
> ```
>
> Of course, the devil lies in the details and it may well be better to integrate this logic into [`prep_childenv()`](https://github.com/git/git/blob/v2.49.0/run-command.c#L451-L498) instead (passing through the `no_git_exec_path` parameter).

Aha! Now we're cooking ;) I'll take a look at upstreaming this.

benknoble:
> upstreaming

v1: https://lore.kernel.org/git/20250520193506.95199-1-ben.knoble+github@gmail.com/T/#m3c581e1e9fee9b9390c4c49e41255a7b206f2c69

dscho:
> Just a little cleanup: this went upstream, I ended up making some tweaks to my setup that make it less necessary, and a few parts of that series might still be worth doing. But this can certainly get closed.

Thank you for the update!

PR #617, opened 2025-05-20
install.iss: add Microsoft Edit as an editor option

mjcheetham:
Add the newly re-written and open-source 'Edit' terminal-based text editor, inspired by the old MS-DOS Editor.

Note that at this time the edit.exe program is only available as a simple zip download, or via the WinGet package manager for Windows. Since we can only rely on a stable path for the WinGet package manager we should look there (per-user).

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -348,8 +348,9 @@ const
     GE_VSCodium                 = 7;
     GE_Notepad                  = 8;
     GE_Wordpad                  = 9;
-    GE_CustomEditor             = 10;
-    GE_MaxEditor                = 10;
+    GE_MicrosoftEdit            = 10;
+    GE_CustomEditor             = 11;
+    GE_MaxEditor                = 11;
 
     // Git Path options.
     GP_BashOnly       = 1;
@@ -472,6 +473,7 @@ var
     SublimeTextPath:String;
     AtomPath:String;
     VSCodiumPath:String;
+    MicrosoftEditPath:String;
     CustomEditorPath:String;
     CustomEditorOptions:String;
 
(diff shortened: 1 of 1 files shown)


installer/install.iss
```diff
@@ -1884,6 +1886,11 @@ begin
     Result:=ExtractFilePath(ExtractCommandPath(Command)) + Suffix;
 end;
 
+function GetMicrosoftEditPath:String;
+begin
+    Result:=ExpandConstant('{localappdata}\Microsoft\WinGet\Packages\Microsoft.Edit_Microsoft.Winget.Source_8wekyb3d8bbwe\edit-1.0.0-x86_64-windows\edit.exe');
```
mjcheetham, line 1891: Need to support ARM64, and also not hardcode the package version here.

dscho (commented):
Looks good so far!

rimrul:
> Note that at this time the edit.exe program is only available as a simple zip download, or via the WinGet package manager for Windows.

There might soon be a [MINGW-package we could theoretically use](https://github.com/msys2/MINGW-packages/pull/24309)

ognevny:
> > Note that at this time the edit.exe program is only available as a simple zip download, or via the WinGet package manager for Windows.
> 
> There might soon be a [MINGW-package we could theoretically use](https://github.com/msys2/MINGW-packages/pull/24309)

I wanted to check one extra thing and think about package name more. `edit` seems a lot generic, it can be `ms-edit` or something

mati865:
Also, it sort of conflicts with `/usr/bin/edit` which, albeit absent in MSYS2, is present in most of Linux distributions.

ognevny:
> Also, it sort of conflicts with `/usr/bin/edit` which, albeit absent in MSYS2, is present in most of Linux distributions.

there shouldn't be conflict in case of MINGW environments. ${MINGW_PREFIX}/bin/edit.exe will be preferred anyway

PR #626, opened 2025-06-21
astextplain: restore passthrough for plaintext DOC/DOT files

ClaudioHoffmann:
The fix for https://github.com/git-for-windows/git/issues/5641 broke diffs for plaintext DOC and DOT files:

    $ echo Test > test.doc
    $ git add test.doc
    $ git diff --cached
    E: unsupported filetype test.doc

Since `file` correctly identifies the mimetype of such files as `text/plain`, this can be fixed by simply covering this case and printing the file content as it was done previously.


dscho:
Thank you for the contribution! The checksums were incorrect, though, and unfortunately I lack the permission to fix it in this PR, therefore I opened https://github.com/git-for-windows/build-extra/pull/628.

PR #640, opened 2025-09-09
Add check over Windows Integration to already exists during upgrade

theclickman:
It is my suggestion for fixing Bug#5683 I opened months ago. https://github.com/git-for-windows/git/issues/5683

It will enable preserving Windows Integration in any context of update - manually (it will check the option automatically if detected) or via silent update (winget or other path) by detecting if the Registry entries already exists and match the current installation at that time.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1904,6 +1904,40 @@ begin
 #endif
 end;
 
+procedure PreselectShellIntegrationIfMatchingInstall();
+var
+  RootKey: Integer;
+  Cmd: String;
+  AppPath: String;
+begin
+  AppPath := UninstallAppPath;  // Set earlier in QueryUninstallValues()
+
+  if AppPath = '' then
+    Exit;
+
+  if IsAdminLoggedOn then
+    RootKey := HKEY_LOCAL_MACHINE
+  else
+    RootKey := HKEY_CURRENT_USER;
+
+  // Git Bash Here
+  if RegQueryStringValue(RootKey,
+    'SOFTWARE\Classes\Directory\shell\git_shell\command', '', Cmd) then
+  begin
+    if Pos(AppPath, Cmd) > 0 then
+      WizardSelectComponents('ext\shellhere');
+  end;
+
+  // Git GUI Here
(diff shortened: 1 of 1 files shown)


dscho (changes requested):
> Add check over Windows Integration to already exists during upgrade

Even though I tried for several minutes to parse this message, and I am still failing. What is this supposed to tell the reader?

> It is my suggestion for fising Bug 5683 I opened months ago. https://github.com/git-for-windows/git/issues/5683

This might help you understand what this commit is about. Leaving everybody else puzzled. So please follow the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to improve it, in particular with a strong focus on this part:

  | What you’re doing | Why you’re doing it
-- | -- | --
High-level (strategic) | **Intent** (what does this accomplish?) | **Context** (why does the code do what it does now?)
Low-level (tactical) | **Implementation** (what did you do to accomplish your goal?) | **Justification** (why is this change being made?)

> It will enable preserving Windows Integration in any context of update - manually (it will check the option automatically if detected) or via silent update (winget or other path) by detecting if the Registry entries already exists and match the current installation at that time. 

Again, I have a _really_ tough time trying to understand what you are saying.

> Signed-off-by: Nicolas DENIS <38191912+theclickman@users.noreply.github.com>

If you don't want to provide your real email address, it is not a real Signed-off-by: footer. The purpose of including the email address in the [DCO](https://git-scm.com/docs/SubmittingPatches#dco) is so that you can be contacted in case there are questions about your changes.

Apart from the commit message, it strikes me as a work-around that addresses one tiny fraction of the problem that `winget` updates fail to preserve the user's choices. _Two_ of those choices, to be precise. All the other user choices are missed.

Maybe a more fruitful approach would be to find out why the registry setting `Inno Setup: Selected Components` in `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Git_is1` is not respected.

theclickman:
Cancelled and replace with Pull Request #641

PR #641, opened 2025-09-10
Add UsePreviousSetupType to Setup section to preserve settings during upgrade

theclickman:
The intend here is to resolve the issue when settings are not carried over when upgrading in silent mode - the most visible were the Windows Integration to me but it can impact other settings as well. 

After reviewing documentation, I found out that UsePreviousSetupType directive should handle it, but was missing from the installer script.

As per documentation: 
When this directive is yes, the default, at startup Setup will look in the registry to see if the same application is already installed, and if so, it will use the setup type and component settings of the previous installation as the default settings presented to the user in the wizard. Note that Setup cannot re-use settings from a previous installation that had Uninstallable set to no, since the registry entries it looks for are not created when Uninstallable is no. -- end of documentation.

As discussed on my previous attempt, here is the new one trying to account for the proper way to carry over any and all settings, even future one.

Signed-off-by: Nicolas DENIS <the.clickman@gmail.com>

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -85,6 +85,7 @@ PrivilegesRequired=none
 #endif
 UninstallDisplayName={#APP_NAME}
 UninstallDisplayIcon={app}\{#MINGW_BITNESS}\share\git\git-for-windows.ico
+UsePreviousSetupType=yes
 #ifndef COMPILE_FROM_IDE
 VersionInfoVersion={#FILE_VERSION}
 #endif


dscho (commented):
@rimrul what's your take on this? I cannot really wrap my head around the risk side of this PR... especially _this close_ to v2.51.0(2)...

martijnlaan:
UsePreviousSetupType defaults to yes, so just adding a line saying UsePreviousSetupType=yes like this PR seems to do makes no difference.

Also the comment about uninstall and reinstall is confusing. If you uninstall all storage of previous settings is lost.

Perhaps I am misunderstanding things.

theclickman:
hi, I agree... that maybe unnecessary. (But sometimes, setting default value explicitely make sure behavior do not change). But I digress.

The issue ultimately lie in the way the setup of GIT is declared as part of winget - I did try again, but in vesbose mode to try to determine what winget is actually doing.
Winget do execute the installer as:

2025-09-26 19:53:52.659 [CLI ] Starting: 'C:\Users\Nicolas\AppData\Local\Temp\WinGet\Git.Git.2.51.0\Git-2.51.0-64-bit.exe' with arguments '/SP- /SILENT /SUPPRESSMSGBOXES /NORESTART /LOG="C:\Users\Nicolas\AppData\Local\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir\WinGet-Git.Git.2.51.0-2025-09-26-19-53-52.659.log" /COMPONENTS=icons,ext,gitlfs,assoc,assoc_sh,consolefont,windowsterminal,scalar'

With hardcoded components. That is why winget is killing my config everytime, as it force the settings it use. Would it be possible to remove that from the manifest used when you submit the update to Microsoft ? 
It should not be required to have it - either it is default in the installer from the configuration so new install take the default, or it is an upgrade and we get to preserve whatever was elected originally. 

You can keep or kill the PR as you see fit - defaulting to the default won't hurt anyway - but only asking to get the winget declaration to be improved (I could not find it in the repo to suggest an update to it)

theclickman:
any idea who is maintaining it ? cannot say from the Microsoft winget repo as it come from a bot.

PR #669, opened 2026-01-08
install.iss: add Microsoft Edit as an editor option

rimrul:
Add the newly re-written and open-source 'Edit' terminal-based text editor, inspired by the old MS-DOS Editor.

While the edit.exe program is also available as a simple zip download, or via the WinGet package manager for Windows, it is also distributed via Windows Update on Windows 11 24H2 and newer. Since that's the only system-wide install with a stable path, we currently only support that install method.

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -348,8 +348,9 @@ const
     GE_VSCodium                 = 7;
     GE_Notepad                  = 8;
     GE_Wordpad                  = 9;
-    GE_CustomEditor             = 10;
-    GE_MaxEditor                = 10;
+    GE_MicrosoftEdit            = 10;
+    GE_CustomEditor             = 11;
+    GE_MaxEditor                = 11;
 
     // Git Path options.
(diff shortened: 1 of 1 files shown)


mjcheetham (approved):
> it is also distributed via Windows Update on Windows 11 24H2 and newer.

Ah that's great! Makes it easier to reason about the availability of the editor this way.

dscho (approved):
Technically, we should probably set `core.editor` to the full path of `edit.exe`, but I guess this is good enough!

lhecker:
Thank you so much for opening this PR. I was very happy seeing the editor today when I installed Git!

I do have one question though: Is there a reason an absolute path to edit.exe is used? Why not use just "edit.exe" and let the system find it in the PATH? I'm currently thinking about ways to allow people to override the system edit.exe when installing winget and one idea I had was to installed it into ProgramFiles and put it in the PATH even before System32. If Git used just "edit.exe" the winget version would get picked up automatically.

PR #682, opened 2026-03-13
installer: refine git pull behavior text to prevent misconfiguration

jzbakh:
### Description
The previous phrasing of the `git pull` behavior defaults could lead to unintended configurations of the `pull.rebase` vs `pull.ff` parameters in the global `.gitconfig`. This PR refactors the text to introduce strict mechanical representations of the git state machine. The goal is to reduce ambiguity concerning commit graph resolution and align the UI text with current upstream defaults.

### Technical Scope
* **Textual and UI Layout Update:** This modification strictly targets the strings passed to the `CreatePage` and `CreateRadioButton` functions for the `GP_GitPullMerge`, `GP_GitPullRebase`, and `GP_GitPullFFOnly` options. 
* **Added Label:** A standard `TLabel` (`LblInfo`) was added at the bottom of the page to clarify how Git handles file conflicts.
* **Logic Intact:** No underlying Pascal Script conditional logic, registry parsing (`ReplayChoice`), or silent install vectors were altered. The internal component IDs remain untouched.

### Visual Regression Testing

| Before | After |
|---|---|
| git_pull_behavior | git_pull_behavior_2 |

Signed-off-by: Jason Zbakh <jzbakh@gmail.com>

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -1917,6 +1917,7 @@ var
     LblInfo:TLabel;
     AslrSetting: AnsiString;
     WasFSMonitorEnabled:Boolean;
+    Dummy:array of TLabel;
 begin
     SanitizeGitEnvironmentVariables();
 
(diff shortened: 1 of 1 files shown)


JZBAKH:
Reopened: head repository restored after accidental deletion.

installer/install.iss
```diff
@@ -2557,8 +2557,10 @@ begin
 #endif
 
     PageIDBeforeInstall:=CurrentCustomPageID;
+#ifdef HAVE_EXPERIMENTAL_OPTIONS
     if (PageIDBeforeInstall=ExperimentalOptionsPage.ID) and IsHiddenExperimentalOptionsPageEmpty then
         PageIDBeforeInstall:=PageIDBeforeInstall-1;
+#endif
```
dscho, line 2563: I am a big fan of the principle to separate concerns into separate commits. In this instance, the change to allow for no experimental options is separate from the change to describe the `git pull` options better.

However, I don't want this split to be a blocker for this PR, so I'll merge it as-is.

PR #689, opened 2026-04-14
Skip distributing `git instaweb`

dscho:
While [analyzing Perl users](https://github.com/git-for-windows/git/issues/5393#issuecomment-4242978450), I noticed that `git instaweb` is included in Git for Windows, even if GitWeb (on which it relies) is not.

Let's fix that.


rimrul (approved):

dscho:
Gah, I wanted to include this in -rc2. Oh well.

dscho:
/add relnote bug The `git instaweb` command [is no longer distributed with Git for Windows](https://github.com/git-for-windows/build-extra/pull/689) because it would require GitWeb (which has not been distributed with Git for Windows for quite a few years).

The workflow run [was started](https://github.com/git-for-windows/build-extra/actions/runs/24424114540)

PR #695, opened 2026-05-06
pacman-helper: maintain per-package .versions.json index files

dscho:
The Pacman repository for Git for Windows migrated from Azure Blob Storage to branches in git-for-windows/pacman-repo. Unlike the previous setup where all package versions accumulated in the same container, the Git-based repository only retains the latest version of each package in its branch. Older versions are still available as GitHub Release assets, but there was no way to discover which Release holds a particular version.

This has already caused CI failures when a job needed a package version that had been superseded, e.g. https://github.com/git-for-windows/git-sdk-64/actions/runs/25404775441/job/74515324990#step:7:28

This PR adds per-package `.versions.json` index files to each architecture branch of pacman-repo, mapping every version-pkgrel to the GitHub Release tag where the `.pkg.tar` asset can be downloaded.

### Changes

**pacman-helper.sh**: Maintains `.versions.json` during deployment. The `update_versions_json` helper uses `sed`/`sort` (no perl) to read, insert, sort, and write entries. In the concurrent-deployment rebase path, `.versions.json` changes are stripped before rebase and re-inserted afterwards to avoid conflicts.

**please.sh**: Replaces hard-coded curl 8.18.0-1 workarounds with a `.versions.json` lookup for PDB package downloads. Falls back to the branch URL when the index is unavailable.

**get-sources.sh**: Uses the x86_64 binary package's `.versions.json` to deduce the Release tag for source tarballs, making source packages downloadable from GitHub Releases ahead of turning off Azure Blob Storage.

### pacman-repo seeding

The three architecture branches have been seeded with `.versions.json` files built from all 1172 existing releases (97 packages / 2796 versions for x86_64, 112 / 2696 for i686, 46 / 932 for aarch64): git-for-windows/pacman-repo@c5122e6 git-for-windows/pacman-repo@84eb68d git-for-windows/pacman-repo@cfcc281

### DIFF
--- a/get-sources.sh
+++ b/get-sources.sh
@@ -29,6 +29,13 @@ sf_repos_url=http://sourceforge.net/projects/msys2/files/REPOS
 msys_sf_source_url=$sf_repos_url/MSYS2/Sources
 mingw_sf_source_url=$sf_repos_url/MINGW/Sources
 azure_blobs_source_url=https://wingit.blob.core.windows.net/sources
+pacman_repo_url=https://raw.githubusercontent.com/git-for-windows/pacman-repo/refs/heads
+
+# When $GITHUB_TOKEN is non-empty, pass it via Basic Auth so the otherwise
+# strictly rate-limited raw.githubusercontent.com (and api.github.com release
+# download) endpoints accept more requests.
+github_curl_auth=
+test -z "$GITHUB_TOKEN" || github_curl_auth="-u x-access-token:$GITHUB_TOKEN"
 
 cd "$(dirname "$0")" ||
 die "Could not change directory to build-extra/"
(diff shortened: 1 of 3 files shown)


dscho:
We should merge this _before_ the next `/deploy`, of course, otherwise the `.versions.json` information will have to be backfilled manually.

get-sources.sh
```diff
@@ -29,6 +29,7 @@ sf_repos_url=http://sourceforge.net/projects/msys2/files/REPOS
 msys_sf_source_url=$sf_repos_url/MSYS2/Sources
 mingw_sf_source_url=$sf_repos_url/MINGW/Sources
 azure_blobs_source_url=https://wingit.blob.core.windows.net/sources
+pacman_repo_url=https://raw.githubusercontent.com/git-for-windows/pacman-repo/refs/heads
```
mjcheetham, line 32: Are we concerned about hitting usage limits unauthenticated?

mjcheetham (approved):

get-sources.sh
dscho: Hrm. That's a good point. Raw endpoint _is_ rate-limited without a token.

How about using a `GITHUB_TOKEN` if it is found in the environment?

dscho: > How about using a `GITHUB_TOKEN` if it is found in the environment?

That's what I've now done: https://github.com/git-for-windows/build-extra/compare/ee467ea72c80221a58043914f39db68b65e6a790..5b44c34581782c792de7d5236b7bac6762efe9e4

mjcheetham (approved):

dscho:
Mental note: this will need a follow-up PR in git-for-windows-automation, to pass GITHUB_TOKEN in `git-artifacts` when building the installer (for which the `.pdb` files e.g. for `mingw-w64-curl` need to be fetched).

PR #697, opened 2026-05-11
Ensure Bash-specific completion commands run only in Bash

victordou:
Modified git-prompt.sh to wrap Bash-only constructs with checks for $BASH_VERSION. This prevents errors when the script is sourced in non-Bash shells and maintains compatibility with MSYS2 environments and Wine.

- Wrap shopt and completion script evaluation with Bash check
- Add Bash check to WINELOADERNOEXEC conditionals

I'm a Windows user, and I'm using Zsh on top of git for windows, this fix the error when you run Zsh -i -l directly.

### DIFF
--- a/git-extra/PKGBUILD
+++ b/git-extra/PKGBUILD
@@ -5,7 +5,7 @@ pkgbase="mingw-w64-${_realname}"
 pkgname=($_realname
         "${MINGW_PACKAGE_PREFIX}-${_realname}")
 _ver_base=1.1
-pkgver=1.1.689.fe8c5259c
+pkgver=1.1.691.e78bebdf5
 pkgrel=1
 pkgdesc="Git for Windows extra files"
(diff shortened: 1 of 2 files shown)


dscho:
/updpkgsums

The workflow run [was started](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/25665405333).

dscho:
/deploy git-extra

The [i686/x86_64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/25873429972) and the [arm64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/25873431038) workflow runs were started.

dscho:
@victordou thank you! I do not believe that there are many users out there who use Git for Windows with a Unix shell other than Bash, but it is good to receive a contribution from one of those users!

PR #699, opened 2026-05-20
check-for-missing-dlls: fixes for ARM64

rimrul:
Our `check-for-missing-dlls` workflow in the git-sdk-arm64 repo currently does nothing. These fixes should make it work as expected.

### DIFF
--- a/check-for-missing-dlls.sh
+++ b/check-for-missing-dlls.sh
@@ -124,7 +124,7 @@ grep '\.dll$' "$tmp_file.all" |
 		-e '^usr/lib/sasl2/' \
 		-e '^usr/lib/coreutils/libstdbuf.dll' \
 		-e "^$MINGW_PREFIX/bin/libcurl\(\|-openssl\)-4.dll" \
-		-e "^$MINGW_PREFIX/bin/\(atlassian\|azuredevops\|bitbucket\|gcmcore.*\|github\|gitlab\|microsoft\|newtonsoft\|system\..*\|webview2loader\|avalonia\|.*harfbuzzsharp\|microcom\|.*skiasharp\|av_libglesv2\|msalruntime\(\|_x86\|arm64\)\)\." \
+		-e "^$MINGW_PREFIX/bin/\(atlassian\|azuredevops\|bitbucket\|gcmcore.*\|github\|gitlab\|microsoft\|newtonsoft\|system\..*\|webview2loader\|avalonia\|.*harfbuzzsharp\|microcom\|.*skiasharp\|av_libglesv2\|msalruntime\(\|_x86\|_arm64\)\)\." \
 		-e "^$MINGW_PREFIX/lib/ossl-modules/" \
 		-e "^$MINGW_PREFIX/lib/\(engines\|reg\|thread\)" |
 	sed 's/^/unused dll: /' |
--- a/check-for-missing-dlls.sh
+++ b/check-for-missing-dlls.sh
@@ -129,10 +129,10 @@ grep '\.dll$' "$tmp_file.all" |
 		-e '^usr/lib/openssl/engines' \
 		-e '^usr/lib/sasl2/' \
 		-e '^usr/lib/coreutils/libstdbuf.dll' \
-		-e '^mingw../bin/libcurl\(\|-openssl\)-4.dll' \
-		-e '^mingw../bin/\(atlassian\|azuredevops\|bitbucket\|gcmcore.*\|github\|gitlab\|microsoft\|newtonsoft\|system\..*\|webview2loader\|avalonia\|.*harfbuzzsharp\|microcom\|.*skiasharp\|av_libglesv2\|msalruntime\(\|_x86\|arm64\)\)\.' \
-		-e '^mingw../lib/ossl-modules/' \
-		-e '^mingw../lib/\(engines\|reg\|thread\)' |
+		-e "^$MINGW_PREFIX/bin/libcurl\(\|-openssl\)-4.dll" \
+		-e "^$MINGW_PREFIX/bin/\(atlassian\|azuredevops\|bitbucket\|gcmcore.*\|github\|gitlab\|microsoft\|newtonsoft\|system\..*\|webview2loader\|avalonia\|.*harfbuzzsharp\|microcom\|.*skiasharp\|av_libglesv2\|msalruntime\(\|_x86\|arm64\)\)\." \
+		-e "^$MINGW_PREFIX/lib/ossl-modules/" \
+		-e "^$MINGW_PREFIX/lib/\(engines\|reg\|thread\)" |
 	sed 's/^/unused dll: /' |
 	tee "$unused_dlls_file" >&2
 
--- a/check-for-missing-dlls.sh
+++ b/check-for-missing-dlls.sh
@@ -27,14 +27,20 @@ die "Could not determine architecture"
 
 case "$MSYSTEM" in
 MINGW64) MINGW_PREFIX=mingw64;;
-CLANGARM64) MINGW_PREFIX=clangarm64;;
+CLANGARM64)
+	MINGW_PREFIX=clangarm64
+	ARCH=aarch64
+	;;
 MINGW32) MINGW_PREFIX=mingw32;;
 *)
 	case "$ARCH" in
 	i686) MINGW_PREFIX=mingw32;;
 	x86_64)
 		case $(uname -s) in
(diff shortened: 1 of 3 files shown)


rimrul:
```
/clangarm64/lib/engines-3/capi.dll is missing capi.dll
/clangarm64/lib/engines-3/loader_attic.dll is missing loader_attic.dll
/clangarm64/lib/engines-3/padlock.dll is missing padlock.dll
 is missing legacy.dll
 is missing p11-kit-trust.dll
 is missing tclreg13.dll
 is missing thread2812.dll
```

interesting.

rimrul:
> unused dll: clangarm64/bin/libnghttp2-14.dll

This seems odd. Curl probably uses libnghttp.

> unused dll: clangarm64/bin/libunwind.dll
unused dll: clangarm64/bin/libwinpthread-1.dll

These seem really odd.

> unused dll: clangarm64/bin/msalruntime_arm64.dll

This one seems to be just a missing `_` in the regex.

dscho:
> > unused dll: clangarm64/bin/libnghttp2-14.dll
> 
> This seems odd. Curl probably uses libnghttp.

Right... [here](https://github.com/git-for-windows/MINGW-packages/blob/842fc4bb4be65f90624f4b5b9591326cd56a6561/mingw-w64-curl/PKGBUILD#L17-L18) and [here](https://github.com/git-for-windows/MINGW-packages/blob/842fc4bb4be65f90624f4b5b9591326cd56a6561/mingw-w64-curl/PKGBUILD#L137).

> > unused dll: clangarm64/bin/libunwind.dll
> > unused dll: clangarm64/bin/libwinpthread-1.dll
> 
> These seem really odd.

Indeed. It's _possible_, of course, that they are usually only pulled in via GCC, but why are they there? If only I had a convenient way to spin up a Windows/ARM64 VM I could investigate a lot better :-(

> > unused dll: clangarm64/bin/msalruntime_arm64.dll
> 
> This one seems to be just a missing `_` in the regex.

Good point.

dennisameling:
> Indeed. It's _possible_, of course, that they are usually only pulled in via GCC, but why are they there? If only I had a convenient way to spin up a Windows/ARM64 VM I could investigate a lot better :-(

@dscho happy to test some things on my Surface Pro X, but currently a bit time-constrained. I have the GfW arm64 SDK already installed on it. Do you mind giving me some instructions on what exactly you'd like to test? Thanks!

dscho:
> > Indeed. It's _possible_, of course, that they are usually only pulled in via GCC, but why are they there? If only I had a convenient way to spin up a Windows/ARM64 VM I could investigate a lot better :-(
> 
> `@dscho` happy to test some things on my Surface Pro X, but currently a bit time-constrained. I have the GfW arm64 SDK already installed on it. Do you mind giving me some instructions on what exactly you'd like to test? Thanks!

@dennisameling Thank you for offering! But I think that I actually now have a relatively good way to let Copilot run in CI. So if there's any more problems, I will just add a step where Copilot gets a prompt and has to churn through all the remaining problems. It will run on the actual Actions runner and therefore will have a proper Windows/ARM64 environment that I lack.

dscho:
> [...] I think that I actually now have a relatively good way to let Copilot run in CI. So if there's any more problems, I will just add a step where Copilot gets a prompt and has to churn through all the remaining problems. It will run on the actual Actions runner and therefore will have a proper Windows/ARM64 environment that I lack.

For lurkers, this commit implements that strategy that I was talking about: https://github.com/git-for-windows/build-extra/commit/628a5e1261cdad244d1a71297d01932160ada89f

dscho:
And in case it was unclear why I run this Copilot experiment in a separate PR: The prerequisite token is stored as an Action secret, which is unavailable in cross-fork PRs' builds.

dscho:
@rimrul I'm pretty happy with the current shape of the PR. How about you?

rimrul:
I'm happy with the shape, but want to take a closer look at 7f10fcb2e08c6524dfa5e188ddfb1da6f3532e6b. I'll hopefully get around to that in a few hours.

dscho:
> I'm happy with the shape, but want to take a closer look at [7f10fcb](https://github.com/git-for-windows/build-extra/commit/7f10fcb2e08c6524dfa5e188ddfb1da6f3532e6b). I'll hopefully get around to that in a few hours.

I found a couple of gaps that I closed: delay-load DLLs were not listed, and some incorrect offsets were used (even if in practice, well-formed `.exe` files would render these bugs harmless). Or in more detail (thanks, Opus):

1. `VirtualSize` vs `SizeOfRawData`: Section lookup now uses `VirtualSize` (offset +8) instead of `SizeOfRawData` (offset +16) for RVA containment, per the PE spec.
2. Delay-load imports: Data directory index 13 is now parsed using the same `$walkImports` helper, with 32-byte descriptors and Name RVA at offset +4.

check-for-missing-dlls.sh
```diff
@@ -82,26 +82,20 @@ do
 	paths=$(sed -ne 's,[][],\\&,g' -e "s,^$dir[^/]*\.\(dll\|exe\)$,/&,p" "$tmp_file.all")
 	/usr/bin/objdump -p $paths 2>"$tmp_file" >"$tmp_file.ldd"
 	paths="$(sed -n 's|^/usr/bin/objdump: \([^ :]*\): file format not recognized|\1|p' <"$tmp_file")"
-	test -z "$paths" ||
-	ldd $paths >>"$tmp_file.ldd"
+	if test -n "$paths"
+	then
+		powershell.exe -NoProfile -ExecutionPolicy Bypass \
+			-File "$thisdir/pe-imports.ps1" $paths >>"$tmp_file.ldd" ||
```
rimrul, line 88: Keeping the file called `.ldd` is gonna be a weird historical artifact, but it's probably fine.

pe-imports.ps1
```diff
@@ -0,0 +1,134 @@
...
+            $bytes[0] -ne 0x4D -or $bytes[1] -ne 0x5A) {
+            Write-Error "pe-imports: ${file}: not a PE file"
+            continue
+        }
+
+        $peOffset = [BitConverter]::ToInt32($bytes, 0x3C)
+        if ($peOffset + 24 -gt $bytes.Length -or
+            $bytes[$peOffset] -ne 0x50 -or $bytes[$peOffset+1] -ne 0x45 -or
+            $bytes[$peOffset+2] -ne 0 -or $bytes[$peOffset+3] -ne 0) {
+            Write-Error "pe-imports: ${file}: bad PE signature"
```
rimrul, line 32: technically the signature could be fine and some of the COFF header cut off, but I don't think we need an extra error message for that specific case

pe-imports.ps1
```diff
@@ -0,0 +1,134 @@
...
+            param([uint32]$descRVA, [int]$descSize, [int]$nameField)
+            $p = & $rvaToOffset $descRVA
+            if ($p -lt 0) { return }
+
+            while ($p + $descSize -le $bytes.Length) {
+                $nRVA = [BitConverter]::ToUInt32($bytes, $p + $nameField)
+                if ($nRVA -eq 0) { break }
+
+                $nOff = & $rvaToOffset $nRVA
+                if ($nOff -lt 0 -or $nOff -ge $bytes.Length) { break }
```
rimrul, line 117: This loop feels weird to me. The directory entries come with the [RVAs and a size](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#optional-header-data-directories-image-only), but we seem to be relying on the data after import table/delay import descriptor to produce a bogus RVA that produces a bogus offset to get us out of this loop.

check-for-missing-dlls.sh
dscho: True. It's a historical wart, but it doesn't hurt much, does it? To the contrary, for people familiar with Linux, `ldd` will give the exact right hint as to what this is all about...

pe-imports.ps1
dscho: True. There _is_ something to be said about those Null entries mentioned at https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#the-idata-section, but you're right, we do have a size to work with, so it's wise to make use of it.

dscho:
@rimrul thank you for prodding me. After looking at other code and then coming back to the script, I did find a few more confusing parts, and I also found myself wishing for more links into the PE file format specification, so I added those and tried my best to unconfuse the script. How does it look to you now?

rimrul:
LGTM

PR #700, opened 2026-05-22
check-for-missing-dlls: ARM64 fixes with Copilot debugging

dscho:
This PR builds on rimrul's #699 (check-for-missing-dlls ARM64 fixes) and
adds a Copilot CLI debugging step that runs on the ARM64 runner when the
check-for-missing-dlls step fails. The Copilot session has access to the
full SDK artifact and can diagnose and verify fixes end-to-end on the
actual ARM64 hardware.

The key additions beyond #699 are:

- `pe-imports.ps1`: a PowerShell PE import-table parser that replaces
  the broken `ldd` fallback. On ARM64, the MSYS2 objdump cannot parse
  pei-aarch64 PE files, and ldd under WoW64 emulation gives incomplete
  results. The PowerShell script reads the PE data structures directly
  and outputs in objdump-compatible format.

- A fix for the msalruntime exclusion regex (missing underscore before
  `arm64`).

- A TO-DROP commit that adds a Copilot CLI debugging step, triggered on
  failure of the ARM64 job, to diagnose any remaining issues on the
  actual runner.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -399,3 +399,35 @@ jobs:
       - name: check for missing DLLs (MinGit)
         shell: bash
         run: MINIMAL_GIT=1 ./check-for-missing-dlls.sh
+      - name: Set up Copilot CLI
+        if: failure() && matrix.arch.name == 'aarch64'
+        shell: bash
+        run: npm install -g @github/copilot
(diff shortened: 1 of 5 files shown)


dscho:
No longer needed; fixes pushed to PR #699.

PR #701, opened 2026-05-25
Add Pulsar as a default editor option along side atom

BananaDev2000:
I thought as Atom is deprecated it might be better to add Pulsar.
I decided to be not destructive so I added it next to Atom in the list everywhere.
Any suggestions or feedback is welcome maybe removing Atom entirely is better?

I tested the command using a wait, but maybe not a bad idea if it could be testing in a build/compiled installer?

### DIFF
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -345,12 +345,13 @@ const
     GE_VisualStudioCodeInsiders = 4;
     GE_SublimeText              = 5;
     GE_Atom                     = 6;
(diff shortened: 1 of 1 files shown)


dscho:
First time I hear of Pulsar... Maybe Pulsar users can use the fall-back config instead of needing a dedicated option?

dscho:
I take that as a "yes, it can".

PR #703, opened 2026-05-27
Fix `sdk init msys2-runtime`

dscho:
The `sdk init msys2-runtime` function tries to helpfully install an `msys2` remote. The URL for that has changed a _long_ time ago, but while it helpfully redirected for years, now the old URL 404s. Let's use the new URL instead.


dscho:
/deploy git-extra

The [i686/x86_64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/26540086578) and the [arm64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/26540087399) workflow runs were started.

PR #706, opened 2026-05-30
Prepare pacman helper for ucrt64

rimrul:
Some more preparation work for https://github.com/git-for-windows/git/issues/3167.

We still need an initial `git-for-windows-ucrt64.db.tar.xz` in https://github.com/git-for-windows/pacman-repo/tree/x86_64 before this can work, though.

### DIFF
--- a/pacman-helper.sh
+++ b/pacman-helper.sh
@@ -249,6 +249,7 @@ quick_action () { # <action> <file>...
 		mingw-w64-*.pkg.tar.xz|mingw-w64-*.pkg.tar.zst)
 			arch=${file##mingw-w64-}
 			arch=${arch#clang-}
+			arch=${arch#ucrt-}
 			arch=${arch%%-*}
 			key=${arch}_mingw
 			;;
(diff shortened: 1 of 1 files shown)


dscho (approved):
Very nice!

dscho:
> We still need an initial `git-for-windows-ucrt64.db.tar.xz` in https://github.com/git-for-windows/pacman-repo/tree/x86_64 before this can work, though.

Do we? It _might_ work without that...

But before starting to build UCRT64 packages, I'd like to stop deploying to the Azure Blob-backed Pacman repository. Let me start a session for that.

rimrul:
> > We still need an initial `git-for-windows-ucrt64.db.tar.xz` in https://github.com/git-for-windows/pacman-repo/tree/x86_64 before this can work, though.
> 
> Do we? It _might_ work without that...

Yes, I did a local dry run. The script does fail if it can't download the existing database.

dscho:
> > > We still need an initial `git-for-windows-ucrt64.db.tar.xz` in https://github.com/git-for-windows/pacman-repo/tree/x86_64 before this can work, though.
> > 
> > 
> > Do we? It _might_ work without that...
> 
> Yes, I did a local dry run. The script does fail if it can't download the existing database.

Fair enough. Should be easy to add via a PR ;-)

BTW I noticed that you haven't marked this PR as "ready for review". Did you want to do anything else with it before merging?

dscho:
> > > > We still need an initial `git-for-windows-ucrt64.db.tar.xz` in https://github.com/git-for-windows/pacman-repo/tree/x86_64 before this can work, though.
> > > 
> > > 
> > > Do we? It _might_ work without that...
> > 
> > 
> > Yes, I did a local dry run. The script does fail if it can't download the existing database.
> 
> Fair enough. Should be easy to add via a PR ;-)

I just realized something, while working on a PR to build `git-extra` also for UCRT64: We do not actually need a separate Pacman database for ucrt64. But we _can_, of course. But if we don't want to, we can simply put the UCRT64 packages also into `git-for-windows-mingw64`... Not the _cleanest_, but it would have been _possible_ ;-)

PR #708, opened 2026-06-05
Replace docx2txt with the Rust docx2txt-rs for 64-bit installers

dscho:
Replaces the Perl `docx2txt` MSYS package with the Rust `mingw-w64-docx2txt-rs` rewrite (https://packages.msys2.org/packages/mingw-w64-x86_64-docx2txt-rs, source at https://github.com/dscho/docx2txt-rs) for the 64-bit Git for Windows flavors. The new binary produces byte-identical output to the original on every fixture and drops the Perl dependency entirely.

To this end, the `docx2txt-rs` package is added as a dependency of `git-extra`; This will cause the (64-bit) Git for Windows SDKs to pull in that package. (32-bit does not matter, we do not include `docx2txt` in MinGit, and 32-bit `docx2txt-rs` would not be available anyway.)


mjcheetham (approved):

rimrul:
How does this degrade on Windows 8.1?

rimrul:
> How does this degrade on Windows 8.1?

It seems to gracefully fall back to the `perl` script as long as we keep shipping `perl`.

dscho:
@rimrul thank you for checking! I do not currently have any Windows 8.1 setup to test this...

The gentle downgrade path you verified is actually really good! That way, v2.55 can ship with both the Perl and the Rust version, and users already get the benefit of the latter as long as they're not on EOLed Windows versions ;-)

dscho:
Oh, but that means I should revert https://github.com/git-for-windows/build-extra/pull/708/commits/44b1470a2ec3f2bea7bb30767634530e8815e7b4 for now, right?

dscho:
> Oh, but that means I should revert [44b1470](https://github.com/git-for-windows/build-extra/commit/44b1470a2ec3f2bea7bb30767634530e8815e7b4) for now, right?

Done.

dscho:
/deploy git-extra

The [i686/x86_64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/27336077506) and the [arm64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/27336078771) workflow runs were started.

dscho:
/add relnote feature The diff helper handling Word documents was [ported](https://github.com/git-for-windows/build-extra/pull/708) from Perl to Rust.

The workflow run [was started](https://github.com/git-for-windows/build-extra/actions/runs/27336635246)

dscho:
> > Oh, but that means I should revert [44b1470](https://github.com/git-for-windows/build-extra/commit/44b1470a2ec3f2bea7bb30767634530e8815e7b4) for now, right?
> 
> Done.

For the record, I moved this into a follow-up PR that I intend to merge after Git for Windows v2.55: #713

rimrul:
> /add relnote feature The diff helper handling Word documents was https://github.com/git-for-windows/build-extra/pull/708 from Perl to Rust.

Maybe we should mention that we're currently keeping the perl script as a fallback for Windows 8.1.

dscho:
> > /add relnote feature The diff helper handling Word documents was #708 from Perl to Rust.
> 
> Maybe we should mention that we're currently keeping the perl script as a fallback for Windows 8.1.

Let's first see whether anyone screams? ;-)

PR #717, opened 2026-06-17
git-extra.install.in: fix failures when run outside Pacman

dscho:
When `please.sh` sources the install script and calls `post_install` directly (as it does during MinGit and installer builds), neither Pacman's `$arch` variable nor `builtins.txt`-listed files are guaranteed to exist. This causes two distinct failures that surfaced when testing UCRT64 git-artifacts builds:

The first commit guards the builtin-replacement `rm` with `test ! -f` so that the loop does not fail when installing `git-extra` into a fresh prefix where no prior Git installation exists. Without the guard, every builtin listed in `builtins.txt` produces a "No such file or directory" error that cascades into `please.sh`'s `die "Could not install packages"`.

The second commit replaces the three remaining `$arch` shell variable references (a bintray URL migration, the pacman sync database rename loop, and the i686 libunistring workaround) with the `@@MSYSTEM_CARCH@@` build-time template variable. These were missed in 50464f3a ("git-extra: hard-code which Pacman repositories to add at build-time"). The most visible symptom was a `test: syntax error: '-e' unexpected` crash at line 270 of the installed script, because the unquoted `test i686 = $arch` expanded to `test i686 = -a -e ...` when `$arch` was empty.


dscho:
/updpkgsums

The workflow run [was started](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/27701975510).

dscho:
/deploy git-extra

The [i686/x86_64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/27825689804) and the [arm64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/27825690534) workflow runs were started.

PR #728, opened 2026-08-13
arm64: inventory installer payload architectures

crutkas:
## Summary

- extend the existing architecture-neutral `pe-imports.ps1` parser with COFF machine and CLR AnyCPU reporting
- scan every installer file-list candidate regardless of filename extension
- inventory the exact ARM64 installer payload as deterministic TSV (`path`, `architecture`, `machine`, `package`, `area`) and publish it as a CI artifact
- check the current 432-file x64 set against a checked-in baseline, rejecting newly introduced x64 payloads
- provide `--strict` for the final zero-x64 release gate
- cover extension-independent discovery, deterministic output, package/area attribution, AnyCPU detection, regression detection, and strict mode with synthetic PE fixtures

The default inventory uses the same `INCLUDE_GIT_UPDATE=1` file-list configuration as `installer/release.sh`, including installer-only payloads such as WinToast. The workflow runs the inventory only for the ARM64 staging tree and leaves the existing x86/x64/UCRT installer behavior unchanged.

## Authoritative baseline provenance

The checked-in baseline and per-path historical manifest come from the latest released ARM64 payload:

- release: `v2.55.0.windows.4`
- annotated tag object: `76e15f45d670eab29f6d55611c252b3f2b70188c`
- release commit: `a93524749d7806870fd2b4b00a3812da1d6e5f4a`
- asset: `Git-2.55.0.4-arm64.tar.bz2` (171,467,184 bytes)
- asset SHA-256: `7cc28b4431c9448c310d0093fbba5646517cd702690a9b965014d7df85319ad9`
- manifest SHA-256: `ae1e311fd81258150c2300d02c58655f30b190a15bcbe3ea8bbaccc1ce8c1c9a`

| Architecture | COFF machine | Files |
| --- | --- | ---: |
| ARM64 | `0xAA64` | 209 |
| x64 | `0x8664` | 432 |
| AnyCPU | `0x014C` + CLR IL-only flags | 90 |
| x86 | `0x014C` | 1 |

The x64 set consists of 267 executables and 165 DLLs: 322 under `usr/bin`, 105 under `usr/lib`, and five elsewhere under `usr`. The release archive has no Pacman ownership database, so its historical manifest leaves `package` blank; CI-generated manifests populate package ownership from `pacman -Ql`.

## Usage

```sh
# Generate the TSV manifest and reject x64 paths not in the baseline
./check-payload-architecture.sh

# Final all-native release gate: reject every x64 payload
./check-payload-architecture.sh --strict

# Scan an extracted release from its exact tar member list
tar -tjf Git-2.55.0.4-arm64.tar.bz2 |
  sed -e '/\/$/d' -e 's|^\./||' | LC_ALL=C sort -u >release-files.txt
: >packages.txt
./check-payload-architecture.sh \
  --root=extracted-release \
  --file-list=release-files.txt \
  --package-list=packages.txt \
  --manifest=release-manifest.tsv
```

## Validation

- `./t/check-payload-architecture.sh`
- full release scan reproduced the checked-in 732-row manifest byte-for-byte after LF normalization
- x64 baseline exactly matches all 432 x64 paths in the release manifest
- import-table mode remains compatible with `check-for-missing-dlls.sh`


dscho:
What problem is this trying to solve? Have we accidentally shipped x64 binaries in `/mingw64/bin/` or something?

If so, then I suspect that we can make this dramatically simpler, by avoiding a huge baseline file and long shell scripts that mix production and test code, and instead simply state that at present, `/usr/**/*.exe` and `/usr/**/*.dll` is allowed to be x64 even in Git for Windows/ARM64, but all other `.exe`/`.dll` files in that flavor must be `arm64`. That is, slightly modify the `pe-imports.ps1` patch so that it shows the architecture _in addition_ to the imports, and then make a precise change to `check-for-missing-dlls.sh` in case of arm64 to require all but `/usr/` to have the `arm64` architecture. That would be a lot simpler to review, too.

rimrul:
I'm honestly more interested in the 91 `x86` files. The `x64` files are all in `/usr` and we'll need to switch them all over at the same time, no matter the exact set.

crutkas:
sorry, i am working on a prototype and this is me trying to fix a e2e.  Yes there are x64 binaries in Arm but it is a larger fix.  This should have been isolated to my fork and accidently wasn't contained.

PR #731, opened 2026-08-17
`render-release-notes`: fix the release-notes cache check

PtJade-Ceramic:
Fixes the broken release-notes cache check in `render-release-notes.sh`.

Problem
-------

The incremental cache check referenced a `Release.html` file that is never produced (the script writes `ReleaseNotes.html`), so the cache never hit and the release notes were always re-rendered.

Change
------

<details>
<summary>The original commit</summary>

40ebbc2 switched the cache to a content-fingerprint (sha256sum/stamp) check.

https://github.com/git-for-windows/build-extra/pull/731#pullrequestreview-5052343408
</details>
The current commit keeps only the `Release` -> `ReleaseNotes` typo fix, restoring the mtime-based check. Net effect: a previously rendered `ReleaseNotes.html` that is newer than both `ReleaseNotes.md` and this script is reused, avoiding an unnecessary re-render on repeated local runs.

Validation
----------

This change is confined to the cache-check condition; it does not alter how the HTML is rendered.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,4 @@ edit-git-bash.exe
 !/archive/root/etc/protocols
 !/archive/root/etc/services
 /versions/cached-files/
+.release-notes.stamp
--- a/.gitignore
+++ b/.gitignore
@@ -66,4 +66,3 @@ edit-git-bash.exe
 !/archive/root/etc/protocols
 !/archive/root/etc/services
 /versions/cached-files/
-.release-notes.stamp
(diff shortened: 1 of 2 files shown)


dscho (changes requested):
The only time when the script is really used is in the [`git-artifacts` workflow](https://github.com/git-for-windows/git-for-windows-automation/actions/workflows/git-artifacts.yml), when a new Git for Windows version is built.

In other words, this script is run only in a very controlled environment in which no previously-rendered release notes exist. Therefore I consider a full hash check that adds more than a one-line check quite overengineered and therefore undesirable (maintenance is hard enough without adding logic that is not needed).

I do agree that the `Release` -> `ReleaseNotes` typo is valuable, though.

dscho (approved):

PR #732, opened 2026-08-25
installer: actually build a 64-bit installer

### LINKED ISSUE #6372: Git for Windows installer is still 32-bit as of v2.55.0.5

- [x] I was not able to find an open or closed issue matching what I'm seeing

### Git for Windows version

```shell
C:\Temp>git --version --build-options
git version 2.55.0.windows.5
cpu: x86_64
built from commit: 32c4f7689275d233577576630e1ac5b7eb354eb0
sizeof-long: 4
sizeof-size_t: 8
shell-path: D:/git-sdk-64-build-installers/usr/bin/sh
rust: disabled
feature: fsmonitor--daemon
gettext: enabled
libcurl: 8.21.0
OpenSSL: OpenSSL 3.5.7 9 Jun 2026
zlib: 1.3.2
SHA-1: SHA1_DC
SHA-256: SHA256_BLK
default-ref-format: files
default-hash: sha1
```

### Windows version

Windows 11

### Windows CPU architecture

x86_64 (64-bit)

```shell
Microsoft Windows [Version 10.0.26200.9168]
```

```shell
C:\Temp>type "C:\Program Files\Git\etc\install-options.txt"
Editor Option: VIM
Custom Editor Path:
Default Branch Option:
Path Option: Cmd
SSH Option: OpenSSH
Tortoise Option: false
CURL Option: OpenSSL
CRLF Option: CRLFCommitAsIs
Bash Terminal Option: MinTTY
Git Pull Behavior Option: Merge
Enable Symlinks: Disabled
Enable FSMonitor: Disabled
```

### Other interesting things

The [Git for Windows v2.55.0.windows.5](https://github.com/git-for-windows/git/releases/tag/v2.55.0.windows.5) release said:

> Git for Windows' installer [is now a 64-bit executable](https://github.com/git-for-windows/build-extra/pull/726) (x64 instead of 86).

Did this fix actually take effect? I observe in Task Manager, and directly via `dumpbin`, that the installer is still 32-bit. (Below, `hello64.exe` and `hello32.exe` are Hello World executables I built to verify that `dumpbin /headers` can detect the difference.)

```
C:\Temp>dumpbin /headers hello64.exe | rg machine
            8664 machine (x64)

C:\Temp>dumpbin /headers hello32.exe | rg machine
             14C machine (x86)
                   32 bit word machine

C:\Temp>dumpbin /headers Git-2.55.0.5-64-bit.exe | rg machine
             14C machine (x86)
                   32 bit word machine

C:\Temp>coreutils sha256sum Git-2.55.0.5-64-bit.exe
d065a4e23c3d9a6b5073d609b5be0830227ec3ca053c083ba385061ddfaf94c6 *Git-2.55.0.5-64-bit.exe
```

### Terminal/shell

Windows Terminal, Command Prompt

```shell
Run the Git installer and look at it in Task Manager. Or inspect it via `dumpbin /headers`.
```

### Expected behaviour

I expected the installer to be 64-bit, per the Changelog.

### Actual behaviour

The installer is 32-bit. Screenshot:

Image

### Repository

_No response_

dennisameling:
This fixes https://github.com/git-for-windows/git/issues/6372.

Confirming that the installer is now truly a x64 executable:

image

Inno Setup 7 introduced 64-bit installers, but building 32-bit ones stays the default in both editions; a 64-bit one is only produced when the `[Setup]` section directive `SetupArchitecture` is set to `x64`. Updating the compiler to Inno Setup 7.0.2 therefore did not change the architecture of the installer, even though the v2.55.0(5) release notes already announced it as a 64-bit executable.

Set `SetupArchitecture=x64` for the flavors that ship 64-bit Git, i.e. the ones where `release.sh` sets `BITNESS` to 64: MINGW64, UCRT64 and CLANGARM64. The i686 installer keeps building as a 32-bit executable. The directive does not exist in Inno Setup 6, hence the version guard.

Flipping that directive alone leaves the installer broken, though: Pascal Script records are always packed, and in a 64-bit installer pointers are eight bytes wide, so the Windows API records declared in `exec-with-capture.inc.iss` no longer match the layout those functions expect. `CreatePipe()` reads `lpSecurityDescriptor` from the wrong offset, fails with `ERROR_NOACCESS`, and `ExecWithCapture()` returns `False` without ever having run the command. Starting the wizard then greets the user with

	Unable to get system config (exit code 4294967295):

Widen the pointer-sized fields of `SECURITY_ATTRIBUTES` and `STARTUPINFO` to `ULONG_PTR` and add the padding that the aligned C structs have and a packed record does not, which brings them to the 24 and 104 bytes that the x64 ABI mandates. `MSG` needs the same treatment for a different reason: `PeekMessage()` writes the full 48 bytes of that struct, which is more than an unpadded record offers, corrupting the stack and killing the message pump with a fatal exception once a captured command runs long enough for the pump to kick in.

The padding is conditional on the new `SETUP_IS_X64` symbol, so 32-bit installers keep the record layout they have always had.

The `Win95` and `Win2000` code paths in `modules.inc.iss` declare records with the same problem (`PROCESSENTRY32`, `MODULEENTRY32`), but `MinVersion` is 6.3 and those paths are only taken on Windows versions before Vista, so they are left alone.

### DIFF
--- a/installer/exec-with-capture.inc.iss
+++ b/installer/exec-with-capture.inc.iss
@@ -6,8 +6,14 @@ type
 #endif
     SECURITY_ATTRIBUTES = record
         nLength:DWORD;
-        lpSecurityDescriptor:LongInt;
+#ifdef SETUP_IS_X64
+        nLengthPadding:DWORD;
+#endif
+        lpSecurityDescriptor:ULONG_PTR;
         bInheritHandle:BOOL;
+#ifdef SETUP_IS_X64
+        bInheritHandlePadding:DWORD;
+#endif
     end;
 
 function CreatePipe(var hReadPipe,hWritePipe:HANDLE;var lpPipeAttributes:SECURITY_ATTRIBUTES;nSize:DWORD):BOOL;
@@ -26,10 +32,13 @@ type
         dwProcessId:DWORD;
         dwThreadId:DWORD;
     end;
-    LPSTR = LongInt;
-    LPBYTE = LongInt;
+    LPSTR = ULONG_PTR;
+    LPBYTE = ULONG_PTR;
     STARTUPINFO = record
         cb:DWORD;
+#ifdef SETUP_IS_X64
+        cbPadding:DWORD;
+#endif
         lpReserved:LPSTR;
         lpDesktop:LPSTR;
         lpTitle:LPSTR;
@@ -43,6 +52,9 @@ type
         dwFlags:DWORD;
         wShowWindow:WORD;
         cbReserved2:WORD;
+#ifdef SETUP_IS_X64
+        cbReserved2Padding:DWORD;
+#endif
         lpReserved2:LPBYTE;
         hStdInput:HANDLE;
         hStdOutput:HANDLE;
@@ -67,10 +79,16 @@ type
     TMsg = record
         hwnd:HWND;
         message:UINT;
-        wParam:LongInt;
-        lParam:LongInt;
+#ifdef SETUP_IS_X64
+        messagePadding:DWORD;
+#endif
+        wParam:ULONG_PTR;
+        lParam:ULONG_PTR;
         time:DWORD;
         pt:TPoint;
+#ifdef SETUP_IS_X64
+        lPrivate:DWORD;
+#endif
     end;
 
 const
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -53,6 +53,12 @@ SolidCompression=yes
 #define SOURCE_DIR SourcePath+'\..\..\..\..'
 #endif
 SourceDir={#SOURCE_DIR}
+#if BITNESS=='64' && Ver>=EncodeVer(7, 0, 0)
+; Inno Setup 7 builds 32-bit installers unless told otherwise. SETUP_IS_X64
+; lets the Pascal Script code declare Windows API records for 64-bit.
+#define SETUP_IS_X64
+SetupArchitecture=x64
+#endif
 #if BITNESS=='64' || INSTALLER_FILENAME_SUFFIX=='arm64'
 ArchitecturesInstallIn64BitMode=x64 arm64
 #endif


installer/install.iss
```diff
@@ -53,6 +53,12 @@ SolidCompression=yes
 #define SOURCE_DIR SourcePath+'\..\..\..\..'
 #endif
 SourceDir={#SOURCE_DIR}
+#if BITNESS=='64' && Ver>=EncodeVer(7, 0, 0)
+; Inno Setup 7 builds 32-bit installers unless told otherwise. SETUP_IS_X64
+; lets the Pascal Script code declare Windows API records for 64-bit.
+#define SETUP_IS_X64
+SetupArchitecture=x64
+#endif
```
dennisameling, line 61: As a follow-up, we'll probably want to remove all 32-bit / x86-related code from the installer, [given that it's no longer supported](https://gitforwindows.org/32-bit.html) (except for MinGit which doesn't use this installer).

installer/exec-with-capture.inc.iss
```diff
@@ -67,9 +79,15 @@ type
     TMsg = record
         hwnd:HWND;
         message:UINT;
-        wParam:LongInt;
-        lParam:LongInt;
+#ifdef SETUP_IS_X64
+        messagePadding:DWORD;
```
dscho, line 83: Ah, I now get it! This was a bit puzzling for me: The `wParam` and `pt` attributes are 8-byte aligned, even if their actual data types are 32-bit and would therefore only require 4 bytes.

Looks good!

dscho (approved):

installer/exec-with-capture.inc.iss
```diff
@@ -67,9 +79,15 @@ type
...
+#endif
+        wParam:ULONG_PTR;
+        lParam:ULONG_PTR;
         time:DWORD;
+#ifdef SETUP_IS_X64
+        timePadding:DWORD;
+#endif
         pt:TPoint;
```
dscho: Ah, there is a tiny problem here (will fix in a moment): `TPoint` does _not_ need to be 8-byte aligned, apparently:

```c
#include <stdio.h>
#include <windows.h>

int main(int argc, char **argv)
{
	MSG msg;
#define P(name) printf("offset of " #name ": 0x%02x\n", (int)((char *)&msg. name - (char *)&msg))
        P(hwnd);
        P(message);
        P(wParam);
        P(lParam);
        P(time);
        P(pt);
        printf("offset of lPrivate: 0x%02x\n", (int)((char *)&msg.pt - (char *)&msg) + sizeof(msg.pt));
	return 0;
}
```

prints

```
offset of hwnd: 0x00
offset of message: 0x08
offset of wParam: 0x10
offset of lParam: 0x18
offset of time: 0x20
offset of pt: 0x24
offset of lPrivate: 0x2c
```

And this test InnoSetup program with the adjusted definition agrees:

```pascal
[Setup]
AppName=MSG Layout Probe
AppVersion=1
DefaultDirName={tmp}\MsgProbe
OutputBaseFilename=msg-probe
SetupArchitecture=x64
Uninstallable=no

[Code]
type
  TByteArray = array of Byte;

  TMsg = record
    hwnd:HWND;
    message:UINT;
    messagePadding:DWORD;
    wParam:ULONG_PTR;
    lParam:ULONG_PTR;
    time:DWORD;
    pt:TPoint;
    lPrivate:DWORD;
  end;

procedure RtlMoveMemory(var Destination: TMsg; Source: TByteArray; Length: SIZE_T);
  external 'RtlMoveMemory@kernel32.dll stdcall';

function InitializeSetup: Boolean;
var
  Bytes: TByteArray;
  Info: TMsg;
  I: Integer;
begin
  SetArrayLength(Bytes, 48);
  for I := 0 to GetArrayLength(Bytes) - 1 do
    Bytes[I] := 0;

  Bytes[$00] := 1;
  Bytes[$08] := 2;
  Bytes[$10] := 3;
  Bytes[$18] := 4;
  Bytes[$20] := 5;
  Bytes[$24] := 6;
  Bytes[$28] := 7;
  Bytes[$2c] := 8;

  RtlMoveMemory(Info, Bytes, SizeOf(Info));

  Log('SizeOf=' + IntToStr(SizeOf(Info)));
  Log('hwnd=' + UIntToStr(Info.hwnd));
  Log('message=' + UIntToStr(Info.message));
  Log('wParam=' + UIntToStr(Info.wParam));
  Log('lParam=' + UIntToStr(Info.lParam));
  Log('time=' + UIntToStr(Info.time));
  Log('pt.x=' + UIntToStr(Info.pt.x));
  Log('pt.y=' + UIntToStr(Info.pt.y));
  Log('lPrivate=' + UIntToStr(Info.lPrivate));

  Result := False;
end;
```

Output:

```
2026-08-31 13:04:05.040   SizeOf=48
2026-08-31 13:04:05.040   hwnd=1
2026-08-31 13:04:05.040   message=2
2026-08-31 13:04:05.040   wParam=3
2026-08-31 13:04:05.040   lParam=4
2026-08-31 13:04:05.040   time=5
2026-08-31 13:04:05.040   pt.x=6
2026-08-31 13:04:05.040   pt.y=7
2026-08-31 13:04:05.040   lPrivate=8
```

dscho:
For the record, I inspected the other `record` changes (and had GPT look at the remaining DLL calls) and could not find any other required change, so now it's merge time!

PR #733, opened 2026-08-25
installer: drop 32-bit support

dennisameling:
Git for Windows v2.48.1 was the last version to ship a 32-bit installer; since then only 32-bit MinGit is built, and that only until April 2029. See https://gitforwindows.org/32-bit.html for the background. The i686 code paths in the installer have been dead weight ever since, and the end-of-support warning shown by the 32-bit installer can never be reached by anything that is still being built.

Remove the MINGW32 arm of the architecture switch in `release.sh`, replacing it with a `die` that names the reason, so that anyone who still tries gets an explanation rather than "Unhandled MSYSTEM: MINGW32".

With MINGW32 gone, `BITNESS` can only ever be 64, so drop the variable altogether rather than keep a switch with a single position. That means `OutputBaseFilename` spells out `-64-bit` instead of interpolating it, which produces the exact same name the released asset has always had, as verified by compiling `install.iss` against a `config.iss` without `INSTALLER_FILENAME_SUFFIX`: `Git-2.55.0.5-64-bit.exe`.

The `pack.packsizelimit` entry needs to go from both the outer and the inner `case` in the system config handling: the inner arm `continue`d for anything that was not 32-bit, so leaving the outer pattern in place would have started writing the key into the 64-bit installers' system config for the first time. Generating `config.iss` before and after this commit confirms the output is unchanged apart from the dropped `BITNESS` define.

CI built the installer for i686 as part of the `sdk-artifacts` job, which would now fail on that `die`, so exclude that combination the same way `minimal` and `makepkg-git` already are. The `check-for-missing-dlls` job keeps building the i686 `build-installers` SDK artifact, so that coverage is retained. The two invocations of the installer lose `/ALLOWINSTALLING32ON64=1`, whose only consumer was the removed warning.

Note that `make-file-list.sh` and `mingit/` keep their i686 support, since 32-bit MinGit is still being built.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -176,7 +176,7 @@ jobs:
         shell: pwsh
         run: |
           $exePath = Get-ChildItem -Path ${{ steps.build.outputs.result }}/*.exe | %{$_.FullName}
-          $installer = Start-Process -PassThru -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART /SUPPRESSMSGBOXES /ALLOWDOWNGRADE=1 /ALLOWINSTALLING32ON64=1 /LOG=installer.log"
+          $installer = Start-Process -PassThru -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART /SUPPRESSMSGBOXES /ALLOWDOWNGRADE=1 /LOG=installer.log"
           $exitCode = $installer.ExitCode
           if ($exitCode -ne 0) {
             Write-Host "::error::Installer failed with exit code $exitCode!"
@@ -314,7 +314,7 @@ jobs:
         shell: pwsh
         run: |
           $exePath = Get-ChildItem -Path installer-${{ matrix.arch.name }}/*.exe | %{$_.FullName}
-          $installer = Start-Process -PassThru -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART /SUPPRESSMSGBOXES /ALLOWDOWNGRADE=1 /ALLOWINSTALLING32ON64=1 /LOG=installer.log"
+          $installer = Start-Process -PassThru -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART /SUPPRESSMSGBOXES /ALLOWDOWNGRADE=1 /LOG=installer.log"
           $exitCode = $installer.ExitCode
           if ($exitCode -ne 0) {
             Write-Host "::error::Installer failed with exit code $exitCode!"
--- a/installer/HowToRelease.txt
+++ b/installer/HowToRelease.txt
@@ -7,7 +7,9 @@ A setup executable is named as follows
  - followed by the Git for Windows version of the form A.B.C.D where
    A.B.C is the Git version on which the release is based, and
    A.B.C.windows.D is the tag in Git for Windows' repository
- - followed by the "bitness", i.e. "-32-bit" or "-64-bit"
+ - followed by an architecture suffix, i.e. "-arm64" or "-64-bit"
+   (we used to distinguish "x86" and "x86_64" as "32-bit" and "64-bit"
+   and the suffix stuck)
 
 Examples:
 
@@ -52,7 +54,6 @@ Here is a step-by-step instruction for the maintainer:
 
    # ensure current setup is up-to-date (including the Git packages)
    pacman -Syu
-   # on 32-bit, quit all MSys shells and run \autorebase.bat now
 
    # make installer (after updating the release notes)
    /usr/src/build-extra/installer/release.sh <git-for-windows-version>
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -4,8 +4,8 @@
 
 #include "config.iss"
 
-#if !defined(APP_VERSION) || !defined(BITNESS) || !defined(MINGW_BITNESS)
-#error "config.iss should define APP_VERSION, BITNESS and MINGW_BITNESS"
+#if !defined(APP_VERSION) || !defined(MINGW_BITNESS)
+#error "config.iss should define APP_VERSION and MINGW_BITNESS"
 #endif
 
 #define APP_NAME      'Git'
@@ -25,10 +25,6 @@
 #define DEFAULT_BRANCH_NAME 'master'
 #endif
 
-#ifndef INSTALLER_FILENAME_SUFFIX
-#define INSTALLER_FILENAME_SUFFIX ''
-#endif
-
 [Setup]
 ; Compiler-related
 Compression=lzma2/ultra64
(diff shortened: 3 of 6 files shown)


installer/HowToRelease.txt
```diff
@@ -7,7 +7,7 @@ A setup executable is named as follows
  - followed by the Git for Windows version of the form A.B.C.D where
    A.B.C is the Git version on which the release is based, and
    A.B.C.windows.D is the tag in Git for Windows' repository
- - followed by the "bitness", i.e. "-32-bit" or "-64-bit"
+ - followed by the "bitness", i.e. "-64-bit"
```
rimrul:
```suggestion
 - folllwed by an architecture suffix, i.e. "-arm64" or "-64-bit" (we used to distinguish "x86" and "x86_64" as "32-bit" and "64-bit" and the suffix stuck)
```

rimrul (approved):
The bitness bullet point turns a little weird, but the restlooks good.

installer/HowToRelease.txt
dennisameling: Agreed, thanks! Updated.

installer/HowToRelease.txt
```diff
@@ -7,7 +7,9 @@ A setup executable is named as follows
  - followed by the Git for Windows version of the form A.B.C.D where
    A.B.C is the Git version on which the release is based, and
    A.B.C.windows.D is the tag in Git for Windows' repository
- - followed by the "bitness", i.e. "-32-bit" or "-64-bit"
+ - folllwed by an architecture suffix, i.e. "-arm64" or "-64-bit"
```
StephanTLavavej: This is introducing a "folllwed" typo.

dennisameling: Good catch, thanks! Fixed.

dscho:
Before we merge this PR, we need to adjust [this line](https://github.com/git-for-windows/git-sdk-32/blob/e622347e6743782459e3f9fd27f438d46ffc86f5/.github/workflows/git-artifacts.yml#L145) in `git-sdk-32`'s `git-artifacts` workflow (it still diligently verifies that we could build installers and Portable Gits if we wanted to).

.github/workflows/main.yml
```diff
@@ -263,6 +263,12 @@ jobs:
               mingw-prefix: mingw32
               msystem: MINGW32
               sdk-repo: git-sdk-32
+          - artifact: build-installers
+            arch:
+              name: i686
+              mingw-prefix: mingw32
+              msystem: MINGW32
+              sdk-repo: git-sdk-32
```
dscho: I fear that we still need to exercise the `build-installers` flavor, even for i686, since that's what is used by [the automation](https://github.com/git-for-windows/git-for-windows-automation/blob/1603eef1c0fca27e30ce5719e96f7936cf1a5c82/.github/workflows/git-artifacts.yml#L213-L216) to build full releases' MinGits.

installer/install.iss
```diff
@@ -40,7 +40,7 @@ OutputDir={#GetEnv('TEMP')}
 #if INSTALLER_FILENAME_SUFFIX!=''
 OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION+'-'+INSTALLER_FILENAME_SUFFIX}
 #else
-OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION}-{#BITNESS}-bit
+OutputBaseFilename={#APP_NAME+'-'+FILENAME_VERSION}-64-bit
```
dscho: It would be pretty neat if we always defined `INSTALLER_FILENAME_SUFFIX`, and never hard-coded 64-bit except in one central place?

dennisameling:
> Before we merge this PR, we need to adjust [this line](https://github.com/git-for-windows/git-sdk-32/blob/e622347e6743782459e3f9fd27f438d46ffc86f5/.github/workflows/git-artifacts.yml#L145) in `git-sdk-32`'s `git-artifacts` workflow (it still diligently verifies that we could build installers and Portable Gits if we wanted to).

Great catch, thank you. Here's a PR to do exactly that: https://github.com/git-for-windows/git-sdk-32/pull/64

dscho (commented):
This PR conflicts a bit with https://github.com/git-for-windows/build-extra/pull/732, right? Could we take care of that first, seeing as it caused user-visible bugs (cf. external OpenSSH issue, https://github.com/git-for-windows/git/issues/6374)?

.github/workflows/main.yml
dennisameling: Ah sorry, that was a mistake. Will update.

installer/install.iss
dennisameling: Agreed, will update!

installer/install.iss
```diff
@@ -53,15 +45,11 @@ SolidCompression=yes
...
 #endif
 SourceDir={#SOURCE_DIR}
-#if BITNESS=='64' && Ver>=EncodeVer(7, 0, 0)
-; Inno Setup 7 builds 32-bit installers unless told otherwise. SETUP_IS_X64
-; lets the Pascal Script code declare Windows API records for 64-bit.
-#define SETUP_IS_X64
+#if Ver>=EncodeVer(7, 0, 0)
+; Inno Setup 7 builds 32-bit installers unless told otherwise.
 SetupArchitecture=x64
 #endif
```
dennisameling: We could go a step further and remove this version check altogether, given that this repository already runs on Inno Setup 7. WDYT?

dscho: @dennisameling yep, removing the logic makes sense to me!

dennisameling: Done! ✅

installer/install.iss
```diff
@@ -53,15 +45,9 @@ SolidCompression=yes
...
 SourceDir={#SOURCE_DIR}
-#if BITNESS=='64' && Ver>=EncodeVer(7, 0, 0)
-; Inno Setup 7 builds 32-bit installers unless told otherwise. SETUP_IS_X64
-; lets the Pascal Script code declare Windows API records for 64-bit.
-#define SETUP_IS_X64
+; Inno Setup 7 builds 32-bit installers unless told otherwise.
 SetupArchitecture=x64
-#endif
-#if BITNESS=='64' || INSTALLER_FILENAME_SUFFIX=='arm64'
 ArchitecturesInstallIn64BitMode=x64 arm64
```
StephanTLavavej, line 50: Pre-existing, but FYI:

https://jrsoftware.org/ishelp/index.php?topic=archidentifiers

> Before Inno Setup 6.3, `x64os` was named `x64`. The compiler still accepts `x64` as an alias for `x64os`, but will emit a deprecation warning when used.

dennisameling: Good catch. I just learned that this [is no longer needed](https://jrsoftware.org/ishelp/index.php?topic=archidentifiers) if `SetupArchitecture` is set to x64:

> Building 32-bit installers is the default in both editions. Set [SetupArchitecture](https://jrsoftware.org/ishelp/topic_setup_setuparchitecture.htm) to x64 to build a 64-bit x64 installer instead of a 32-bit x86 installer. This also changes the default values of [ArchitecturesAllowed](https://jrsoftware.org/ishelp/topic_setup_architecturesallowed.htm) and [ArchitecturesInstallIn64BitMode](https://jrsoftware.org/ishelp/topic_setup_architecturesinstallin64bitmode.htm) to x64compatible, making 64-bit installers default to [64-bit install mode](https://jrsoftware.org/ishelp/topic_32vs64bitinstalls.htm).

This line was originally added to ensure that GfW x64 and arm64 were installed in `C:\Program Files` instead of `C:\Program Files (x86)`, but that's no longer needed now that `SetupArchitecture` has been set to `x64`. I'll add an additional commit to clean it up.

dscho (approved):
Excellent, thank you!

installer/install.iss
dscho: Ooops... I guess I merged this prematurely. Sorry @dennisameling!

FWIW I think that `x64` should be replaced with `x64compatible` instead of `x64os`, if at all ;-)

dennisameling: No worries! New PR incoming :)

dennisameling: https://github.com/git-for-windows/build-extra/pull/735

PR #735, opened 2026-09-07
installer: remove superfluous `ArchitecturesInstallIn64BitMode`

dennisameling:
SetupArchitecture=x64 is now defined unconditionally with the removal of 32-bit support in the installer. This setting also sets `ArchitecturesInstallIn64BitMode=x64compatible` by default, which matches on both arm64 and x64 systems. This line no longer needs to be defined explicitly.

Ref: https://jrsoftware.org/ishelp/index.php?topic=64bit


dscho (approved):
Thank you!

PR #736, opened 2026-09-07
installer: address Inno Setup 7 compiler warnings

dennisameling:
Since the update to Inno Setup 7.1.0 (#727), the installer compiles with 28 warnings, e.g. in [this run](https://github.com/git-for-windows/build-extra/actions/runs/34155632457/job/101846779340). This addresses all of them; the compiler output is now warning-free.

* **`LZMAUseSeparateProcess`** is obsolete and ignored — dropped.
* **Renamed functions and constants** (`GetFileVersion`, `{pf}`, `IsAdminLoggedOn`, `IsComponentSelected`, `FileCopy`) — switched to `GetVersionNumbersString`, `{commonpf}`, `IsAdmin`, `WizardIsComponentSelected` and `CopyFile`. Each old name is a plain alias for the new one (`GetFileVersion` is literally defined as a deprecation wrapper in `ISPPBuiltins.iss`), so this is a pure rename.

  Note that `{pf}` is an alias for `{commonpf}`, *not* for `{autopf}`: an unprivileged installation keeps defaulting to `{commonpf}\Git` and still falls back to `{userpf}\Git` in `CurPageChanged()` when that turns out not to be writable. Likewise `IsAdmin`, not `IsAdminInstallMode`.
* **Unused local variables** — removed. All are leftovers from earlier refactorings, except `CurrentVersion`/`Msg` in `InitializeSetup()`, which the downgrade check uses but which is compiled in only for real versions; their declaration is guarded by the same condition.
* **`release.sh`** synthesizes `SetSystemConfigDefaults()` into `config.iss` and still used the old `IsComponentSelected` there, so that one warning survived in real builds even after the hand-written call sites were converted.

The `Architecture identifier "x64" is deprecated` warning from that run is not addressed here: it came from `ArchitecturesInstallIn64BitMode=x64 arm64`, which #735 already removed.

### Verification

`./installer/release.sh 0-test` produces a complete installer with a successful compile and zero warnings. `ISCC.exe install.iss` was additionally run against a release-like config (real `APP_VERSION`/`GIT_VERSION`, no `DO_NOT_INSTALL`, so the guarded declaration is exercised) and against an x64-style config without `ARCHS_ALLOWED` — both clean. Each commit compiles on its own (27 → 22 → 0 warnings).

### DIFF
--- a/installer/helpers.inc.iss
+++ b/installer/helpers.inc.iss
@@ -84,7 +84,7 @@ end;
 // Returns the path to the common or user shell folder as specified in "Param".
 function GetShellFolder(Param:string):string;
 begin
-    if IsAdminLoggedOn then begin
+    if IsAdmin then begin
         Param:='{common'+Param+'}';
     end else begin
         Param:='{user'+Param+'}';
@@ -92,7 +92,7 @@ begin
     Result:=ExpandConstant(Param);
 end;
 
-// As IsComponentSelected() is not supported during uninstall, this work-around
+// As WizardIsComponentSelected() is not supported during uninstall, this work-around
 // simply checks the Registry. This is unreliable if the user runs the installer
 // twice, the first time selecting the component, the second deselecting it.
 function IsComponentInstalled(Component:String):Boolean;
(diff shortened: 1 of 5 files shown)


dscho (approved):
Excellent "spring" cleaning, thank you so much!

PR #738, opened 2026-09-08
Add notice about external OpenSSH to release notes

### LINKED ISSUE #6374: v2.55.0.windows.5 installer always bundles OpenSSH, also overwrites previously saved choice of ExternalOpenSSH

- [x] I was not able to find an open or closed issue matching what I'm seeing

### Git for Windows version

```shell
git version 2.55.0.windows.5
cpu: x86_64
built from commit: 32c4f7689275d233577576630e1ac5b7eb354eb0
sizeof-long: 4
sizeof-size_t: 8
shell-path: D:/git-sdk-64-build-installers/usr/bin/sh
rust: disabled
feature: fsmonitor--daemon
gettext: enabled
libcurl: 8.21.0
OpenSSL: OpenSSL 3.5.7 9 Jun 2026
zlib: 1.3.2
SHA-1: SHA1_DC
SHA-256: SHA256_BLK
default-ref-format: files
default-hash: sha1
```

### Windows version

Both Windows 11 and Windows 10

### Windows CPU architecture

x86_64 (64-bit)

- Windows 11: `Microsoft Windows [Version 10.0.26200.9168]`
- Windows 10: `Microsoft Windows [Version 10.0.19045.7663]`

```shell
Editor Option: VIM
Custom Editor Path:
Default Branch Option: main
Path Option: Cmd
SSH Option: OpenSSH
Tortoise Option: false
CURL Option: OpenSSL
CRLF Option: CRLFCommitAsIs
Bash Terminal Option: MinTTY
Git Pull Behavior Option: Merge
Enable Symlinks: Disabled
Enable FSMonitor: Disabled
```

### Other interesting things

I have a third-party SSH agent running that listens on the named OpenSSH pipe, `\\.\pipe\openssh-ssh-agent` (specifically the 1Password SSH Agent). After updating to 2.55.0.windows.5, `ssh` could suddenly no longer see my agent running.

After some searching (see [my comment on #6373](https://github.com/git-for-windows/git/issues/6373#issuecomment-5381496563)), I discovered that `C:\Program Files\Git\usr\bin` suddenly included a bunch of extra SSH stuff which was not there before, despite me having selected "Use external SSH" during my original installation (though this does not appear in the `install-options.txt` above; you're just going to have to trust me when I say that I _did_ select "Use external SSH"; see investigation below):

```
[Matthew@MATT-FW13 ~] $ ls /usr/bin | grep ssh
ssh.exe*
ssh-add.exe*
ssh-agent.exe*
ssh-copy-id*
ssh-keygen.exe*
ssh-keyscan.exe*
ssh-pageant.exe*
```

I thought I might try to re-install Git Bash, but when I did so, **I was not even given the option to use an external SSH.**

#### My investigation

Following [this discussion comment](https://github.com/git-for-windows/git/discussions/3451#discussioncomment-1424427), I _believe_ that the issue comes down to [this section of the installer script](https://github.com/git-for-windows/build-extra/blob/7b61c1c82052183d0d865fa807d69636c5245dbc/installer/install.iss#L2244-L2339):

```pascal
(*
 * Create a custom page for using (Tortoise)Plink or a self-supplied OpenSSH instead of bundled OpenSSH
 * if at least one PuTTY session is found in the Registry or an OpenSSH binary is found on the PATH.
 *)

if (RegGetSubkeyNames(HKEY_CURRENT_USER,'Software\SimonTatham\PuTTY\Sessions',PuTTYSessions) and (GetArrayLength(PuTTYSessions)>0)) or
   (FileSearch('ssh.exe', GetEnv('PATH')) <> '') then begin
    SSHChoicePage:=CreatePage(PrevPageID,'Choosing the SSH executable','Which Secure Shell client program would you like Git to use?',TabOrder,Top,Left);

// ...
```

I don't have PuTTY installed, and I never have; but I do have `ssh.exe` on my PATH (specifically, `%SYSTEMROOT%\System32\OpenSSH` is on my **system** PATH, not my **user** PATH):

```
PS C:\Users\Matthew> where.exe ssh.exe
C:\Windows\System32\OpenSSH\ssh.exe
```

So, specifically, **I believe the problem comes down to `FileSearch('ssh.exe', GetEnv('PATH'))`. It appears that, in the new installer, this function fails to locate `ssh.exe` on the system path for some reason.** Maybe this has something to do with the move to a 64-bit installer? Anyways, the `else` branch of that check sets `SSHChoicePage` to `NIL`, which means that:

1.  [This block](https://github.com/git-for-windows/build-extra/blob/7b61c1c82052183d0d865fa807d69636c5245dbc/installer/install.iss#L3565-L3571), to delete the files from OpenSSH, never runs:
    
    ```pascal
    #ifdef DELETE_OPENSSH_FILES
        if (SSHChoicePage<>NIL) and (RdbSSH[GS_ExternalOpenSSH].Checked) then begin
            WizardForm.StatusLabel.Caption:='Removing bundled Git OpenSSH binaries';
            if not DeleteOpenSSHFiles() then
                LogError('Failed to remove OpenSSH file(s)');
        end;
    #endif
    ```

2.  [This block](https://github.com/git-for-windows/build-extra/blob/7b61c1c82052183d0d865fa807d69636c5245dbc/installer/install.iss#L3707-L3721), later in the script, overwrites the previously selected `ExternalOpenSSH` with the default `OpenSSH`:

    ```pascal
    // Git SSH options.
    Data:='';
    Data2:='false';
    if (SSHChoicePage=NIL) or RdbSSH[GS_OpenSSH].Checked then begin
        Data:='OpenSSH';
    end else if RdbSSH[GS_ExternalOpenSSH].Checked then begin
        Data:='ExternalOpenSSH'
    end else if RdbSSH[GS_Plink].Checked then begin
        Data:='Plink';
        RecordChoice(PreviousDataKey,'Plink Path',EdtPlink.Text);
        if TortoisePlink.Checked then
            Data2:='true';
    end;
    RecordChoice(PreviousDataKey,'SSH Option',Data);
    RecordChoice(PreviousDataKey,'Tortoise Option',Data2);
    ```

Unless I am mistaken, the second point means that, even if this issue is resolved, running that fixed installer with "only show new options" ticked will not be able to detect that the user had previously selected `ExternalOpenSSH`.

### Terminal/shell

Git Bash in Windows Terminal.

1. Verify that `C:\Windows\System32\OpenSSH` is on the system-level PATH (not the user-level PATH; though I'm not sure if it makes a difference), and that `ssh.exe` lives there (`where.exe ssh`). (NB: I am not sure if this issue occurs with other external SSH installations other than the one from System32).
2. Verify that `HKEY_CURRENT_USER\SOFTWARE\SimonTatham\PuTTY` does not appear in the Windows Registry.
3. Install or downgrade to `v2.55.0.windows.4` and select **"Use external SSH":**
   1. Run the `Git-2.55.0.4-64-bit.exe` installer from [the previous release](https://github.com/git-for-windows/git/releases/tag/v2.55.0.windows.4).
   2. Untick "Only show new options."
   3. Click through the installer and, when prompted, select **"Use external SSH".**
3. Verify that `ssh` points to `C:\Windows\System32\OpenSSH\ssh.exe`:
   1. Open Git Bash.
   2. Verify that `where.exe ssh` displays only the one instance of `ssh.exe`.
   3. Verify that `ls /usr/bin | grep ssh` does not include any of the `ssh*.exe` files from OpenSSH (should only display `ssh-pageant.exe`).
4. Upgrade to `v2.55.0.windows.5`:
   1. Run the `Git-2.55.0.5-64-bit.exe` installer from [the latest release](https://github.com/git-for-windows/git/releases/tag/v2.55.0.windows.5).
   2. Click through the installation wizard either **with** or **without** "Only show new options."
5. Check whether or not `ssh` points to `C:\Windows\System32\OpenSSH\ssh.exe` (repeat step 3).

### Expected behaviour

After upgrading to `v2.55.0.windows.5`:
 
1. Running `where.exe ssh` should display **only** the external SSH program in `C:\Windows\System32\OpenSSH\ssh.exe`. Other OpenSSH files (`ssh-add.exe`, `ssh-copy-id.exe`, etc.) should **not** appear in `C:\Program Files\Git\usr\bin`.

2. Additionally, if run _without_ "Only show new options," then the **"Choosing the SSH executable"** page should appear during installation (between "Adjusting your PATH environment" and "Choosing HTTPS transport backend"). The page should have the **"Use external OpenSSH"** option available, and it should be selected by default.

3. Furthermore, re-running `Git-2.55.0.4-64-bit.exe` _without_ "Only show new options" should have "Use external SSH" selected by default when the "Choosing the SSH executable" page appears; re-running it _with_ "Only show new options" should install `v2.55.0.windows.4` **without** the bundled SSH included in `C:\Program Files\Git\usr\bin`.

### Actual behaviour

After upgrading to `v2.55.0.windows.5`:

1. Running `where.exe` now displays `C:\Program Files\Git\usr\bin\ssh.exe` _above_ `C:\Windows\System32\OpenSSH\ssh.exe`. Other OpenSSH files appear in `C:\Program Files\Git\usr\bin`.

2. Additionally, if the installer was run _without_ "Only show new options," then the **"Choosing the SSH executable"** page never appears.

3. Finally, when re-running `Git-2.55.0.4-64-bit.exe` _without_ "Only show new options," the "Choosing the SSH executable" page now defaults to "Use bundled OpenSSH"; re-running it _with_ "Only show new options" installs `v2.55.0.windows.4` **with** the bundled SSH included in `C:\Program Files\Git\usr\bin`.

### Repository

_No response_

matthew-e-brown:
Adds a warning to the "Changes since v2.55.0(5)" section about external OpenSSH installations and provides some instructions for affected users.

This closes https://github.com/git-for-windows/git/issues/6374.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -40,6 +40,8 @@ Git for Windows is distributed with other components yet, such as Bash, zlib, cu
 
 Following the [MSYS2 project](https://www.msys2.org/news/#2026-02-28-dropping-support-for-windows-81), on which Git for Windows is based, Windows 8.1 support will be dropped after Git for Windows v2.55.
 
+An issue with the installer for the previous version (v2.55.0.windows.5) caused the "Use external OpenSSH" option to be disabled for some users. This caused the bundled version of OpenSSH to be installed and overwrote any previously-saved choice of external OpenSSH. **If you rely on an external OpenSSH installation, and you updated to v2.55.0(5),** you should consider re-running the latest installer with "Only show new options" unchecked so that you can re-enable the external OpenSSH option. The bundled version of OpenSSH will be uninstalled automatically. If you do not rely on an external OpenSSH installation, or you did not install v2.55.0(5) specifically, you can safely ignore this notice.
+
 ### New Features
 
 * Comes with [Git LFS v3.8.0](https://github.com/git-lfs/git-lfs/releases/tag/v3.8.0).
@@ -47,7 +49,7 @@ Following the [MSYS2 project](https://www.msys2.org/news/#2026-02-28-dropping-su
 
 ### Bug Fixes
 
-* The installer [is now _actually_ a 64-bit one](https://github.com/git-for-windows/build-extra/pull/732), which fixes the problem that [the external OpenSSH option was broken in Git for Windows v2.55.0(5)](https://github.com/git-for-windows/git/issues/6374).
+* The installer [is now _actually_ a 64-bit one](https://github.com/git-for-windows/build-extra/pull/732), which fixes the problem that [the external OpenSSH option was broken in Git for Windows v2.55.0(5)](https://github.com/git-for-windows/git/issues/6374) (see notice above).
 * It is now [finally possible](https://github.com/git-for-windows/git/pull/6353) to commit 4GB objects or larger in Git for Windows.
 
 ## Changes since Git for Windows v2.55.0(4) (August 11th 2026)


dscho (approved):
It's slightly on the verbose side, but I think it's still fine.

matthew-e-brown:
I was a little worried about that; but I figured it best to cover all the bases so it never needs to come up again instead of risking leaving something out. That said, making things more verbose than they need to be is kind of my specialty... I'm not particularly attached to the wording, if you can think of some ways to make it shorter. :)

Otherwise, is there nothing else you need from me to merge this & close that other issue?

dscho:
Hahaha, I forgot to merge! ;-) Darn distractions...

PR #742, opened 2026-09-10
git-extra: avoid creating a new Desktop shortcut for the Git SDK Bash after the UCRT64 migration

dscho:
As part of running the Git for Windows SDK, it tries to guarantee that a Desktop shortcut is available to launch the Git SDK Bash quickly. After switching to UCRT64, it would try to create  one that ends in ` ucrt64`, but we want to reuse the previous ` 64-bit` suffix instead (to avoid ending up with _two_ shortcuts).


dscho:
/deploy git-extra

The [i686](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/34539690457), the [x86_64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/34539691597), the [mingw64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/34539692753) and the [arm64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/34539693975) workflow runs were started.

PR #743, opened 2026-09-11
installer: fix a couple regressions

dscho:
First of all: let `ExecAndCaptureOutput()` surround the Filename with quotes. Because when it comes to that `Filename` parameter, the docs of `ExecAndCaptureOutput()` at https://jrsoftware.org/ishelp/topic_isxfunc_execandcaptureoutput.htm refer to the documentation of the `Exec()` function, which say:

  Do not include quotes in the Filename parameter; the function will add
  them automatically.

While at it, also fix the parser of the output of `fsmonitor-daemon --help` (it joint the lines with Carriage Returns, but then looked for `stop` followed by a Line Feed; let's just use Line Feeds to join, too).

Also fix an infinite loop in `RPos()`. In fact, let's revert to the previous parsing logic, which was simpler because it did not have to parse line by line (which isn't the natural unit to parse `config -z` output, NUL would be the delimiter of choice).

While working on that, I also stumbled across a bug where multi-line values were mishandled (only the last line would be used, the rest would be packed into a "multi-line key", which is not allowed). That was an easy fix.

And finally, don't assume that `/ucrt64/bin/git.exe` was previously installed... That hid the bug where initially, the PR succeeded on x64 but failed on aarch64 (because the faulty config-parsing logic was skipped by mistake).

### DIFF
--- a/installer/helpers.inc.iss
+++ b/installer/helpers.inc.iss
@@ -68,6 +68,7 @@ begin
             Result:=i;
             Exit;
         end;
+	i:=i-1;
     end;
 
     Result:=0;
--- a/installer/helpers.inc.iss
+++ b/installer/helpers.inc.iss
@@ -54,26 +54,6 @@ begin
     Result[i]:=#0;
 end;
 
-// Search for the last occurence of a substring in another string or zero if not found.
-// RPos exists as a preprocessor function in innosetup, but not as a scripting function.
-function RPos(SubStr, S: AnyString): Integer;
-var
-    Len,i:Longint;
-begin
-    Len:=Length(SubStr);
-    
-    i:=Length(S)-Len+1;
-    while i>0 do begin
-        If (SameStr(SubStr,Copy(S,i,Len))) then begin
-            Result:=i;
-            Exit;
-        end;
-	i:=i-1;
-    end;
-
-    Result:=0;
-end;
-
 function AppendToArray(var AnArray:TArrayOfString;Str:String):Integer;
 begin
     Result:=GetArrayLength(AnArray)+1;
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -845,7 +845,7 @@ begin
     i:=1; j:=i; k:=i;
     while (j<=Length(StdOut)) do begin
         c:=Ord(StdOut[j]);
-        if (c=10) then // found the end of key
+        if (i=k) and (c=10) then // first found LF marks the end of key
             k:=j
         else if (c=0) then begin // found the end of the value
             if (i<>k) then begin // Ignore keys without values (LF is missing)
--- a/installer/install.iss
+++ b/installer/install.iss
@@ -801,7 +801,7 @@ end;
 
 function GetDefaultsFromGitConfig(WhichOne:String):Boolean;
 var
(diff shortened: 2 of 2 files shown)


dscho:
I don't know what's going on with the ARM64 installer, [the PR build](https://github.com/git-for-windows/build-extra/actions/runs/34634656826/job/103379558483?pr=743) seems to hang trying to run that installer.

I have access to an ARM64 VM where I tried to run it, and it does not even show the smart screen prompt, let alone shows any wizard, not even after waiting 10 minutes for it (I wasn't _really_ waiting, of course, I'm way too impatient to wait for 10 minutes for _anything_). Maybe it's because it is x64 _and_ an unsigned installer?

@dennisameling do you have any idea what could be going on here?

Maybe it's the difference between code-signed and unsigned installer, where the former works, the latter just hangs?

It can't be InnoSetup 7: [the earlier -rc0 build, using InnoSetup 7 as x64 installer](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/34619696918) (which _is_ code-signed) installs fine (modulo the issues fixed by _this here_ PR, as I verified on an x64 box installing the artifact from [this run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/34635915261) which used this PR branch).

rimrul (approved):
The changes look good to me, but I probably wont get around to investigating the ARM64 issues in a timely manner.

If you need to revert #593 for now to get this unstuck, that's fine by me.

dscho:
> The changes look good to me,

Thanks!

> but I probably wont get around to investigating the ARM64 issues in a timely manner.

That's okay.

I opened a PR reverting the first of these two commits, for testing, and [the PR build passes](https://github.com/git-for-windows/build-extra/actions/runs/34681630556/job/103521218046?pr=744)! So now I'm laboring under the idea that the `ExecAndCaptureOutput()` function might not work as expected when lauched from an x64 installer to call a native arm64 binary. Will keep investigating.

dscho:
> So now I'm laboring under the idea that the `ExecAndCaptureOutput()` function might not work as expected when lauched from an x64 installer to call a native arm64 binary.

Totally missed the mark. The problem is an infinite loop parsing the config. Will keep you posted.

dennisameling:
Thanks for the investigation. I have a physical arm64 device with the GfW arm64 SDK installed if needed, happy to help.

dscho:
> The problem is an infinite loop parsing the config.

And it's in `RPos()`: The decrementor is missing.

So why didn't it fail for x64? Because it tried to call `C:\Program Files\Git\ucrt64\bin\git.exe`, which does not exist, and therefore the installer assumes that it's not an upgrade, and totally skips parsing the existing system config!

dscho:
> Thanks for the investigation. I have a physical arm64 device with the GfW arm64 SDK installed if needed, happy to help.

@dennisameling whoops, I missed your reply... GitHub's UI is... let's say: improvable.

I _think_ that it's good to test now, the installer was [just uploaded in this PR build](https://github.com/git-for-windows/build-extra/actions/runs/34687197006?pr=743), could you test?

dscho:
No more hang!!!

PR #745, opened 2026-09-16
git-askpass: try harder to launch in the foreground

### LINKED ISSUE #6369: Git GUI password popup (for SSH-Key) don't get the focus

- [ ] I was not able to find an open or closed issue matching what I'm seeing

### Git for Windows version

```shell
2.55.0.2
```

### Windows version

Windows 11

### Windows CPU architecture

x86_64 (64-bit)

```shell

```

```shell
Windows Installer
```

### Other interesting things

_No response_

### Terminal/shell

Windows GUI

```shell
If I push the local Repository with SSH using an ed25519-key the password popup-window gets not the focus. So I need to use ALT+TAB to select the password-popup-window.
```

### Expected behaviour

Password-Popup-Window should get automatically the focus if a password is required.

### Actual behaviour

The Password-Popup-Window gets not the focus.

### Repository

_No response_

rimrul:
git-askpass should launch in the foreground to ask users for their Passwords. We've had reports that it doesn't always launch that way. Try harder to launch in the foreground and avoid callers seemingly hanging while waiting for input in a background window.

This fixes https://github.com/git-for-windows/git/issues/6369


rimrul:
/updpkgsums

The workflow run [was started](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/35090124249).

rimrul:
/updpkgsums

The workflow run [was started](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/35090723774).

